Issue #3, opened 2020-07-03
Terraform0.13 module "count" & "depends_on" does not work


teranos:
This tool is having problems with the Beta of Terraform0.13 configurations.

```
INFO Running command: /Users/teranos/go/bin/terraform init 
ERROR There are some problems with the configuration, described below. 
ERROR                                              
ERROR The Terraform configuration must be valid before initialization so that 
ERROR Terraform can determine which modules and providers need to be installed. 
ERROR                                         
ERROR Error: Reserved argument name in module block 
ERROR                                              
ERROR   on ingest.tf line 4, in module "ingest": 
ERROR    4:   count = var.enable_raven_ingest ? 1 : 0 
ERROR                                          
ERROR The name "count" is reserved for use in a future version of Terraform. 
ERROR                                      
ERROR                                         
ERROR Error: Reserved argument name in module block 
ERROR                                              
ERROR   on ingest.tf line 27, in module "ingest": 
ERROR   27:   depends_on = [               
ERROR                                          
ERROR The name "depends_on" is reserved for use in a future version of Terraform. 
ERROR                                      
exit status 1
```

aliscott:
@teranos thanks, I've now added initial module support in [v0.2.0](https://github.com/aliscott/infracost/releases/tag/v0.2.0).

It also uses your installed `terraform` binary to generate the plan files, so if you have a separate terraform 0.13 binary you can use the `TERRAFORM_BINARY` env variable to specify it, e.g.:

```
TERRAFORM_BINARY=~/bin/terraform_0.13 infracost ...
```

Let me know if you are still seeing any issues.

PR #5, opened 2020-07-03
Feature/module support

aliscott:
This adds initial support for Terraform modules. It supports resources specified in `child_modules` in the Terraform plan files, which works with the `count`, `for_each` and `depends_on` functionality from Terraform 0.13. I haven't done extensive testing on this yet so it's likely there's still missing edge cases. This 

This also adds a `TERRAFORM_BINARY` env variable which you can use to pass in a path to a non-default Terraform binary.


aliscott:
Fixed #3.

PR #21, opened 2020-08-22
AWS Lambdas and Price Range concept

sashman:
Hey, I had a brief conversation with Ali about this...

I would be very helpful for our team to display calculation for Lambdas, however, it's a lot more nuanced than static resources. So I suggested a range estimation would be much more useful instead of not showing anything.

I tried giving this a little try with the code base as it is, sadly I didn't quite grasp all of the data modellings in time, you live and learn! But the concept can be very roughly proven with what I have below.

### Idea

Given the very basic defaults from TF about AWS lambdas using `aws_lambda_function` resource, for example, timeout and memory limit, you can estimate an upper bound on the number of requests possible within an hour. Same with GB/s costs.

Further, if the TF specifies any triggering mechanisms, for example, rate/schedule, it's possible to estimate the lower bounds of executions.

And this is all prior to using empirical data from CloudWatch, which can make this estimation process even more useful.

Of course, using lower and upper bounds is just the start, and this idea can be expanded into more glorious statistical wonders!

### Implementaiton

Given an example TF file with the following resource:

```terraform
resource "aws_lambda_function" "lambda" {
  ...
  # timeout default to 3 seconds
}
```

I can generate a table output with infracost like this:

image

Note: The quantity is screwed up and should be the number of possible requests in an hour, etc. Also the subtotal and total is not calculated. However, the important part is that the price change information is now given in a `MIN - MAX` format which allows me to make a decision/start a conversation about the new resources I'm adding.

### DIFF
--- a/internal/terraform/aws/lambda.go
+++ b/internal/terraform/aws/lambda.go
@@ -1,0 +1,38 @@
+package aws
+
+import (
+	"infracost/pkg/resource"
+
+	"github.com/shopspring/decimal"
+)
+
+func NewLambdaFunction(address string, region string, rawValues map[string]interface{}) resource.Resource {
+	r := resource.NewBaseResource(address, rawValues, true)
+
+	costPerReq := decimal.NewFromFloat(0.2 / 1000000)
+
+	// From attached aws_cloudwatch_event_rule
+	// schedule_expression = "rate(1 minute)"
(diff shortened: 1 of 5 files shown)


aliscott:
Thanks @sashman, implementation of this will start with https://github.com/aliscott/infracost/issues/25.

PR #37, opened 2020-08-29
Add initial prices for aws lambda

### LINKED ISSUE #25: Add initial prices for AWS Lambda
Following on from discussions with @sashman.

We can add the prices for Lambda initially, so at least we can see that this resource does have a price, and what the price is for different memory configurations. We can do this initially with something like this:

```
  NAME                             QUANTITY  UNIT        PRICE PER UNIT  HOURLY COST      MONTHLY COST

  aws_instance.web_app
  ├─ instance hours (t3.micro)          730  hour                0.0104       0.0104           7.5920
  └─ root_block_device GB                15  GB/month            0.0021       0.0021           1.5000
  Total                                                                       0.0125           9.0920

  aws_lambda_function.test_lambda
  └─ requests                            1K  100ms request       0.0002  coming soon      coming soon
  Total                                                                  coming soon      coming soon

  OVERALL TOTAL                                                               0.0700          51.0670
```

In the future we can then work out the following to be able to show cost ranges:
* How we can recognise min/max bounds for monthly requests and execution time.
* How users can specify their own bounds for these.

sashman:
Fixes #25

* Create an overriding label in a price component to use as a price place holder
* Add a lambda resource which uses the above label to say "coming soon"

Important: The label value defaults to `""` and if so it will not be displayed. If any other string is used, the label will be displayed in the breakdown instead of the calculated cost.

Example:

```
  NAME                        QUANTITY  UNIT     HOURLY COST  MONTHLY COST

  aws_lambda_function.lambda
  └─ Requests                        1  request  coming soon   coming soon
  Total                                               0.0000        0.0000

  OVERALL TOTAL                                       0.0000        0.0000
```


pkg/output/table.go
```diff
@@ -94,12 +94,22 @@ func ToTable(resourceCostBreakdowns []costs.ResourceCostBreakdown) ([]byte, erro
 			totalHourly = totalHourly.Add(priceComponentCost.HourlyCost)
 			totalMonthly = totalMonthly.Add(priceComponentCost.MonthlyCost)
 
+			hourlyCost := formatDecimal(priceComponentCost.HourlyCost, "%.4f")
```
alikhajeh1, line 97: should we apply a similar logic to the json.go file?

sashman: that's up to you guys, do you want the JSON output to include a placeholder label @alikhajeh1 ?

alikhajeh1: We can add that when we update the json output to include more details.

PR #38, opened 2020-08-30
GH Action to automatically create a release and artifact based on git tags

sashman:
All that's needed to cut a release is:

```
git tag v*
git push --tags
```


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,39 @@
+on:
+  push:
+    branches: [master]
```
sashman, line 3:
```suggestion
```

alikhajeh1:
Nice and simple! Is it possible for it to upload the following 3 files like we have in https://github.com/aliscott/infracost/releases/tag/v0.4.1 ?
```
infracost-darwin-amd64.tar.gz
infracost-linux-amd64.tar.gz
infracost-windows-amd64.tar.gz
```

sashman:
> Nice and simple! Is it possible for it to upload the following 3 files like we have in https://github.com/aliscott/infracost/releases/tag/v0.4.1 ?
> 
> ```
> infracost-darwin-amd64.tar.gz
> infracost-linux-amd64.tar.gz
> infracost-windows-amd64.tar.gz
> ```

Shall have a look into this

sashman:
@alikhajeh1 updated to include all targets using `make release`

sashman:
Example release here: https://github.com/sashman/infracost/releases/tag/v0.0.1-test

.github/workflows/create-release.yml
```diff
@@ -0,0 +1,58 @@
...
+          make release
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: false
```
alikhajeh1, line 27: @sashman what do you think about setting this to true? GH says "This is a draft and won’t be seen by the public unless it’s published." for drafts, so after we `git tag v*` and `git push --tags`, we could go to the GH release tab, add the release notes then hit the publish button. We could avoid that and Edit a published release to add the notes into the desc box. Both methods achieve the same thing but if we want to add some other hooks/actions in the future that use the desc, the former might be cleaner?

alikhajeh1:
Thanks @sashman - this is awesome!

.github/workflows/create-release.yml
sashman: Sounds good, there is more tooling available to create changelog notes on a release, however, this makes sense at the moment.

PR #47, opened 2020-09-01
Feature docker

### LINKED ISSUE #45: Dockerhub image for infracost
We can modify https://github.com/aliscott/infracost/blob/master/.github/workflows/create-release.yml to push to DockerHub (we have the dockerhub infracost org now)

unfor19:
- Required GitHub Actions secrets: 
   1. `DOCKER_USER`
   1. `DOCKER_PASSWORD`
   1. `DOCKER_ORG`
   1. `DOCKER_REPOSITORY`
- Runs on alpine and includes terraform
- Build ARGS: `ARCH=linux` and  `TERRAFORM_VERSION=0.12.25`
- Image size: compressed 26MB, uncompressed 86MB
- Test Docker image- git clone this repo and then execute:
```
docker run  -v $PWD/:/code/  --rm unfor19/infracost -tfdir /code/examples/small_terraform
```

Closes #45

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.*
+!.env
+Dockerfile
+build/
+.terraform/
+*tfstate*
+*.png
+*.md
+examples/
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,25 @@
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - "README.md"
+
+name: Publish to DockerHub
+
+jobs:
+  dockerhub:
+    name: Build and Push to DockerHub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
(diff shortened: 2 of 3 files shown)


sashman (approved):

alikhajeh1:
@unfor19 thank you! I'll setup the required GH Action secrets later today so we start pushing to https://hub.docker.com/u/infracost...

@sashman / @unfor19 do you know if there's an easy way we can tweak https://github.com/aliscott/infracost/pull/47/files#diff-12a86cef0c4707531fdbabac3e38cb2aR24 so we could also call this action from https://github.com/aliscott/infracost/blob/master/.github/workflows/create-release.yml so tagged releases get pushed to dockerhub too? or is there a better way to handle those pushes?

sashman:
@alikhajeh1 w/o trying anything you can always copy-paste the action with a `on` tags condition and read the git tag, rather that hardcoding `latest`

unfor19:
@alikhajeh1 , I agree with @sashman, this is what I usually do.
It's **not** a bad practice, since the logic of building is in the Dockerfile, and this file is used by both pipelines, so it's a one-time copy-paste. I wish we could put it in the same file, but I don't think it's possible with GitHub Actions (for different triggers).

Regarding the tag - you can use the flag `tag_with_ref`, this will tag the version, read more about it in the [Docker GitHub Action](https://github.com/marketplace/actions/build-and-push-docker-images#tag_with_ref)

PR #65, opened 2020-09-04
Refactor lambda

aliscott:


aliscott:
@sashman FYI this adds some usage in for Lambda. I've removed the override label for now since we can now specify usage. We should think of a nice way of notifying the user if they haven't added usage that it is possible.

PR #66, opened 2020-09-05
Bump go to v1.15

chenrui333:

PR #67, opened 2020-09-05
Fix go module path

chenrui333:
The go module path is actually not correct.


aliscott:
Closing in favor of #66.

Issue #76, opened 2020-09-09
Panic using tfjson file


anthonydahanne:
Hello!
Using the latest `0.5.1` version, I got the following with a `plan.json`file that I generated from a `plan.save` (same workflow as described in the docs)

```
infracost --tfdir=. --tfjson=plan.json
⠋ Calculating costs… panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/infracost/infracost/pkg/schema.AddRawValue(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x87d84f, 0x6, 0x7ef520, ...)
        /home/runner/work/infracost/infracost/pkg/schema/resource_data.go:49 +0x106
github.com/infracost/infracost/internal/providers/terraform.parseResourceData(0x5, 0xc000270000, 0x48de2, 0x0, 0x0, 0x0, 0x0, 0x5, 0xc000296d2f, 0x45f, ...)
        /home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:64 +0x5c1
github.com/infracost/infracost/internal/providers/terraform.parsePlanJSON(0xc0001f0000, 0x48de2, 0x7fe00, 0x48de2, 0x7fe00, 0x0)
        /home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:32 +0x367
github.com/infracost/infracost/internal/providers/terraform.(*terraformProvider).LoadResources(0xc000099830, 0xc00010fd88, 0x2, 0x2, 0x0, 0x0)
        /home/runner/work/infracost/infracost/internal/providers/terraform/provider.go:59 +0x75
main.main.func2(0xc0000bcf00, 0xc0000acf00, 0xe)
        /home/runner/work/infracost/infracost/cmd/infracost/main.go:116 +0x253
github.com/urfave/cli/v2.(*App).RunContext(0xc000082900, 0x91f680, 0xc0000b4010, 0xc0000bc040, 0x4, 0x4, 0x0, 0x0)
        /home/runner/go/pkg/mod/github.com/urfave/cli/v2@v2.2.0/app.go:315 +0x70b
github.com/urfave/cli/v2.(*App).Run(...)
        /home/runner/go/pkg/mod/github.com/urfave/cli/v2@v2.2.0/app.go:215
main.main()
        /home/runner/work/infracost/infracost/cmd/infracost/main.go:147 +0x4b0
```

I can provide the json file if it helps
Thanks !

aliscott:
[...]

PR #80, opened 2020-09-09
Fix nil map entry panic when Terraform JSON values are empty

aliscott:
Potential fix for https://github.com/infracost/infracost/issues/76


alikhajeh1:
LGTM

PR #89, opened 2020-09-10
feat(cli): use pwd as default value for `--tfdir`

hakamadare:


cmd/infracost/main.go
```diff
@@ -41,9 +41,11 @@ func main() {
 				TakesFile: true,
 			},
 			&cli.StringFlag{
-				Name:      "tfdir",
-				Usage:     "Path to the Terraform project directory",
-				TakesFile: true,
+				Name:        "tfdir",
+				Usage:       "Path to the Terraform project directory",
+				TakesFile:   true,
+				Value:       get_cwd(),
```
aliscott, line 47:
```suggestion
				Value:       getcwd(),
```

cmd/infracost/main.go
```diff
@@ -150,3 +152,13 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func get_cwd() string {
```
aliscott, line 156:
```suggestion
func getcwd() string {
```

cmd/infracost/main.go
```diff
@@ -150,3 +152,13 @@ func main() {
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Warn(err)
+		cwd = "UNKNOWN"
```
aliscott, line 160:
```suggestion
		cwd = ""
```
Any benefit of having "UNKNOWN" over just a blank string? We check for a blank string when validating the args here: https://github.com/infracost/infracost/blob/master/internal/providers/terraform/provider.go#L23

aliscott (commented):
Thanks @hakamadare! This is a good idea.

cmd/infracost/main.go
hakamadare: not particularly; i don't tend to like blank strings b/c i like to interpolate values into log messages, and blank strings don't interpolate as helpfully as placeholders like "UNKNOWN", but this is a weak preference.

aliscott:
Thanks @hakamadare!

PR #93, opened 2020-09-10
Support nil cost components

### LINKED ISSUE #92: Support nil cost components
There are cases where combining usage data with actual resources creates conditions that a cost component makes no sense. For instance, a provisioned DynamoDB shall not have `read/write request units`.
Here is a simple code for the `write request units` cost component:
```go
func NewDynamoDBTable(d *schema.ResourceData, u *schema.ResourceData) *schema.Resource {
  costComponents := make([]*schema.CostComponent, 0)
  costComponents = append(costComponents, wruCostComponent(d))
}

func wruCostComponent(d *schema.ResourceData) *schema.CostComponent {
  costComponent := &schema.CostComponent{}
  if billingMode == "PROVISIONED" {
    log.Warnf("Skipping write request units usage data for %s. Provisioned mode does not support write request units.", d.Address)
    return costComponent
  }
  ...
  return costComponent
}

```
This way, the core produces a warning for the empty cost component and the output shows an empty line.

The other way around is to return `nil` and check the result of the function and skip it if it's nil but it makes the code filled with the same conditions.

I propose we add support for `nil` cost components in the core.

sinabakh:
**Issues:**
- Fixes #92

### DIFF
--- a/cmd/infracost/main.go
+++ b/cmd/infracost/main.go
@@ -117,6 +117,10 @@ func main() {
 			if err != nil {
 				return err
 			}
+			err = provider.PostProcessResources(resources)
+			if err != nil {
+				return err
+			}
 			err = prices.PopulatePrices(resources)
 			if err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


sinabakh:
It's not complete, it needs test and etc... But it's enough for discussion.

aliscott:
@sinabakh should we close this if we don't need it just now?

sinabakh:
Yeah, the current approach seems to be more clean. Close it for now.

PR #96, opened 2020-09-11
Change price explorer scritps for a better output and fix bug

### LINKED ISSUE #95: Enhancing price explorer scripts
Thanks to @alikhajeh1 our simple scripts are way better than before 😃 . There are still some improvements which we can make:

- Price explorer script `3_find_attribute_keys.sh` always outputs `2 different products found.`
- creating tmp files pollute the directory and it's not a clean solution
- The output for the scripts can be enhanced

m-yosefpor:
- [x] Price explorer script `3_find_attribute_keys.sh` always outputs `2 different products found.`
- [x] creating tmp files pollute the directory and it's not a clean solution
- [x] The output for the scripts can be enhanced

Closes #95

### DIFF
--- a/scripts/price_explorer/3_find_attribute_keys.sh
+++ b/scripts/price_explorer/3_find_attribute_keys.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 attribs=$(
 gq https://pricing.infracost.io/graphql -q "
 query {
@@ -13,18 +13,21 @@ query {
     	productHash
     	attributes { key , value }
     }
-}" | jq -r '.data.products[] | @base64' | head -2
+}" | jq -r '.data.products[] | @base64'
 )
 
-i=0
-for x in $attribs; do
-    echo $x | base64 --decode | jq '.attributes | map ({ (.key): .value} ) | add' | jq ".$3"  | tee tmp-$i >/dev/null
-    i=$(expr $i + 1)
-done
-echo Found $i different products
-echo "\n#####################################\n"
+N_files=$(echo $attribs | wc -w)
 
-if [ $i -ne 1 ]; then
-     diff -U1 tmp-0 tmp-1 | grep -v "__typename"
+if [ $N_files -ne 1 ]; then
+  echo -e "\n#####################################\n"
+  echo Found $N_files different products
+  echo -e "------------------\n"
+
+  f=$(echo $attribs | cut -d' ' -f1 | base64 --decode | jq '.attributes ')
+  s=$(echo $attribs | cut -d' ' -f2 | base64 --decode | jq '.attributes ')
+
+  diff -U1 <( echo "$f" ) <( echo "$s" ) | grep -v "__typename" | grep -v "/dev/fd/"
+else
+  echo "Only 1 Match product found"
 fi
 
--- a/scripts/price_explorer/3_find_attribute_keys.sh
+++ b/scripts/price_explorer/3_find_attribute_keys.sh
@@ -28,6 +28,6 @@ if [ $N_files -ne 1 ]; then
 
   diff -U1 <( echo "$f" ) <( echo "$s" ) | grep -v "__typename" | grep -v "/dev/fd/"
 else
-  echo "Only 1 Match product found"
+  echo "Only 1 match product found"
 fi
 
--- a/scripts/price_explorer/3_find_attribute_keys.sh
+++ b/scripts/price_explorer/3_find_attribute_keys.sh
@@ -18,7 +18,12 @@ query {
 
 N_files=$(echo $attribs | wc -w)
 
-if [ $N_files -ne 1 ]; then
+if [ $N_files -eq 0 ]; then
+  echo "No match product found"
+elif [ $N_files -eq 1 ]; then
+  echo "Only 1 match product found"
+  echo ProductHash: $(echo $attribs | cut -d' ' -f1 | base64 --decode | jq '.productHash')
+else
   echo -e "\n#####################################\n"
   echo Found $N_files different products
   echo -e "------------------\n"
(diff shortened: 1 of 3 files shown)


m-yosefpor:
Should I remove `scripts/price_explorer/tmp-*` line from `.gitignore` now? These files are not generated anymore, but if people have used old scripts before, they might accidentally push those files to their PR. We might want to keep that line for a while, and remove it later.

alikhajeh1:
Thanks! The .gitignore isn't harming anyone just now so I'd keep it for a few releases then we can remove it to tidy up.

scripts/price_explorer/3_find_attribute_keys.sh
```diff
@@ -13,18 +13,21 @@ query {
...
+  echo -e "\n#####################################\n"
+  echo Found $N_files different products
+  echo -e "------------------\n"
+
+  f=$(echo $attribs | cut -d' ' -f1 | base64 --decode | jq '.attributes ')
+  s=$(echo $attribs | cut -d' ' -f2 | base64 --decode | jq '.attributes ')
+
+  diff -U1 <( echo "$f" ) <( echo "$s" ) | grep -v "__typename" | grep -v "/dev/fd/"
+else
+  echo "Only 1 match product found"
```
alikhajeh1, line 31: can we output the 1 key that matched? or is that already done?

m-yosefpor: Sorry I don't get it. Do you mean line 35? Or for the first part of `if`?
The else here happens if there is only a single unique product with a given service and Family product (which I think is rare, but it's here in case just to avoid a bug when list has 1 member). I'll also add for when `no product found`. 

But for the case if 2 or more products are found, the diff will be printed, but also in a different way than before (I changed jq part) , so first the common key appears (if any), and then the +/- files. Sth like:

```
key: "databaseEngine"
+ value: "MySQL"
- value: "PostgreSQL"
```

But not only one of them, all the diff part. For non-common keys, the out put is like:

```
+ key: "sth"
+ value: "value1"
- key: "sthelse"
- value: "value2"
```

scripts/price_explorer/3_find_attribute_keys.sh
```diff
@@ -13,18 +13,21 @@ query {
...
-     diff -U1 tmp-0 tmp-1 | grep -v "__typename"
+if [ $N_files -ne 1 ]; then
+  echo -e "\n#####################################\n"
+  echo Found $N_files different products
+  echo -e "------------------\n"
+
+  f=$(echo $attribs | cut -d' ' -f1 | base64 --decode | jq '.attributes ')
+  s=$(echo $attribs | cut -d' ' -f2 | base64 --decode | jq '.attributes ')
+
+  diff -U1 <( echo "$f" ) <( echo "$s" ) | grep -v "__typename" | grep -v "/dev/fd/"
```
sashman, line 29: @m-yosefpor is there a chance of instead of using a `diff` the attributes, could you collect them all and present a unique list of keys (and/or key/value pairs)?

m-yosefpor: This actually was supposed to be a simple script which has a useful output, but yeah, I think we should rewrite them with python subprocess, so that we can more easily manipulate the json files and be able to implement more complex logics easily. Bash is not handy to handle this kind of works easily (or I just don't know how to do it easily).

sashman: I think it should be possible with a `jq` script, I'll have a test and see what's possible. Might not need to reach for python

scripts/price_explorer/3_find_attribute_keys.sh
alikhajeh1: @m-yosefpor I meant the following case (sorry should have mentioned it in my original comment), where there is only one attribute key for the productFamily, can we just output what is it so the user can see it and pass it to the next step?

```
 ./3_find_attribute_keys.sh AmazonDocDB "Database Storage"
Executing query... done
Found 1 different products

#####################################
```

scripts/price_explorer/3_find_attribute_keys.sh
alikhajeh1: I got a little bit frustrated with bash/jq too but @sashman feel free to give it a shot in another PR, I copped-out in my previous scripts PR and went with `echo "$attribs" | sort | uniq` for `4_find_attribute_values.sh`

scripts/price_explorer/3_find_attribute_keys.sh
m-yosefpor: @alikhajeh1 I'm really sorry ali jan, I still don't get your point 😢. If there is only a single product that we have found, so what is the point of continuing to next step? I mean isn't the whole productFilter idea to get to a unique productHash? and I can't recall, but should we always pass a key/value attributeFilter to the productFilter? I thought we can ignore it, and not providing that if we get to a unique product.

For the case of AmazonDocDB, the following query returns a unique productHash (without specifying any key/value attributeFilter):
```

query {
    products (
    filter: {
      vendorName: "aws"
      region: "us-east-1"
      service: "AmazonDocDB"
      productFamily: "Database Storage"
    }
  ){
    	productHash
      attributes { key , value }

    }
}
```

And BTW, I see it has many key/value attributes, WDYM by `there is only one attribute key for the productFamily` part?

I added `productHash` to the output in case of a unique product (to be able to use in tests).

Sample output:
```
./3_find_attribute_keys.sh AmazonDocDB "Database Storage"
Executing query... done
Only 1 match product found
ProductHash: "856b9e5bd87c953bffd1df698a6a1b3d"
```

alikhajeh1: Good point, you're right, I could just run `./4_find_attribute_values.sh AmazonDocDB "Database Storage" "usagetype"` and see it has 1 value and use that in the resource go file. I could also pass the same product/attributeFilters to https://pricing.infracost.io/graphql to find the priceHash to write my integration test.

PR #98, opened 2020-09-11
Add version flag

### LINKED ISSUE #97: Adding a `version` flag to infracost cli
We shall have a version flag, which outputs both the infracost version, and the current terraform version which infracost uses for that environemnt. That will come in handy and also all most all cli tools have, so people expect it to be there.

m-yosefpor:
Sample output of `infracost --version`

```
Infracost v0.5.1-15-g779d8bf-dirty
Terraform v0.12.29

Your version of Terraform is out of date! The latest version
is 0.13.2. You can update by downloading from https://www.terraform.io/downloads.html
```

Fixes #97

### DIFF
--- a/internal/providers/terraform/cmd.go
+++ b/internal/providers/terraform/cmd.go
@@ -3,7 +3,6 @@ package terraform
 import (
 	"bufio"
 	"bytes"
-	"fmt"
 	"os"
 	"os/exec"
 	"strings"
@@ -46,5 +45,4 @@ func TerraformVersion() (string, error) {
 	}
 	out, err := exec.Command(terraformBinary, "-version").Output()
 	return strings.SplitN(string(out), "\n", 2)[0], err
-}
 }
--- a/internal/providers/terraform/cmd.go
+++ b/internal/providers/terraform/cmd.go
@@ -39,16 +39,12 @@ func terraformCmd(options *cmdOptions, args ...string) ([]byte, error) {
 	return outbuf.Bytes(), err
 }
 
-func TerraformVersion() error {
+func TerraformVersion() (string, error) {
 	terraformBinary := os.Getenv("TERRAFORM_BINARY")
 	if terraformBinary == "" {
 		terraformBinary = "terraform"
 	}
 	out, err := exec.Command(terraformBinary, "-version").Output()
-	firstLine := string(out)[0:strings.Index(string(out), "\n")]
-	if err == nil {
-		fmt.Println(firstLine)
-	}
-
-	return err
+	return strings.SplitN(string(out), "\n", 2)[0], err
+}
 }
--- a/internal/providers/terraform/cmd.go
+++ b/internal/providers/terraform/cmd.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/infracost/infracost/pkg/config"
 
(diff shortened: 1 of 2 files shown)


alikhajeh1:
Pulumi has `pulumi version` (a subcommand vs a flag)

Terraform has `terraform --version` and `terraform version` 

GH's cli has `gh --version` (as well as other subcommands but --version doesn't work when used with subcommands)

I don't have a strong opinion on the way to do it, just wanted to mention the above as I think we'll move towards subcommands soon.

Is it a bit awkward/confusing to be told that my Terraform is out of date when I check the version of Infracost?

m-yosefpor:
I thought about it and felt like keeping `terraform version` output integrity. I have no idea. we can keep only the first line of its output, so no warning commands like this. which do you thing works best?

For the former concern, I shall just mention we are using `-v` shorthand flag for `verbose` (ansible is the same too, also --version works) I saw this pattern (--version) is prevailing in cli tools, esp the ones which are flag based, instead of subcommand based. We can follow terraform method too. Supporting both!

aliscott:
Yeah I'm now leaning towards `--version` as well, but have no strong preference.
I'm also leaning towards not showing the Terraform out of date warnings. We'll have other IaC dependencies in the future that we might want to list as well.

m-yosefpor:
Code changed as suggested. Sample output:
```
Infracost v0.5.1-16-gbee2627-dirty
Terraform v0.12.29
```

internal/providers/terraform/cmd.go
```diff
@@ -36,3 +38,17 @@ func terraformCmd(options *cmdOptions, args ...string) ([]byte, error) {
...
+	if terraformBinary == "" {
+		terraformBinary = "terraform"
+	}
+	out, err := exec.Command(terraformBinary, "-version").Output()
+	firstLine := string(out)[0:strings.Index(string(out), "\n")]
+	if err == nil {
+		fmt.Println(firstLine)
+	}
+
+	return err
```
aliscott, line 53:
```suggestion
func TerraformVersion() (string, error) {
	terraformBinary := os.Getenv("TERRAFORM_BINARY")
	if terraformBinary == "" {
		terraformBinary = "terraform"
	}
	out, err := exec.Command(terraformBinary, "-version").Output()
	return strings.SplitN(string(out), "\n", 2)[0], err
}
```

I wonder if we should return the version here in case we want to re-use this anywhere else?

And then in main.go we do something like this?

```go
				v, err := terraform.TerraformVersion()
				if err != nil {
					return err
				}
				fmt.Println(v)
```
What do you think?

m-yosefpor: Great 👍

m-yosefpor: Shouldn't we just `return err` in anycase?  I mean we only want to print the version and return, not executing the rest. I checked it's also the default behavior for terraform. e.g. `terraform --version apply` only print version. Moreover, the output of this block would be:
```
Infracost v0.5.1-18-g2573a74-dirty
Terraform v0.12.29
Please provide either the path to the Terrafrom project (tfdir) or a Terraform Plan JSON file (tfjson)
NAME:
   infracost - Generate cost reports from Terraform plans

USAGE:
   infracost [global options] command [command options] [arguments...]

COMMANDS:
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --tfjson value            Path to Terraform Plan JSON file
   --tfplan value            Path to Terraform Plan file. Requires tfdir to also be set
   --tfdir value             Path to the Terraform project directory
   --log-level value         Log level (TRACE, DEBUG, INFO, WARN, ERROR) (default: "WARN")
   --output value, -o value  Output (json, table) (default: "table")
   --no-color                Turn off colored output (default: false)
   --api-url value           Price List API URL
   --version                 Prints the version of infracost and terraform (default: false)
   --help, -h                show help (default: false)
```

So I committed it in another way.

PR #106, opened 2020-09-13
Simplify Release Workflow

anton-yurchenko:
Simplify and speed up a release workflow by using [git-release](https://github.com/marketplace/actions/git-release) action.


sashman:
Could you add the Dockerfile locally, so we can see what it does, please?

anton-yurchenko:
It's an OSS action published on Marketplace, there is no need to build it during the execution.

sashman (approved):

sashman:
@anton-yurchenko oky doke, LGTM 👍

Issue #121, opened 2020-09-17
Show price for resources with no cost


sinabakh:
According to `Level 1 support` of https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md we should make all currently existing resources compatible with these requirements.

Resources:
- [ ] aws_alb
- [ ] aws_db_instance
- [ ] aws_docdb_cluster_instance
- [ ] aws_ecs_service
- [ ] aws_elb
- [ ] aws_lambda_function
- [ ] aws_lb
- [ ] aws_rds_cluster_instance
- [ ] aws_elasticsearch_domain
- [x] aws_dynamodb_table
- [x] aws_nat_gateway
- [x] aws_instance (https://github.com/infracost/infracost/pull/126, https://github.com/infracost/infracost/pull/129)
- [ ] aws_ebs_snapshot (https://github.com/infracost/infracost/pull/143)
- [ ] aws_ebs_snapshot_copy (https://github.com/infracost/infracost/pull/143)
- [ ] aws_ebs_volume (https://github.com/infracost/infracost/pull/143)
- [ ] aws_autoscaling_group (https://github.com/infracost/infracost/pull/142)

sinabakh:
The question is shoule we list all possible costs or just the main required costs. Take Pay Per Request DynamoDB for instance, it has 3 main cost components : Read/Write units and storage price, one can't use this service without utilizing all these three components. But there are some additional components like backup storage, shall we add 0 cost components for them also?

2. Do we need to show 0 cost components for resources missing in terraform and covered in infracost terraform plugin? Like most pay per request components?

alikhajeh1:
1. yeah I think that would be good, @aliscott did something similar for `aws_instance`, which I think we can mark as done in this issue list.
2. I wouldn't work on the infracost terraform plugin until we know users want it and will use it. Despite pushing users for feedback on that, I've gotten very few replies, and one strong rejection (the user saw estimates as dead/stale-code that needs to be removed after the service is deployed).

sinabakh:
> What I think is there might be no point in showing the 0 costs for the resources we can't support.

There is value in developers understanding **what are the cost components of this resource** and **how much is the price of each component**, even if the cost shows 0 (since it's usage-based).

It took me ~1h to understand the pricing of ECS Fargate and I had to spin stuff up and read docs to see if the default 20GB docker disk cost money or was free... `infracost` takes that 1h to 1min, that's the base value it provides.

[...]

PR #122, opened 2020-09-17
Show skipped resources

### LINKED ISSUE #84: Output message showing number of resources that we don't support
e.g. "2 out of 26 resources could not be estimated because they don't have a price or Infracost does not yet support them (https://www.infracost.io/docs/supported_resources)"

Debug level logging could show the resource addresses.

sinabakh:
Fixes #84 

Here's a sample output with no flag:
```
  NAME           MONTHLY QTY  UNIT  PRICE  HOURLY COST  MONTHLY COST  

  OVERALL TOTAL                                 0.0000        0.0000  

3 out of 3 resources were skipped, 1 are free, and 2 aren't supported in Infracost yet(https://www.infracost.io/docs/supported_resources), re-run with --show-skipped to see the list.
```

And with `--show-skipped` flag:
```
  NAME           MONTHLY QTY  UNIT  PRICE  HOURLY COST  MONTHLY COST  

  OVERALL TOTAL                                 0.0000        0.0000  

Skipped aws_proxy_configuration.my_proxy_configuration | This resource is free.
Skipped aws_dynamodb_table.my_dynamodb_table | This resource is not currently supported.
Skipped aws_db_instance.my_db_instance | This resource is not currently supported.
3 out of 3 resources were skipped, 1 are free, and 2 aren't supported in Infracost yet(https://www.infracost.io/docs/supported_resources), re-run with --show-skipped to see the list.
```

### DIFF
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -102,3 +102,19 @@ func generatePlanJSON(dir string, path string) ([]byte, error) {
 
 	return out, nil
 }
+
+func ShowSkippedResources(resources []*schema.Resource, showDetails bool) {
+	skippedCount := 0
+	unSupportedCount := 0
+	for _, r := range resources {
+		if r.IsSkipped {
+			skippedCount++
+			unSupportedCount++
+			if showDetails {
+				fmt.Printf("Skipped %s | This resource is not currently supported.\n", r.Name)
+			}
+		}
+	}
+	message := "%d out of %d resources were skipped, %d aren't supported in Infracost yet (https://www.infracost.io/docs/supported_resources), re-run with --show-skipped to see the list.\n"
+	fmt.Printf(message, skippedCount, len(resources), unSupportedCount)
+}
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -111,7 +111,11 @@ func ShowSkippedResources(resources []*schema.Resource, showDetails bool) {
 			skippedCount++
 			unSupportedCount++
 			if showDetails {
-				fmt.Printf("Skipped %s | This resource is not currently supported.\n", r.Name)
+				message := fmt.Sprintf("Skipped %s", r.Name)
+				if r.SkipMessage != "" {
+					message += fmt.Sprintf(" | %s.", r.SkipMessage)
+				}
+				fmt.Printf("%s\n", message)
 			}
 		}
 	}
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -115,7 +115,7 @@ func ShowSkippedResources(resources []*schema.Resource, showDetails bool) {
 			} else {
 				unSupportedCount++
 			}
-			if showDetails {
+			if showDetails && !r.IsFree() {
 				message := fmt.Sprintf("Skipped %s", r.Name)
 				if r.SkipMessage != "" {
 					message += fmt.Sprintf(" | %s.", r.SkipMessage)
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -124,6 +124,12 @@ func ShowSkippedResources(resources []*schema.Resource, showDetails bool) {
 			}
 		}
 	}
-	message := "%d out of %d resources were skipped, %d are free, and %d aren't supported in Infracost yet(https://www.infracost.io/docs/supported_resources), re-run with --show-skipped to see the list.\n"
-	fmt.Printf(message, skippedCount, len(resources), freeCount, unSupportedCount)
+	message := "%d out of %d resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources)"
+	if showDetails {
+		message += ".\n"
+	} else {
+		message += ", re-run with --show-skipped to see the list.\n"
+	}
+	fmt.Printf(message, unSupportedCount, len(resources))
+	fmt.Println("We're continually adding new resources, please create an issue if you'd like us to prioritize your list.")
 }
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -122,6 +122,9 @@ func ShowSkippedResources(resources []*schema.Resource, showDetails bool) {
 			}
 		}
 	}
+	if unSupportedCount == 0 {
+		return
+	}
 	message := "%d out of %d resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources)"
 	if showDetails {
 		message += ".\n"
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -103,7 +103,7 @@ func generatePlanJSON(dir string, path string) ([]byte, error) {
 	return out, nil
 }
 
-func ShowSkippedResources(resources []*schema.Resource, showDetails bool) {
+func CountSkippedResources(resources []*schema.Resource) (map[string]int, int, int, int) {
 	skippedCount := 0
 	unSupportedCount := 0
 	freeCount := 0
@@ -111,31 +111,13 @@ func ShowSkippedResources(resources []*schema.Resource, showDetails bool) {
 	for _, r := range resources {
 		if r.IsSkipped {
 			skippedCount++
-			if r.IsFree() {
-				freeCount++
-			} else {
-				unSupportedCount++
-				if _, ok := unSupportedTypeCount[r.ResourceType]; !ok {
-					unSupportedTypeCount[r.ResourceType] = 0
-				}
-				unSupportedTypeCount[r.ResourceType]++
+			// FIXME: Count free resources when https://github.com/infracost/infracost/issues/121 is done
+			unSupportedCount++
+			if _, ok := unSupportedTypeCount[r.ResourceType]; !ok {
+				unSupportedTypeCount[r.ResourceType] = 0
 			}
+			unSupportedTypeCount[r.ResourceType]++
 		}
 	}
-	if unSupportedCount == 0 {
-		return
-	}
-	message := "%d out of %d resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources)"
-	if showDetails {
-		message += ".\n"
-	} else {
-		message += ", re-run with --show-skipped to see the list.\n"
-	}
-	fmt.Printf(message, unSupportedCount, len(resources))
-	if showDetails {
-		for rType, count := range unSupportedTypeCount {
-			fmt.Printf("%d x %s \n", count, rType)
-		}
-	}
-	fmt.Println("We're continually adding new resources, please create an issue if you'd like us to prioritize your list.")
+	return unSupportedTypeCount, skippedCount, unSupportedCount, freeCount
 }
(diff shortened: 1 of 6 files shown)


sinabakh:
@alikhajeh1 @aliscott It lacks tests and some final touches. But we can discuss this approach before doing them.
This seemed the most flexible and clean way of doing it to me.

alikhajeh1:
Thank you :) I like where this is going! I thought more about it and I'm changing my mind on seeing free resources as explained below.

The output could look like this:
```
2 out of 26 resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources), re-run with --show-skipped to see the list.

We're continually adding new resources, please create an issue if you'd like us to prioritize your list.
```

Re-running with `--show-skipped` could add this to the top of the above output (a sorted list showing the number of resources per unsupported resource type):
```
10 x aws_sns_platform_application
8 x aws_workspaces_directory
3 x aws_elastic_beanstalk_application
```

Stepping back, here's what I want as a user:
I want to run infracost, understand the output, and see a summary of why the estimate is very different to my last month's bill. If I want to dig deeper, I should be able to see details, but there is no need to show me my resource names, just the resource type and how many of them I'm using is fine. I could create an issue with that list to ask for infracost to support them. I highly prefer not to read docs. 

Long term, hopefully this feature can become mostly redundant as we support more resources.

Let's look at the different cases for resources:
1. There are resources like aws_ecs_capacity_provider that don't have a direct cost, they can include configs. Some of these configs might influence costs but that's already handled in the resource that uses the "config" resource. Infracost should not show these as it's just noise.

2. There are resources that infracost partially supports, e.g. aws_ecs_service.launch_type==FARGATE is supported but launch_type==EC2 is not. If a terraform project uses the non-supported resource config, infracost should count that as a non-supported resource. As a maintainer, I can see the `SkipMessage` to learn about what the details are.

3. There are resources that infracost doesn't support yet. That should obviously be counted as a non-supported resource.

For resource components:
There are "obviously free" resource components, e.g. ecs_task_definition.name. There are suspicious resource components like ecs_task_definition.docker_volume_configuration that they don't cost anything. As long as I trust that infracost is covering all components in the resource, then there is no need to see these in the output.

Finally, for infracost docs:
Maybe just copying the supported resource list from the resource registry is good enough and we can add a comment or json annotation with notes. That means we don't need to ask contributors to send a docs PR, one less step for them.

sinabakh:
So the outputs are like this:

Without the flag:
```
  NAME           MONTHLY QTY  UNIT  PRICE  HOURLY COST  MONTHLY COST  

  OVERALL TOTAL                                 0.0000        0.0000  

7 out of 7 resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources), re-run with --show-skipped to see the list.
We're continually adding new resources, please create an issue if you'd like us to prioritize your list.
```

With `--show-skipped` flag:
```
  NAME           MONTHLY QTY  UNIT  PRICE  HOURLY COST  MONTHLY COST  

  OVERALL TOTAL                                 0.0000        0.0000  

7 out of 7 resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources).
1 x aws_nat_gateway 
1 x aws_dynamodb_table 
2 x aws_ebs_volume 
1 x aws_eip 
1 x aws_instance 
1 x aws_lambda_function 
We're continually adding new resources, please create an issue if you'd like us to prioritize your list.
```

sinabakh:
I just can't figure out why go-lint fails, I think sth is wrong with the ci/cd pipeline.

aliscott:
> I just can't figure out why go-lint fails, I think sth is wrong with the ci/cd pipeline.

CI output is messed up for some reason but I do see this `##[error]unreachable: unreachable code (govet)`.

Running this branch locally gives me this lint error:
```
pkg/schema/resource.go:86:2: unreachable: unreachable code (govet)
	return len(r.CostComponents) == 0
```

sinabakh:
@aliscott Yeah I left a FIXME comment there, the output was wrong because of #121 but since we don't show free resources anymore I will fix it now.

internal/providers/terraform/provider.go
```diff
@@ -102,3 +102,40 @@ func generatePlanJSON(dir string, path string) ([]byte, error) {
...
+	} else {
+		message += ", re-run with --show-skipped to see the list.\n"
+	}
+	fmt.Printf(message, unSupportedCount, len(resources))
+	if showDetails {
+		for rType, count := range unSupportedTypeCount {
+			fmt.Printf("%d x %s \n", count, rType)
+		}
+	}
+	fmt.Println("We're continually adding new resources, please create an issue if you'd like us to prioritize your list.")
```
alikhajeh1, line 140: This seems to look better when it's added in the `message` on line 128.

alikhajeh1:
@sinabakh nice one!

It would be great if we can ensure that infracost produces valid json with `--output json` (I'll add a CI check soon). Can you possibly add a `warnings: ["message here"]` to the json output? The idea is that for now, it's just one long string in a warnings array, but we can split it out into other warning fields in the future. The string warning can include the resource type list if `--show-skipped` is used.

sinabakh:
There are some major changes with a few breaking changes:
* Moved the skipped resource message builder to the outputs
* Moved the logic of counting and aggregating the skipped resource to a method in the provider
* BREAKING: The json output used to be an array, I changed it to an object.

So here is a sample output for the table:
```
  NAME           MONTHLY QTY  UNIT  PRICE  HOURLY COST  MONTHLY COST  

  OVERALL TOTAL                                 0.0000        0.0000  

7 out of 7 resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources).
We're continually adding new resources, please create an issue if you'd like us to prioritize your list.
1 x aws_dynamodb_table
2 x aws_ebs_volume
1 x aws_eip
1 x aws_instance
1 x aws_lambda_function
1 x aws_nat_gateway
```

And a sample output for the json:
```json
{"resources":[],"warnings":["\n7 out of 7 resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources).\nWe're continually adding new resources, please create an issue if you'd like us to prioritize your list.\n1 x aws_eip\n1 x aws_instance\n1 x aws_lambda_function\n1 x aws_nat_gateway\n1 x aws_dynamodb_table\n2 x aws_ebs_volume\n"]}
```

pkg/output/json.go
```diff
@@ -52,12 +59,41 @@ func newResourceJSON(r *schema.Resource) resourceJSON {
 	}
 }
 
-func ToJSON(resources []*schema.Resource) ([]byte, error) {
+func showSkippedResourcesJSON(resources []*schema.Resource, showDetails bool) []string {
+	unSupportedTypeCount, _, unSupportedCount, _ := terraform.CountSkippedResources(resources)
+	if unSupportedCount == 0 {
+		return nil
+	}
+	message := fmt.Sprintf("\n%d out of %d resources couldn't be estimated as Infracost doesn't support them yet (https://www.infracost.io/docs/supported_resources)", unSupportedCount, len(resources))
```
aliscott, line 67: Do we need a \n at the start here?

pkg/output/json.go
```diff
@@ -52,12 +59,41 @@ func newResourceJSON(r *schema.Resource) resourceJSON {
...
+	} else {
+		message += ", re-run with --show-skipped to see the list.\n"
+	}
+	message += "We're continually adding new resources, please create an issue if you'd like us to prioritize your list.\n"
+	if showDetails {
+		for rType, count := range unSupportedTypeCount {
+			message += fmt.Sprintf("%d x %s\n", count, rType)
+		}
+	}
+	return []string{message}
```
aliscott, line 79: Should we just return a string instead of []string?

pkg/output/json.go
```diff
@@ -52,12 +59,41 @@ func newResourceJSON(r *schema.Resource) resourceJSON {
 	}
 }
 
-func ToJSON(resources []*schema.Resource) ([]byte, error) {
+func showSkippedResourcesJSON(resources []*schema.Resource, showDetails bool) []string {
+	unSupportedTypeCount, _, unSupportedCount, _ := terraform.CountSkippedResources(resources)
```
aliscott, line 63: "unsupported" is a valid word, so we could use `unsupportedTypeCount` and `unsupportedCount` as the var names. Alternatively we could use `skippedTypeCount` and `skippedCount`.

pkg/output/table.go
```diff
@@ -68,6 +93,8 @@ func ToTable(resources []*schema.Resource) ([]byte, error) {
 
 	t.Render()
 
+	showSkippedResourcesTable(resources, c.Bool("show-skipped"), bufw)
```
aliscott, line 96: Could we do something like this:
```go
bufw.WriteString(skippedResourcesMessage(resources, c.Bool("show-skipped")))
```
and put the logic for the message into a shared function in `output/output.go`?

internal/providers/terraform/provider.go
```diff
@@ -102,3 +102,22 @@ func generatePlanJSON(dir string, path string) ([]byte, error) {
...
+
+func CountSkippedResources(resources []*schema.Resource) (map[string]int, int, int, int) {
+	skippedCount := 0
+	unSupportedCount := 0
+	freeCount := 0
+	unSupportedTypeCount := make(map[string]int)
+	for _, r := range resources {
+		if r.IsSkipped {
+			skippedCount++
+			// FIXME: Count free resources when https://github.com/infracost/infracost/issues/121 is done
```
aliscott, line 114: @sinabakh / @alikhajeh1 do we think we will add support for this any time soon? If yes then it makes sense to leave this in, but if not, I'd be tempted to remove the logic just now.

aliscott (commented):
This looks great!

> BREAKING: The json output used to be an array, I changed it to an object.

This is a good idea 👍

internal/providers/terraform/provider.go
alikhajeh1: +1 on removing this FIXME comment line, we don't need the free-resource logic just now.

I see #121 as reviewing/tweaking the existing resources so we support cost components for anything we can easily handle, I guess https://github.com/infracost/infracost/pull/126 is doing that for EC2.

pkg/output/json.go
sinabakh: In the table output it becomes hard to distinguish the aggrgated prices (last line of output) from this output but you are right I will remove it for now.

pkg/output/json.go
sinabakh: Right, will fix it.

internal/providers/terraform/provider.go
sinabakh: I think we will add the free resources soon, but for now, I will remove the whole free logic and will add it back in its related PR.

sinabakh: Done

pkg/output/table.go
sinabakh: Done

pkg/output/json.go
sinabakh: Changed the words, I used the unsupported because in the future we have free resources which are skipped but are unsupported.

sinabakh: Just changed the "unsupported" case. I think it's better to stick to skipped/unsupported

pkg/output/json.go
sinabakh: The whole function was removed.

pkg/output/json.go
sinabakh: Done

alikhajeh1:
If https://github.com/infracost/infracost/pull/126 is merged, we might want to rebase and tweak this branch too or send another PR

sinabakh:
@alikhajeh1 I checked the #126 I couldn't understand the need for tweaks? What is changed?

PR #125, opened 2020-09-23
Automatic resources doc generator

### LINKED ISSUE #123: Automatic supported resources docs generation
We need to be able to generate the contents of https://www.infracost.io/docs/supported_resources automatically.

sinabakh:
Fix #123

### DIFF
--- a/cmd/infracost-generate-docs/main.go
+++ b/cmd/infracost-generate-docs/main.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/infracost/infracost/internal/docs"
+	"github.com/infracost/infracost/internal/providers/terraform"
+	"github.com/infracost/infracost/pkg/config"
+	"github.com/infracost/infracost/pkg/version"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+)
+
+var logFormatter log.TextFormatter = log.TextFormatter{
+	DisableTimestamp:       true,
+	DisableLevelTruncation: true,
+}
+
+func customError(c *cli.Context, msg string) error {
+	color.HiRed(fmt.Sprintf("%v\n", msg))
+	_ = cli.ShowAppHelp(c)
+
+	return fmt.Errorf("")
+}
+
+func getcwd() string {
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Warn(err)
+		cwd = ""
+	}
+
+	return cwd
+}
+
+func main() {
+
+	app := &cli.App{
+		Name:                 "infracost",
+		Usage:                "Generate cost reports from Terraform plans",
+		EnableBashCompletion: true,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:      "input",
+				Usage:     "Path to docs templates",
+				TakesFile: true,
+			},
+			&cli.StringFlag{
+				Name:      "output",
+				Usage:     "Path to output of docs",
+				TakesFile: true,
+			},
+			&cli.StringFlag{
+				Name:  "log-level",
+				Usage: "Log level (TRACE, DEBUG, INFO, WARN, ERROR)",
+				Value: "WARN",
+			},
+			&cli.BoolFlag{
+				Name:  "no-color",
+				Usage: "Turn off colored output",
+				Value: false,
+			},
+			&cli.BoolFlag{
+				Name:  "version",
+				Usage: "Prints the version of infracost and terraform",
+				Value: false,
+			},
+		},
+		OnUsageError: func(c *cli.Context, err error, isSubcommand bool) error {
+			return customError(c, err.Error())
+		},
+		Action: func(c *cli.Context) error {
+
+			logFormatter.DisableColors = c.Bool("no-color")
+			log.SetFormatter(&logFormatter)
+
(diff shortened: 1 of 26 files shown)


sinabakh:
Here is the proposed change to the resource registry. Now, the question is how we are going to generate docs?

aliscott:
LGTM - to generate the docs maybe we could have a make target that writes the notes to a markdown file in this repo. We can then create a component in the docs site to read that. What do you think?

sinabakh:
I'm not sure about the features we have in Gihub but I agree, we probably can generalte a markdown file for our releases and in the site repo, download the artifcacts of this project.
I will work on generating the markdown, meanwhile we can think more about the right way to do the aggregation.

sinabakh:
Now we can run:
```
infracost-generate-docs --input docs/templates --output docs/generated
```
@alikhajeh1 Do you think that's enough? This generates the whole `supported_resources.md`.

internal/providers/terraform/aws/ecs_service.go
```diff
@@ -11,6 +11,14 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+func GetECSServiceRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "aws_ecs_service",
+		Notes: []string{"Fargate on-demand only."},
```
alikhajeh1, line 17: Only supports Fargate on-demand.

alikhajeh1 (approved):

internal/providers/terraform/aws/ecs_service.go
sinabakh: Done

alikhajeh1:
> Now we can run:
> 
> ```
> infracost-generate-docs --input docs/templates --output docs/generated
> ```
> 
> @alikhajeh1 Do you think that's enough? This generates the whole `supported_resources.md`.

Looks good. Here's how the end-to-end process could look like:
1. contributors runs that task and checks-in the supported_resources.md file into git. The idea being that master always has the latest, independently viewable, markdown file for supported resources. This file could be imported by https://www.infracost.io/docs/supported_resources/ from the frontend, meaning we never have to send a PR to the docs repo. The issue with using github artifacts is that I think they get zipped, so the docs frontend can't just import them from the browser.
2. contributors don't need to worry about git conflicts as they can always rebase/merge master into their branch and re-run- the task to create the latest docs.

what do you think?

sinabakh:
1. Regarding showing the files in frontend I'm not sure how it's possible.
We can setup a CI step so the docs get generated automatically in master branch, so contributors won't have to run this stage manually.

cmd/infracost-generate-docs/main.go
```diff
@@ -0,0 +1,128 @@
+package main
```
alikhajeh1, line 1: how do you feel about having this in scripts/docs-generator as that folder has other helpers? I was thinking cmd might be used for product subcommands (e.g. `infracost diff`) but I'm not sure yet. 
Also, what do you think about updating the Makefile so we can have `make docs`?

sinabakh: Since this script is an acutal golang command line app, I prefered to put it in `cmd/`, on the other hand `infracost diff`  and other subcommands' logic should reside in the `infracost` cmd since the subcommand is just an arg.
https://github.com/influxdata/influxdb/tree/master/cmd is an example of using multiple cmds
Anyway, there is no silverbullet so we just need to have a standard, what you think?
Regarding the makefile I agree with you.

alikhajeh1: you're right, cmd makes more sense for this.

alikhajeh1:
>     1. Regarding showing the files in frontend I'm not sure how it's possible.
>        We can setup a CI step so the docs get generated automatically in master branch, so contributors won't have to run this stage manually.

I looked into embedding or including markdown files and it doesn't seem possible/easy to do in `docusaurus` (our tools framework). So the options seem to be:
1. keep this PR as is, so we just run the docs generation command, manually copy/paste it into the docs repo when that needs updating (usually after we release a new version of infracost).
2. have supported_resources.md in the infracost repo, and link to it from inside https://www.infracost.io/docs. This is simple to do with the new command you added but it means people need to check that file into git. This seems ok but I'm not sure if conflicts will become annoying as they'll block PRs from being merged until the PR branch is rebased/master-merged.
3. update the infracost CI pipeline so it generates that file and saves it in the docs repo. This needs more investigations.

I think option 1 is ok for now, what do you think?

sinabakh:
Agreed, we can work on more automations later.

alikhajeh1:
@sinabakh plz let me know if this ready to merge. BTW, do we need `no-color` and `version` in this cmd?

sinabakh:
@alikhajeh1 Removed `no-color` and `version`. I added some new github workflows, changes and ... so here is the best approach I could come up with:
* Contributor works on new resources like before
* Upon merging/pushing to master, new docs are generated and are commited to the master using github actions. This way master and its docs are like nightly builds.
* With new releases, a `docs.tar.gz` file is also added to assets.
* In the other repo, we need to create the same version tag, and in the actions we can download the docs.tar.gz and append its contents to the original site.

alikhajeh1:
@sinabakh this is awesome! love it. I'll try to sort out the [other repo](https://github.com/infracost/docs) in the next few days.

alikhajeh1:
@sinabakh I noticed that https://github.com/infracost/infracost/blob/master/docs/generated/supported_resources.md appears `aws_alb / aws_alb` twice

sinabakh:
#127

PR #128, opened 2020-09-26
Remove alias from registry items and docs

### LINKED ISSUE #127: Duplicate resources in supported resources docs
Some resources such as `aws_alb / aws_alb` appear twice in https://github.com/infracost/infracost/blob/master/docs/generated/supported_resources.md

sinabakh:
Fix #127

### DIFF
--- a/docs/templates/supported_resources.md
+++ b/docs/templates/supported_resources.md
@@ -11,7 +11,7 @@ Support for the following is not currently included:
 
 | Terraform resource           | Notes |
 | ---                          | ---   |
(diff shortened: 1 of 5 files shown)


sinabakh:
@alikhajeh1 There seems to be a minor problem regarding the docs, my initial idea was to ignore the `docs/generated` from the git so contributors won't accidentally push to it but it's hard to ignore changes to an already tracked file.
It is not a major problem since even there be an accidental commit it will be overridden.

PR #139, opened 2020-10-01
fix: clarify infracost usage with adding example commands for help command

miladibra10:
Added example commands with the `UsageText` property of the cli app.


alikhajeh1:
Thanks @miladibra10!

Issue #160, opened 2020-10-08
Handle null_resource resources


sinabakh:
There are `null_resources` which are no-cost components. We can address this issue after #154

aliscott:
Added in #162.

PR #162, opened 2020-10-08
Add free resources

### LINKED ISSUE #154: Add free resources feature
In complex resources, there are tons of additional resources like iam policies and ... which have no cost but fill the skipped resources warning. It would be great if we can add a free resource concept.

sinabakh:
Fix #154 #160


sinabakh:
@aliscott I changed a small part of your code if you don't mind.

Also, I think we can merge it without stash, I tried to make the commits standalone enough.

sinabakh:
The changes seem to have created some side-effects. Will work on them.

alikhajeh1 (commented):
@sinabakh thanks! have you checked the list in AWS pricing pages to ensure they are all indeed free? I asked and @aliscott hadn't done that.

Can we possibly rename "NoCost" to "NoPrices" everywhere to clarify that these resources don't have any prices (it implies they have no cost too but it's clearer vs a usage-based resource that shows as 0 cost).

Also, can we delete the existing NoCost=true resources in the master branch as part of this PR so everything moves to using this array? e.g. `aws_launch_configuration`

alikhajeh1:
> @sinabakh thanks! have you checked the list in AWS pricing pages to ensure they are all indeed free? I asked and @aliscott hadn't done that.

I stand corrected, @aliscott had actually checked those.

alikhajeh1:
How should this work with the auto-generated supported_resources docs page? Should we show these in there as supported? I'm thinking we exclude them as it's a long list and I would guess that most users know they are free resources? @sinabakh / @aliscott any thoughts?

sinabakh:
@alikhajeh1 I agree, we need to move them to one place.
Regarding the docs, it's a matter of preference. Currently they are added to the docs but we can skip them. What you think @aliscott 
I'll try to finish this PR by the end of today.

aliscott:
@sinabakh I think we should exclude them from the supported_resources docs since there may be a lot of them and it will become messy.

sinabakh:
@alikhajeh1 @aliscott I made the following changes:
1. Changed the `noCost` name to `noPrice`
2. Skipped them in docs gen
3. Moved all free resources to the strings array
4. Tweaked the test utils for these resources

Please take a look, tests passed on my local env. But the conflicts should be resolved first to run the tests. Unfortunately I have no access for confilct resoloution so you have to do it.

sinabakh:
@aliscott Linter gave me a valid error which was not visible in master branch:
```
cmd/infracost/main.go:188:3: exitAfterDefer: os.Exit clutters `defer func(){...}(...)` (gocritic)
                os.Exit(1)
```
So I changed the main.go Please review it.

aliscott (approved):

PR #178, opened 2020-10-13
Fix aws dms region

### LINKED ISSUE #177: DMS pricing calculated on wrong region
I see support for DMS is added in #149 🙌  

When running `infracost v0.6.4` I see that DMS cost is calculated at $0.0720/hour, however, these resources run in `eu-central-1` which have a different price. 

* Price shown by infracost for `t2.small` (multi-az): `$0.072/hour` --> probably corresponds to price fetched from `us-east-1`?
* Price for `t2.small` (multi-az)  in `eu-central-1`: `$0.084/hour` (According to: https://aws.amazon.com/dms/pricing/)

image

However, I see `region` support in the `dms.go`, but my golang knowledge is zero to none to be able to debug it properly.

sinabakh:
Fix #177

### DIFF
--- a/internal/providers/terraform/aws/dms.go
+++ b/internal/providers/terraform/aws/dms.go
@@ -27,7 +27,7 @@ func getInstanceFamily(d *schema.ResourceData) string {
 }
 
 func instanceCostComponent(d *schema.ResourceData) *schema.CostComponent {
-	region := d.Get("region").String()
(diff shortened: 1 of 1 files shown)


alikhajeh1:
@sinabakh do you think we need to review other resources to see if the availability_zone causes similar issues?

PR #197, opened 2020-11-01
Update LB resources

alikhajeh1:


internal/providers/terraform/aws/dynamodb_table.go
```diff
@@ -311,7 +311,7 @@ func restoreCostComponent(d *schema.ResourceData, u *schema.ResourceData) *schem
 	}
 	return &schema.CostComponent{
 		Name:            "Restore data size",
-		Unit:            "GB",
+		Unit:            "GBs",
```
aliscott, line 314: I would say `GB` is already plural.

aliscott (approved):

internal/providers/terraform/aws/dynamodb_table.go
alikhajeh1: Good point, I updated the PR.

PR #203, opened 2020-11-05
Add --use-tfstate flag

aliscott:
This adds an `--use-tfstate` flag which allows the user to use the state instead of a plan of the Terraform code.
This works by running `terraform show -json` inside the Terraform directory so it works with both local and remote state.

### DIFF
--- a/internal/providers/terraform/parser.go
+++ b/internal/providers/terraform/parser.go
@@ -8,13 +8,26 @@ import (
 
 	"github.com/infracost/infracost/internal/schema"
 	"github.com/pkg/errors"
-
 	log "github.com/sirupsen/logrus"
+
 	"github.com/tidwall/gjson"
 )
 
 // These show differently in the plan JSON for Terraform 0.12 and 0.13.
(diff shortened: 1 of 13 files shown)


internal/providers/terraform/parser.go
```diff
@@ -8,14 +8,26 @@ import (
...
-
 	log "github.com/sirupsen/logrus"
+
 	"github.com/tidwall/gjson"
 )
 
 // These show differently in the plan JSON for Terraform 0.12 and 0.13.
 var infracostProviderNames = []string{"infracost", "registry.terraform.io/infracost/infracost"}
 
+var arnAttributeMap = map[string]string{
```
alikhajeh1, line 19: Would it help to add a comment here saying why this is needed?

alikhajeh1 (approved):
LGTM

tovbinm:
Thank you! Great addition.

tovbinm:
Q: How would one pass `--use-tfstatecode` in Github Actions to work work with Terraform Cloud?

PR #242, opened 2020-11-30
Add google_compute_address

### LINKED ISSUE #232: Add support for GCP IP Address
Pricing details: https://cloud.google.com/vpc/network-pricing#ipaddress
Terraform resource: google_compute_address

It might help to study the infracost/internal/providers/terraform/aws/eip.go file before adding this.

sinabakh:
Fix #232 

@aliscott This is the best I could do based on my discussion with @alikhajeh1 

It would be great if you can double-check the logic and coding style since it's been a while since I committed code to infracost.

### DIFF
--- a/internal/providers/terraform/google/compute_address.go
+++ b/internal/providers/terraform/google/compute_address.go
@@ -23,10 +23,6 @@ func GetComputeGlobalAddressRegistryItem() *schema.RegistryItem {
 func NewComputeAddress(d *schema.ResourceData, u *schema.ResourceData) *schema.Resource {
 	region := d.Get("region").String()
 
-	zone := d.Get("zone").String()
-	if zone != "" {
-		region = zoneToRegion(zone)
-	}
 	addressType := d.Get("address_type").String()
 	if addressType == "INTERNAL" {
 		return &schema.Resource{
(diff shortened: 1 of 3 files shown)


aliscott (commented):
@sinabakh thanks!

From what I can see from the Google pricing there are three potential price options:
 * Unused IP
 * Attached to standard VM
 * Attached to preemptible VM

Since we don't know what state the IP is in just now I think we should show all 3 potential prices here, with a nil quantity for each.

We are looking at adding better relationship mapping between resources as well, so once we have that we can tell if it's attached to a standard or preemptible VM and adjust this accordingly.

sinabakh:
@aliscott Agreed, so now it's like:
```
google_compute_address.static                                                                                                   
  ├─ Static and ephemeral IP addresses in use on standard VM instances               -  hours  0.0040            -             -  
  ├─ Static and ephemeral IP addresses in use on preemptible VM instances            -  hours  0.0020            -             -  
  └─ Static IP address (assigned but unused)                                         -  hours  0.0100            -             -  
  Total                                                                                                          -             - 
```

alikhajeh1:
How about these names:
```
google_compute_address.static                                                                                                   
  ├─ IP address (if used by standard VM)
  ├─ IP address (if used by preemptible VM)
  └─ IP address (if unused)
```

Can we also handle `google_compute_global_address` in this PR? Also, the registry.go file can be updated to remove the comments that mention these 2 resources.

sinabakh:
@alikhajeh1 Changed the names, also added the `google_compute_global_address` with the same resource logic since for now, they are the same.

alikhajeh1:
LGTM, have assigned to @aliscott as he did the initial review.

internal/providers/terraform/google/compute_address.go
```diff
@@ -0,0 +1,110 @@
...
+			NoPrice:   true,
+			IsSkipped: true,
+		}
+	}
+
+	return &schema.Resource{
+		Name: d.Address,
+		CostComponents: []*schema.CostComponent{
+			standardVMComputeAddress(region),
+			preemptibleVMComputeAddress(region),
```
aliscott, line 43: Looks like these two components don't need `region` passed in.

aliscott (commented):
One small comment, otherwise LGTM.

internal/providers/terraform/google/compute_address.go
aliscott: I pushed a commit to fix this.

PR #250, opened 2020-12-03
chore(ci): add test for report html

alikhajeh1:
@aliscott I think this test should pass after we fix the `open internal/output/output.html: no such file or directory` issue


alikhajeh1:
This is using make run so won't show the issue :(

PR #252, opened 2020-12-03
fix(output): fix HTML output so it doesn't load template from file

aliscott:


alikhajeh1 (approved):
we might want to merge https://github.com/infracost/infracost/pull/250 first, watch it fail the master build to ensure it catches this issue in the future, then pass after this one

Issue #57, opened 2020-09-03
[CI/CD] Docker releases should include app version and tf version


unfor19:
For example-

`infracost/infracost:1.0.1-tf0.12.26, infracost/infracost:1.0.1-tf0.13.1`

alikhajeh1:
This is partially done as we build/push released versions to dockerhub (all using Terraform v0.13.4).

unfor19:
@alikhajeh1 feel free to close this issue if it's irrelevant

alikhajeh1:
@unfor19 I think it might be needed long-term as more terraform versions (after the one that infracost supports) come out.

alikhajeh1:
Going to close this issue as the infracost Dockerfile now has the latest patch versions that are supported and users can use [this env](https://www.infracost.io/docs/integrations/environment_variables#infracost_terraform_binary) to define which they'd like to use.

PR #259, opened 2020-12-10
update terraform version on dockerfile to 0.14.2

bmbferreira:


alikhajeh1:
@bmbferreira thanks for the PR! I wonder if we should do something similar to [atlantis](https://github.com/runatlantis/atlantis/blob/master/Dockerfile#L9-L20) for the latest versions of 0.12, 0.13 and 0.14 (the infracost [supported versions](https://www.infracost.io/docs/faq#what-terraform-versions-are-supported)) so users can use the [TERRAFORM_BINARY env](https://www.infracost.io/docs/#environment-variables) to override the default.

Also, if we do that, should the default be the latest stable version of terraform or kept as is to avoid breaking CI pipelines? This image is used in the [ci integrations](https://www.infracost.io/docs/integrations) too.

@aliscott any thoughts?

aliscott:
> I wonder if we should do something similar to atlantis

Another option is to build images for each Terraform version to keep the image size smaller similar to how a lot of packages build for different versions of Alpine (like https://hub.docker.com/_/golang). For example:

`infracost:latest` - uses latest infracost and the latest Terraform
`infracost:v0.7.7` - uses infracost v0.7.7 and the latest Terraform
`infracost:latest-terraform0.14.2` - uses latest infracost and Terraform 0.14.2
`infracost:v0.7.7-terraform0.14.2` - uses infracost v0.7.7 and Terraform 0.14.2

alikhajeh1:
TF 0.12.29 is 67M and 0.14.0 is 78M uncompressed on Linux, so this would add ~150MB to the current 140MB image (from `docker images` on centos, it's 60MB compressed on dockerhub). Doubling the image size isn't great, it slows down CI integrations that don't cache the image. I think we prefer it if users use the latest version of infracost with whatever version of TF they need, so the multiple images sounds good.

alikhajeh1:
I'm going to merge this as I don't see any breaking changes in the [0.14.2 changelog](https://github.com/hashicorp/terraform/blob/v0.14.2/CHANGELOG.md). If required, we can continue the discussion on https://github.com/infracost/infracost/issues/57

bmbferreira:
Hi @alikhajeh1! I'm really sorry for the late response! I think it is a good idea to have multiple images, however I can't promise at this moment that I can take ownership of doing that change right now. 😞  If I find the time I would like to do it. Thanks for merging the PR and for all your work. 🙇

alikhajeh1:
@bmbferreira no problems at all, sorry that we kinda went overboard with the discussion on this 1-line simple change :)

PR #261, opened 2020-12-12
feat(output): add skipped resources message to report

aliscott:
This also updates the JSON output to show the unsupported resources in a structured format, e.g.
```json
  "resourceSummary": {
    "unsupportedCounts": {
      "aws_simpledb_domain": 3,
      "aws_docdb_cluster": 1
    }
  }
```

### DIFF
--- a/cmd/infracost/default.go
+++ b/cmd/infracost/default.go
@@ -54,13 +54,18 @@ func defaultCmd() *cli.Command {
 			},
 			&cli.BoolFlag{
 				Name:  "show-skipped",
-				Usage: "Show unsupported resources, some of which might be free",
+				Usage: "Show unsupported resources, some of which might be free (only for table and HTML output)",
 				Value: false,
 			},
 		},
 		Action: func(c *cli.Context) error {
 			if err := checkAPIKey(); err != nil {
 				return err
+			}
+
+			if c.String("output") == "json" && c.Bool("show-skipped") {
+				msg := color.YellowString("The --show-skipped option is not needed with JSON output as that always includes them\n")
+				fmt.Fprint(os.Stderr, msg)
 			}
 
 			provider := terraform.New()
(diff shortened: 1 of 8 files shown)


cmd/infracost/default.go
```diff
@@ -63,6 +63,11 @@ func defaultCmd() *cli.Command {
 				return err
 			}
 
+			if c.String("output") == "json" && c.Bool("show-skipped") {
+				msg := color.YellowString("The --show-skipped option has no effect with the JSON output\n")
```
alikhajeh1, line 67:
```suggestion
				msg := color.YellowString("The --show-skipped option is not needed with JSON output as that always includes them\n")
```
?

cmd/infracost/report.go
```diff
@@ -29,15 +31,25 @@ EXAMPLES:
...
 			},
+			&cli.BoolFlag{
+				Name:  "show-skipped",
+				Usage: "Show unsupported resources, some of which might be free (only for table and HTML output)",
+				Value: false,
+			},
 		},
 		Action: func(c *cli.Context) error {
+			if c.String("output") == "json" && c.Bool("show-skipped") {
+				msg := color.YellowString("The --show-skipped option has no effect with the JSON output\n")
```
alikhajeh1, line 42:
```suggestion
				msg := color.YellowString("The --show-skipped option is not needed with JSON output as that always includes them\n")
```

internal/output/output.go
```diff
@@ -140,58 +153,173 @@ func Combine(outs ...Root) Root {
...
 	}
 
-	supportedTypeCount := 0
-	for rType := range summary.UnsupportedCounts {
-		if terraform.HasSupportedProvider(rType) {
-			supportedTypeCount++
-		}
+	unsupportedTypeCount := len(*r.ResourceSummary.UnsupportedCounts)
+
+	pluralizedTypesMsg := "resource type"
```
alikhajeh1, line 183: Maybe we should just remove this logic and always use `resource types` as in the message we use `them` vs `it` (we use to use `it` for singular).

alikhajeh1 (approved):
LGTM

internal/output/output.go
aliscott: I've fixed it to show `them` vs `it` as well

PR #315, opened 2021-01-07
Add global AWS data transfer resource and load usage from usage file

aliscott:


internal/config/environment.go
```diff
@@ -29,6 +29,7 @@ type EnvironmentSpec struct {
 	TerraformVersion                    string   `json:"terraformVersion"`
 	TerraformRemoteExecutionModeEnabled bool     `json:"terraformRemoteExecutionModeEnabled"`
 	TerraformInfracostProviderEnabled   bool     `json:"terraformInfracostProviderEnabled"`
+	HasUsageFileResources               bool     `json:"hasUsageFileResources"`
```
alikhajeh1, line 32: just hasUsageFile?

alikhajeh1 (approved):
LGTM. Is this going to break things for users of the Infracost provider? If so, we should add deprecation notices and consider that for the version

alikhajeh1:
Once this is merged, we need to create PRs to update the contributing guide, the docs, and let the contributors know via the community slack channel.

aliscott:
> Is this going to break things for users of the Infracost provider?

This doesn't break the Infracost provider.

PR #320, opened 2021-01-09
Using given path instead of file base path

### LINKED ISSUE #309: The report command's HTML output ignores multiple files with the same name
`infracost report`'s HTML output seems to have a bug where it ignores multiple files with the same name, reproduction steps:
```sh
infracost --tfdir examples/terraform/ -o json > a.json
mkdir project1 && mv a.json project1
mkdir project2 && cp project1/a.json project2
infracost report -o html project1/a.json project2/a.json > report.html
cat report.html | grep -i file
# Notice how only 1 a.json file is in the HTML output, I expected 2 as there were 2 a.json files, one in each folder
```

sashman:
Fixes #309 

Using full given path rather than base file path

### DIFF
--- a/cmd/infracost/report.go
+++ b/cmd/infracost/report.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"strings"
 
 	"github.com/fatih/color"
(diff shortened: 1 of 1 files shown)


aliscott (approved):

aliscott:
Thanks @sashman 🚀

PR #331, opened 2021-01-16
docs: add note about price tier units

alikhajeh1:
@sinabakh can you possibly review this quick change to see if it makes sense?


sinabakh:
@alikhajeh1 LGTM.

PR #338, opened 2021-01-18
feat(aws): Adds support for AWS SSM 

### LINKED ISSUE #251: Add support for AWS Systems Manager's Parameter Store
**Resource details:**
 - Resource name: AWS Systems Manager's Parameter Store (used to store/access config data, e.g. DB creds) 
 - Link to pricing information: https://aws.amazon.com/systems-manager/pricing/ ONLY the Parameter Store pricing
 - Terraform resources: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssm_parameter

The parameter store is an interesting resource to add as it's often used in stacks for credentials management.

robh007:
**Objective:** closes #251 

Adds support for `ssm_parameter` `ssm_activation`

```
  NAME                                         MONTHLY QTY  UNIT              PRICE   HOURLY COST  MONTHLY COST

  aws_ssm_activation.advanced
  └─ On-prem managed instances (advanced)              730  hours             0.0069       0.0069        5.0735
  Total                                                                                    0.0069        5.0735

  aws_ssm_parameter.advanced
  ├─ Parameter storage (advanced)                      730  hours             0.0001       0.0001        0.0511
  └─ API interactions (advanced)                         -  10k interactions  0.0500            -             -
  Total                                                                                    0.0001        0.0511

```


internal/providers/terraform/aws/ssm_activation.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if u != nil && u.Get("instance_tier").Exists() {
+		tier = u.Get("instance_tier").String()
+	}
+
+	if u != nil && u.Get("instance_count").Exists() {
+		instanceCount = decimal.NewFromInt(u.Get("instance_count").Int())
+	}
+
+	if tier == "Standard" {
```
alikhajeh1, line 29: This can be deleted as we don't need to show free resources

internal/providers/terraform/aws/ssm_activation.go
```diff
@@ -0,0 +1,63 @@
...
+				},
+			},
+		}
+	}
+
+	return &schema.Resource{
+		Name: d.Address,
+		CostComponents: []*schema.CostComponent{
+			{
+				Name:           "On-Premises instance management - advanced",
```
alikhajeh1, line 47:
```suggestion
				Name:           "On-prem managed instances (advanced)",
```

internal/providers/terraform/aws/ssm_activation.go
```diff
@@ -0,0 +1,63 @@
...
+		Name:  "aws_ssm_activation",
+		RFunc: NewSSMActivation,
+	}
+}
+
+func NewSSMActivation(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := d.Get("region").String()
+
+	tier := "Standard"
+	instanceCount := decimal.Zero
```
alikhajeh1, line 19: How about this for the logic:
1. Don't show standard cost component as it's free.
2. If `registration_limit` in Terraform is set and it's > 1000 and `aws_ssm_activation.managed_instances` is not set, show the advanced price but no costs (i.e. quantity is nil as we don't know).
3. If `aws_ssm_activation.managed_instances` is set in infracost-usage.yml and it's > 1000, show advanced pricing with the value as quantity.

@aliscott any thoughts/ideas on if we should validate values? e.g. if the registration_limit is less than aws_ssm_activation.managed_instances, should we show an error to say it can't be more than the limit? if so, how should the code handle that? throw an exception seems too harsh?

internal/providers/terraform/aws/ssm_activation.go
```diff
@@ -0,0 +1,63 @@
...
+			},
+		}
+	}
+
+	return &schema.Resource{
+		Name: d.Address,
+		CostComponents: []*schema.CostComponent{
+			{
+				Name:           "On-Premises instance management - advanced",
+				Unit:           "Hours",
```
alikhajeh1, line 48:
```suggestion
				Unit:           "hours",
```

alikhajeh1 (changes requested):
@robh007 thanks! This is great, I left my suggestions, it was one of those resources that looks simple but quite hard to understand the pricing from AWS/Terraform, hence most of my comments are about what logic we could use to figure out the costs.

internal/providers/terraform/aws/ssm_parameter.go
```diff
@@ -0,0 +1,106 @@
...
+			UnitMultiplier: 1,
+			HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
+		}
+	}
+
+	return &schema.CostComponent{
+		Name:           "Parameter storage - advanced",
+		Unit:           "Hours",
+		UnitMultiplier: 1,
+		HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
```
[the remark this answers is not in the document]
robh007: @alikhajeh1 With this are you saying using MonthlyQuantity in number of hours. i.e. 

```
aws_ssm_parameter.my_parameter:
    monthly_parameter_storage_hours: 730 # Number of hours in the month the parameter will be stored for.
```

and default to 0 hours. I think if you do this the cost that is displayed by default is misleading as you will show a zero cost for the Advanced parameter & If you create the resource you would incur a cost of 0.05 for the month. The output below is shown.

```
 aws_ssm_parameter.advanced                                                                                 
  ├─ Parameter storage (advanced)                    0  Parameter-months  0.0001       0.0000        0.0000  

```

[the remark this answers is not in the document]
alikhajeh1: @robh007 sorry I didn't explain well, I meant this:
```
aws_ssm_parameter.my_ssm:
    parameters: 10 # Number of parameters stored, assumes that each parameter is stored for 1 month.
```

So if the user sets it to 1, we assume 730 hours.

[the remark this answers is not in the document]
robh007: @alikhajeh1 I've changed this to monthly & set the default value to 730. This way the cost is shown correctly without anyone providing any input. There's a parameter in the usage data to change the storage hours if someone needs/ wants to do that. I think the main cause of the confusion is that the unit cost price for SSM parameter storage is actually $0.00007 but the monthly pricing charge is actually $0.05. So the way I've implemented means if you create a parameter the default charge is shown, but if you want to only store a parameter for a number of hours/days this option is available. Please feel free to change if you don't agree. Thanks.

internal/providers/terraform/aws/ssm_activation.go
robh007: So with this the registration_limit can be below 100 if you want to take advantage of the Advanced tier features. Advanced tier isn't just for when you have more than 1000 instances. You need to switch to advanced tier if you want  the additional features i.e. Session manager or Windows application patching. This flag is set at the account/region within AWS. There isn't a TF resource so you don't know whether it's set or not. I think I've added logic to cover the scenarios.

alikhajeh1:
Thanks @robh007! I tweaked a few things in https://github.com/infracost/infracost/pull/338/commits/9c2f1fbac44501f4147b4bf5d8951ea2a9fd3af4 (the commit message explains).

Issue #317, opened 2021-01-08
Terragrunt is re-running initialization for plan and show steps


aliscott:
When using `TERRAFORM_BINARY=terragrunt` the terraform initialization is run 3 times: during the `init`, `plan` and `show` stages.

This means infracost takes a long time to complete. This has more of an effect now that terraform 14.x  will download the required providers each time (#274).

notnmeyer:
what terragrunt version? based on when you submitted this, you might be describing a bug fixed in https://github.com/gruntwork-io/terragrunt/releases/tag/v0.27.0

> Fix a bug with Terragrunt always re-running init.

aliscott:
Thanks @notnmeyer you are spot on 😄! Have tested with terragrunt v0.27.0 and this is resolved.

alikhajeh1:
I sent a [PR](https://github.com/infracost/infracost/pull/353/files) to update the Terragrunt version we ship in the infracost docker image. @notnmeyer if you use that docker image (it's mostly used for CI integrations), do you think there is a need for multiple terragrunt versions in the dockerfile? (we do [this](https://github.com/infracost/infracost/blob/master/Dockerfile#L17) for terraform versions, which is similar to what Atlantis does in their image).

notnmeyer:
i dont use the image, but i'm familiar with how Terragrunt and Atlantis play together. what youre doing looks like what i'd expect. for the most part, the specific TG version doesnt matter much to me.

alikhajeh1:
@notnmeyer thanks for the feedback! I'll keep bumping up the infracost image to use latest version of TG until someone screams at me then :)

Issue #274, opened 2020-12-16
Skip init if it has already run


aliscott:
**Description**
Currently `terraform init` is run each time infracost is run when using `--tfdir`. This is done to ensure the correct providers are installed.

With terraform 13.x this was okay because the providers were cached, but in terraform 14.x the providers are no longer cached and are reinstalled every time (See https://github.com/hashicorp/terraform/issues/27099 and https://github.com/hashicorp/terraform/issues/27155).

This means that running infracost is slow because every time it is redownloading the providers. We should detect if init has already been run and skip it if that is the case. The trade-off is that we will show an error if the plugins need updated.

sinabakh:
I would like to give it a try but I'm not a terraform's advanced user so it would be great if you could give me some ideas about how I can tackle this issue.

aliscott:
@sinabakh I can think of 3 options:

1. Simple check if `.terraform` dir exists
   *  We had this before (https://github.com/infracost/infracost/commit/985075d3d534a5de65fbc5559b3f9fa79ef5496f#diff-72a6035a570221d6612b04b5[…]c4efc7935dc1a972bb3a9ee55d7a4)
   * It doesn't detect if init needs to be re-run due to updated providers/modules
  
 2.  The same as this (https://github.com/gruntwork-io/terragrunt/blob/bfc313da5695195fae47ca8282e3f2732360b1e4/cli/cli_app.go#L736)
     * Also don't think this works if providers/modules need to be updated
     * Seems like fairly complicated logic
     
 3. When running the plan and show stages, check if it errors on `Error: Initialization required`. If it does, then we run init and re-run the plan/show.
     * Sort of similar to how we detect if we need to run a remote plan here: https://github.com/infracost/infracost/blob/master/internal/providers/terraform/provider.go#L247-L252
     * This will work even if providers/modules need updated.
     * If we go for this we probably want to still output a spinner line to the user to say that it's running init at this stage.
     
I think I'm leaning towards number 3 because it seems the most robust and not too difficult to implement. What do you think?

alikhajeh1:
@mjad-dazn just FYI that this is now available in the newly released 0.7.16 version

PR #347, opened 2021-01-20
Feat price api parallel requests

### LINKED ISSUE #332: Make infracost faster
A Terraform project with 700 resources takes the following times to run. Download [plan.txt](https://github.com/infracost/infracost/files/5824591/bigplan.txt) and rename to plan.json to try it.

- 4mins with `infracost --tfjson plan.json`, this is hitting the hosted Cloud Pricing API. I haven't yet timed how long it takes if it runs a local version - that would be the next thing to try. If the bottleneck is the CLI itself, we can try a few things such as send 10 parallel price lookup API calls at the same time? Currently we send requests serially.
- 23secs with `curl -X POST -H "x-api-key: my-api-key" -F "tfjson=@plan.json" https://pricing.api.infracost.io/tfjson`. This [just runs the CLI](https://www.infracost.io/docs/integrations#infracost-api) internally but it runs in the same server that the hosted Cloud Pricing API runs.

sinabakh:
Partial fix for #332 

Here are some benchmarks against the provided plan file in the issue:

Concurrency Level 1: 4mins & 7seconds -> 247 seconds
Concurrency Level 4: 1mins & 4seconds -> 64 seconds
Concurrency Level 8: 31 seconds
Concurrency Level 16(default on MacBook Pro 2017 with 4 cores): 24 seconds

Although there is a huge boost, I was expecting better results :)). When we change 1 to 4 there must be more than 4x improvement but it seems the network handshake or sth else is preventing it to happen. However, it's a good start.

@alikhajeh1 It would be great if you can run some tests on a more reliable network.

@aliscott @sashman Could you please take a look and give your comments and suggestions?

### DIFF
--- a/cmd/infracost/default.go
+++ b/cmd/infracost/default.go
@@ -62,6 +62,10 @@ func defaultCmd() *cli.Command {
 				Usage:     "Path to Infracost usage file that specifies values for usage-based resources",
 				TakesFile: true,
 			},
+			&cli.IntFlag{
+				Name:  "concurrency",
+				Usage: "Concurrency level of resources processing, defaults to max(4, number of cores * 4)",
+			},
 		},
 		Action: func(c *cli.Context) error {
 			if err := checkAPIKey(); err != nil {
@@ -70,6 +74,10 @@ func defaultCmd() *cli.Command {
 
 			config.Environment.Flags = c.FlagNames()
 			config.Environment.OutputFormat = c.String("output")
+			concurrency := c.Int("concurrency")
+			if concurrency < 1 {
+				return errors.New("concurrency level must be higher than 0. omit to fallback to default")
+			}
 
 			if c.String("output") == "json" && c.Bool("show-skipped") {
 				msg := color.YellowString("The --show-skipped option is not needed with JSON output as that always includes them\n")
@@ -88,7 +96,7 @@ func defaultCmd() *cli.Command {
 
 			spinner = spin.NewSpinner("Calculating cost estimate")
 
-			if err := prices.PopulatePrices(resources); err != nil {
+			if err := prices.PopulatePrices(resources, concurrency); err != nil {
 				spinner.Fail()
 
 				red := color.New(color.FgHiRed)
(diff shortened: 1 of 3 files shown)


aliscott:
@sinabakh this looks great 🚀 🚀 🚀  and is a huge improvements. How about something like this for the concurrency?
```
min: 4
default: cpu * 4
max: 16
```
I don't think we need to add a flag for this, probably better to just have sensible defaults.

I'm guessing the reason this didn't meet your expectations was because of the network overhead and for that we need to better batch the GraphQL queries.

alikhajeh1:
+1 on @aliscott's suggestion, as a user I only care about getting the results as fast as possible. It took ~25s for 700 resources on concurrency 16, which is great, ship it!
FYI the test failed with `concurrency level must be higher than 0. omit to fallback to default`, that's what I got when I ran it locally without a concurrency, so I think the test will fix it self once that flag is removed.

sinabakh:
Your suggestion seems legit to me, I will change it.
@alikhajeh1 The test failure was my fault, by removing the whole concurrency flag it will be fixed too.

sashman:
@alikhajeh1 @aliscott is there room to explore improving the API side?

alikhajeh1:
@sashman there is always room to explore! As @aliscott mentioned in [this comment](https://github.com/infracost/infracost/issues/332#issuecomment-761772178) batching the queries is something that could be explored too. Once this PR is merged, I think the CLI-to-Cloud Pricing API time will be good enough for now, however, https://github.com/infracost/infracost/issues/317 and https://github.com/infracost/infracost/issues/274 mean that the CLI is still slow'ish to start.

alikhajeh1:
@sinabakh just FYI I was chatting with @maskimko who mentioned that for their use-case, ~20s is too slow. I invited them to the community chat so we can keep bouncing ideas around on how to make the CLI faster.

sinabakh:
@alikhajeh1 I remember some more advanced solutions we discussed before.
The one I can think of is that currently, each resource is making a separate request. We probably can merge few resources request so we have less requests but it needs more investigation.

aliscott:
Yes currently we batch all cost components in a resource, whether the resource has 1 or 10 cost components. This isn't efficient if there's lots of resources that only one or two cost components. We could improve this by always batching the GraphQL requests for cost components in batches of size X instead of grouping by resource, and work out what the optimal X is.

PR #350, opened 2021-01-21
Add google cloud storage bucket

### LINKED ISSUE #231: Add support for GCP Cloud Storage
Pricing details: https://cloud.google.com/storage/pricing
Terraform resources: google_storage*

This can be a complicated resource to add so it might help to study how aws_s3_* ones were done in Infracost, e.g. https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/s3_bucket.go and join [our Discord](https://discord.gg/rXCTaH3) if you want to work on this.

sinabakh:
Fix #231


sinabakh:
Sample output:
```
 google_storage_bucket.my_storage_bucket                                                                                                                    
  ├─ Data storage                                                                                        150  GB-months   0.0001        3e-05        0.0200  
  ├─ Data retrieval                                                                                      500  GB          0.0200       0.0137       10.0000  
  ├─ Network egress                                                                              
  │  ├─ Data transfer in same continent                                                                  550  GB          0.0100       0.0075        5.5000  
  │  ├─ Data transfer to worldwide destinations (excluding Asia & Australia) (0-1 TB)                  1,024  GB          0.1200       0.1683      122.8800  
  │  ├─ Data transfer to worldwide destinations (excluding Asia & Australia) (1-10 TB)                 9,216  GB          0.1100       1.3887    1,013.7600  
  │  ├─ Data transfer to worldwide destinations (excluding Asia & Australia) (10+ TB)                  2,260  GB          0.0800       0.2477      180.8000  
  │  ├─ Data transfer to Asia Destinations (excluding China, but including Hong Kong) (0-1 TB)         1,024  GB          0.1200       0.1683      122.8800  
  │  ├─ Data transfer to Asia Destinations (excluding China, but including Hong Kong) (1-10 TB)          476  GB          0.1100       0.0717       52.3600  
  │  ├─ Data transfer to China Destinations (excluding Hong Kong) (0-1 TB)                                50  GB          0.2300       0.0158       11.5000  
  │  └─ Data transfer to Australia Destinations (0-1 TB)                                                 250  GB          0.1900       0.0651       47.5000  
  └─ Operations                                                                                  
     ├─ Class A                                                                                       40,000  operations   1e-05       0.0005        0.4000  
     └─ Class B                                                                                       20,000  operations   5e-06       0.0001        0.1000  
  Total                                                                                                                                2.1475    1,567.7000
```

@alikhajeh1 Any suggestions regarding the naming?

aliscott:
@sinabakh first of all, wow 🤩!

1. The network egress costs look like the same as the standard GCP egress costs (https://cloud.google.com/vpc/network-pricing#premium-tier). I think we should move these into these into a global resource called `google_network_egress` that should act similar to the `aws_data_transfer` resource. What do you think @sinabakh and @alikhajeh1? If you both agree, I think this should be moved to a separate PR.

2. I would also flatten the operations to:
    ```
    Class A operations / monthly_class_a_operations
    Class B operations / monthly_class_b_operations
    ```
    We need to add better contributing docs about when to use the hierarchy, but my initial thoughts is we should do it when the resource is a nested resource (block) in Terraform. But there's going to be exceptions to this, like CloudFront where without nesting it would be unreadable. @alikhajeh1 what do you think about this?

3. `monthly_storage_gb` -> `storage_gb` since the storage is continuous (number 5 here: https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md#usage-file-notes)

sinabakh:
@aliscott Your welcome. I will add my comments in reverse order since the first one is the most complicated one.

3. You are right, I'm not yet fully on-boarded on the new naming convention.

2. I modeled how the pricing page groups them. but the other way also seems completely ok. So it's a matter of taste. I will leave the decision to you and @alikhajeh1 

1. Regarding the first comment, I think it depends on how users expect from terraform. I personally want to get the cost of my storage bucket as a whole, so its network transfer is also part of the same component. Aggregating all the egress (the ones that have a direct relation with the resources) will make things a little complicated. Let's say I make a caching mechanism in my app, it won't change the egress cost but will lower the costs of buckets network data. I personally prefer this approach. However, since you interact more with users, what do you think about the user's expectations?

internal/providers/terraform/google/storage_bucket.go
```diff
@@ -0,0 +1,326 @@
...
+		quantity = decimalPtr(decimal.NewFromInt(u.Get("monthly_storage_gb").Int()))
+	}
+	storageClass := "STANDARD"
+	if d.Get("storage_class").Exists() {
+		storageClass = d.Get("storage_class").String()
+	}
+
+	region, resourceGroup := getDSRegionResourceGroup(location, storageClass)
+	return &schema.CostComponent{
+		Name:            "Data storage",
```
alikhajeh1, line 95:
```suggestion
		Name:            "Storage",
```

alikhajeh1 (approved):
I'd remove the word "destination" as it doesn't add much, I'd also keep the bit in brackets consistent with CloudFront et al, `(first 1TB), (next 9TB)`, I'd also remove the `&` and use `,` so it's easier to parse the output (we did that in CloudFront too). So this:

```
Data transfer to worldwide excluding Asia, Australia (first 1TB)                  1,024  GB          0.1200       0.1683      122.8800  
  │  ├─ Data transfer to worldwide excluding Asia, Australia (next 9TB)                 9,216  GB          0.1100       1.3887    1,013.7600  
  │  ├─ Data transfer to worldwide excluding Asia, Australia (over 10TB)                  2,260  GB          0.0800       0.2477      180.8000  
  │  ├─ Data transfer to Asia excluding China, including Hong Kong (first 1TB)         1,024  GB          0.1200       0.1683      122.8800  
  │  ├─ Data transfer to Asia excluding China, including Hong Kong (next 9TB)          476  GB          0.1100       0.0717       52.3600  
  │  ├─ Data transfer to China, excluding Hong Kong (first 1TB)                                50  GB          0.2300       0.0158       11.5000  
  │  └─ Data transfer to Australia (first 1TB)      
```

Regarding point 2, agree with @aliscott on flattening.

Regarding point 5, longer term we'll need both options, there is a use-case for users wanting to specify the per-resource data transfer amounts (usually for things that use data transfer a lot, like storage buckets or LBs), as well as some top-level data transfer block that covers everything else. I'd probably just start with the top-level one like we did for AWS as we can then extend that to either link/inline certain data transfers within resources. When we do that, we'd have to think about blending vs non-blending of costs (e.g. I sum-up all of your data transfer across all resources, take into account price tiers, and calculate a "blended" rate of your per-GB data transfer price, then apply that price to all of your usage - some users might want this, some users won't want this blending so it'll be configurable).

internal/providers/terraform/google/storage_bucket.go
sinabakh: It's done

sinabakh:
The final output:
```
  google_storage_bucket.my_storage_bucket
  ├─ Data storage                                                                           150  GB-months   0.0001        3e-05        0.0200
  ├─ Data retrieval                                                                         500  GB          0.0200       0.0137       10.0000
  ├─ Network egress
  │  ├─ Data transfer in same continent                                                     550  GB          0.0100       0.0075        5.5000
  │  ├─ Data transfer to worldwide excluding Asia, Australia (first 1TB)                  1,024  GB          0.1200       0.1683      122.8800
  │  ├─ Data transfer to worldwide excluding Asia, Australia (next 9TB)                   9,216  GB          0.1100       1.3887    1,013.7600
  │  ├─ Data transfer to worldwide excluding Asia, Australia (over 10TB)                  2,260  GB          0.0800       0.2477      180.8000
  │  ├─ Data transfer to Asia excluding China, but including Hong Kong (first 1TB)        1,024  GB          0.1200       0.1683      122.8800
  │  ├─ Data transfer to Asia excluding China, but including Hong Kong (next 9TB)           476  GB          0.1100       0.0717       52.3600
  │  ├─ Data transfer to China excluding Hong Kong (first 1TB)                               50  GB          0.2300       0.0158       11.5000
  │  └─ Data transfer to Australia (first 1TB)                                              250  GB          0.1900       0.0651       47.5000
  └─ Operations
     ├─ Class A                                                                          40,000  operations   1e-05       0.0005        0.4000
     └─ Class B                                                                          20,000  operations   5e-06       0.0001        0.1000
  Total                                                                                                                   2.1475    1,567.7000
```

sinabakh:
It will probably take some time to extract the egress and make it a separate resource. I will work on it ASAP.

internal/providers/terraform/google/registry.go
```diff
@@ -87,6 +88,16 @@ var FreeResources []string = []string{
 	"google_service_account_iam_member",
 	"google_service_account_iam_policy",
 	"google_service_account_key",
+	"google_storage_bucket_access_control",
+	"google_storage_bucket_acl",
+	"google_storage_bucket_iam_policy",
```
alikhajeh1, line 93: are we sure this resource is free? [their pricing page](https://cloud.google.com/storage/pricing#operations-pricing) shows `storage.*.setIamPolicy` as a class A operation. Maybe we ignore the resource as free here as it's already included in the bucket cost components so no action is needed. @aliscott any thoughts?

internal/providers/terraform/google/storage_bucket.go
```diff
@@ -0,0 +1,328 @@
...
+		"NEARLINE":       "NearlineOps",
+		"COLDLINE":       "ColdlineOps",
+		"ARCHIVE":        "ArchiveOps",
+	}
+
+	return []*schema.CostComponent{
+		{
+			Name:            "Class A operations",
+			Unit:            "operations",
+			UnitMultiplier:  10000,
```
alikhajeh1, line 124: I updated this and class B to UnitMultiplier=10K per the pricing page

alikhajeh1:
Thanks @sinabakh! I tweaked a few things and commented-out the network stuff until we're ready for that, here's the output:
```
  google_storage_bucket.my_storage_bucket
  ├─ Storage (coldline)                                              150  GB-months       0.0070       0.0014        1.0500
  ├─ Data retrieval                                                  500  GB              0.0200       0.0137       10.0000
  ├─ Object adds, bucket/object list (class A)                         4  10k operations  0.1000       0.0005        0.4000
  └─ Object gets, retrieve bucket/object metadata (class B)            2  10k operations  0.0500       0.0001        0.1000
  Total                                                                                                0.0158       11.5500
```

I added the storage class in brackets, and renamed the operations to use the same wording as the GCP prices do for their examples. The operations are also changed to use 10K unitMultiplier per the pricing page. Plz lmk if this looks ok and I'll merge it, I'll create a separate ticket for data transfer.

sinabakh:
@alikhajeh1 Thanks for the changes 👍🏼 They all look good to me. Regarding the free resources, yes that was the reason but I'm not sure if it's the right decision or not.

internal/providers/terraform/google/registry.go
aliscott: I agree with ignoring it here and including it in the bucket for now.

internal/providers/terraform/google/storage_bucket.go
```diff
@@ -169,7 +174,7 @@ func dataRetrieval(d *schema.ResourceData, u *schema.UsageData) *schema.CostComp
 	resourceGroup := storageClassResourceGroupMap[storageClass]
 	// Skipping standard, regional and multi-regional since they are free
 	if resourceGroup == "" {
-		return &schema.CostComponent{IgnoreIfMissingPrice: true}
```
alikhajeh1: @sinabakh FYI `IgnoreIfMissingPrice` means: look it up in the pricing API, if it's not there then don't add it. We use it for EBS Optimized instances since we don't know if they should have that cost component without looking it up. This doesn't seem like that use-case so I changed the code to not add the cost component if it's nil.

sinabakh: @alikhajeh1 I used nil first, but it raised a pointer exception. I thought it's a design choice. I think it's shipped with this bug, I will test it and report it as a bug.

sinabakh: @alikhajeh1 It was my mistake, it must have been some other problem. Thanks for fixing it.

Issue #349, opened 2021-01-21
Cost calculation differ between remote state and plan based calculation


dawidmalina:
I am prototyping infracost and I have faced strange issue. For some unknown reasons cost calculation are different between those two calculations:
- `infracost --no-color --tfdir . --tfplan ${PLAN}`
- `infracost --no-color --tfdir . --use-tfstate`

Terraform plan is empty:

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:

Terraform will perform the following actions:

Plan: 0 to add, 0 to change, 0 to destroy.
```

Diff between calculations:
```
11c11 
<   ├ Transit gateway attachment                                                              730  hours        0.0500       0.0500       36.5000    
---
>   ├ Transit gateway attachment                                                              730  hours        0.0600       0.0600       43.8000    
13c13
<   Total                                                                                                                     0.0500       36.5000   
---
>   Total                                                                                                                     0.0600       43.8000   
45c45
<   OVERALL TOTAL (USD)                                                                                                       0.1940      141.6200  
---
>   OVERALL TOTAL (USD)                                                                                                       0.2040      148.9200   
```

aliscott:
@dawidmalina thanks. This looks like it's detecting the wrong region for some reason. Do you have an example of the Terraform file you are using for this?

dawidmalina:
It is working environment but i will try to provide code example soon.

dawidmalina:
Additional input. I am using here nasted modules (maybe this will help)
```
.
├── emea (dedicated provider)
│   └── main.tf -> use vpc module
└── noram (dedicated provider)
    └── main.tf -> here i don't use vpc module
```

The module defines this tgw attachment like this:
```
resource "aws_ec2_transit_gateway_vpc_attachment" "vpc_vc01" {
  count              = var.tgw_id == null ? 0 : 1
  subnet_ids         = aws_subnet.vpc_vc01.*.id
  transit_gateway_id = var.tgw_id
  vpc_id             = aws_vpc.vpc_vc01.id

  tags =var.tags
}
```

aliscott:
I'm wondering if it's hitting this issue: https://github.com/infracost/infracost/blob/master/internal/providers/terraform/parser.go#L193-L195

Can you run with `INFRACOST_LOG_LEVEL=debug` and see it if it outputs "Falling back to default region"?

If you have an example of how the AWS region is passed into your modules that would be useful as well.

dawidmalina:
```
[...]

PR #351, opened 2021-01-21
fix(aws): Use EC2 Transit Gateway's region for the attachments

aliscott:
These resources don't return an ARN from the AWS API so we can't tell the region, but we can get it from their transit gateway


aliscott:
Fixes https://github.com/infracost/infracost/issues/349

alikhajeh1 (approved):
https://media.giphy.com/media/Hw8vYF4DNRCKY/giphy.gif 

I bet @dawidmalina will be happy to see this :D

dawidmalina:
i'm really appreciate your support and involvement with bug fixing. You are amazing!!!

alikhajeh1:
There are two parts to bug fixing (finding them and fixing them) - thanks you for helping us with the first part @dawidmalina!

PR #357, opened 2021-01-24
refactor(test): update how usage data is passed in integration tests

aliscott:


alikhajeh1 (approved):
LGTM, do we need to update the CONTRIBUTING.md file with details of test helpers? or there are plenty of examples so no need?

aliscott:
> do we need to update the CONTRIBUTING.md file with details of test helpers?
The contributing doc refers to NAT Gateway as an example which has this now.

PR #363, opened 2021-01-25
Run terraform init selectively

### LINKED ISSUE #274: Skip init if it has already run

sinabakh:
Fix #274

### DIFF
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -239,7 +239,8 @@ func runPlan(opts *CmdOptions, planFlags string, initOnFail bool) (string, []byt
 			config.Environment.TerraformRemoteExecutionModeEnabled = true
 			planJSON, err = runRemotePlan(opts, args)
 		} else if initOnFail == true && (strings.Contains(extractedErr, "Error: Could not load plugin") ||
-			strings.Contains(extractedErr, "Error: Initialization required")) {
+			strings.Contains(extractedErr, "Error: Initialization required") ||
+			strings.Contains(extractedErr, "Error: Module not installed")) {
 			spinner.Stop()
 			err = runInit(opts)
 			if err != nil {
--- a/internal/providers/terraform/provider.go
+++ b/internal/providers/terraform/provider.go
@@ -125,7 +125,7 @@ func (p *terraformProvider) generatePlanJSON() ([]byte, error) {
 	if p.planFile == "" {
 
 		var planJSON []byte
-		p.planFile, planJSON, err = runPlan(opts, p.planFlags)
+		p.planFile, planJSON, err = runPlan(opts, p.planFlags, true)
 		defer os.Remove(p.planFile)
 
 		if err != nil {
@@ -211,7 +211,7 @@ func runInit(opts *CmdOptions) error {
 	return nil
 }
 
-func runPlan(opts *CmdOptions, planFlags string) (string, []byte, error) {
+func runPlan(opts *CmdOptions, planFlags string, initOnFail bool) (string, []byte, error) {
 	spinner := spin.NewSpinner("Running terraform plan")
 	var planJSON []byte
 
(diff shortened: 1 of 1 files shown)


sinabakh:
It's the minimum working change, there are two obvious problems with this solution:
1. There might be other error messages that imply the need to run `init`
2. I think this recursive logic could cause some never-ending situations where the error is not resolved with running init. Maybe we can add a third flag to the function to handle just one level of recursion.

Just created the PR to know what you think regarding the changes and the 2 points mentioned. Am I on the right track?

aliscott:
@sinabakh this looks good to me. Agree with your idea for 2.

internal/providers/terraform/provider.go
```diff
@@ -234,14 +230,25 @@ func runPlan(opts *CmdOptions, planFlags string) (string, []byte, error) {
...
-
 	if err != nil {
+		extractedErr := extractStderr(err)
+
+		// If the plan returns this error then Terraform is configured with remote execution mode
+		if strings.HasPrefix(extractedErr, "Error: Saving a generated plan is currently not supported") {
+			log.Info("Continuing with Terraform Remote Execution Mode")
+			config.Environment.TerraformRemoteExecutionModeEnabled = true
+			planJSON, err = runRemotePlan(opts, args)
+		} else if initOnFail == true && (strings.Contains(extractedErr, "Error: Could not load plugin") ||
```
alikhajeh1, line 241:
```suggestion
		} else if initOnFail && (strings.Contains(extractedErr, "Error: Could not load plugin") ||
```
From CI's output:
 Error: S1002: should omit comparison to bool constant, can be simplified to `initOnFail` (gosimple)

sinabakh:
@alikhajeh1 Thanks for the quick fix. Not sure if you reviewed it completely or not but it's done from my side.

alikhajeh1:
Thanks, @aliscott mentioned he's going to test it locally (needs testing with remote execution mode too), I'll assign to him

aliscott:
This looks excellent.

One small thing I noticed was the glitch in the spinner changing from "Running terraform plan" to "Running terraform init". Other option would be to have:
```
"✖ Running terraform plan"
"✔ Running terraform init"
"⠹ Re-running terraform plan"
```
But I think the way you've done it is best. If you agree then let's leave as is.

I found 2 more cases that an init would be needed.

**Case 1**

When a provider version changes, i.e. change `examples/terraform/main.tf` to have:
```
provider "aws" {
  version = "3.24.0"
```

Error is:

```
    Error: Provider requirements cannot be satisfied by locked dependencies
    The following required providers are not installed:
    - registry.terraform.io/hashicorp/aws (3.24.0)
    Please run "terraform init".

Error running terraform plan: exit status 1
```

**Case 2**

When a module version changes, i.e. change `examples/terraform/main.tf` to have:
```
module "ec2-instance" {
  source                 = "terraform-aws-modules/ec2-instance/aws"
  version                = "2.16.0"
  name                   = "web-app"
  instance_count         = 1
  ami                    = "ami-674cbc1e"
  instance_type          = "t2.micro"
  subnet_id              = "subnet-eddcdzz4"
}
```
Run init, then change `version = "2.15.0"`

Error is:
```
    Error: Module version requirements have changed
      on main.tf line 11, in module "ec2-instance":
      11:   source                 = "terraform-aws-modules/ec2-instance/aws"
    The version requirements have changed since this module was installed and the
    installed version (2.16.0) is no longer acceptable. Run "terraform init" to
    install all modules required by this configuration.

Error running terraform plan: exit status 1
```

Once these are covered this is good to ship. I think this will have a huge impact on performance.

sinabakh:
@aliscott 
I personally prefer the spinner we have now. However, users won't notice that small glitch.

I added the new errors.

PR #364, opened 2021-01-25
Fix contributing file snippet

MohammadKarimi23:
The current version of snippet in CONTRIBUTING.md file leads to failure once build. 
This PR will fix the problem by using int64 variable for [NewFromInt function](https://github.com/shopspring/decimal/blob/v1.2.0/decimal.go#L93)


aliscott:
Thanks @MohammadKarimi23!

Issue #254, opened 2020-12-03
Add support for GCP Cloud DNS


alikhajeh1:
**Resource details:**
 - Resource name: GCP Cloud DNS
 - Link to pricing information: https://cloud.google.com/dns/pricing
 - Terraform resources: google_dns*

Issue #253, opened 2020-12-03
Add support for GCP projects/services


alikhajeh1:
**Resource details:**
 - Resource name: GCP project and service resources
 - Link to pricing information: I think these might be free but we need to check https://cloud.google.com/pricing/list
 - Terraform resources: google_project* and google_service_account*

alikhajeh1:
Fixed by https://github.com/infracost/infracost/pull/369

PR #369, opened 2021-01-26
feat(google): add Cloud DNS resources

alikhajeh1:
Closes https://github.com/infracost/infracost/issues/254 and https://github.com/infracost/infracost/issues/253
Also adds other free google resources and fixes a typo in aws_secretsmanager_secret

Example output:
```
  NAME                                               MONTHLY QTY  UNIT        PRICE   HOURLY COST  MONTHLY COST

  google_dns_managed_zone.example_zone
  └─ Managed zone                                              1  months      0.2000       0.0003        0.2000
  Total                                                                                    0.0003        0.2000

  google_dns_record_set.frontend
  └─ Queries                                                   1  1M queries  0.4000       0.0005        0.4000
  Total                                                                                    0.0005        0.4000

  OVERALL TOTAL (USD)                                                                      0.8079      589.8000
```

### DIFF
--- a/internal/providers/terraform/google/dns_managed_zone.go
+++ b/internal/providers/terraform/google/dns_managed_zone.go
@@ -1,0 +1,39 @@
+package google
+
+import (
+	"github.com/infracost/infracost/internal/schema"
+	"github.com/shopspring/decimal"
+)
+
+func GetDNSManagedZoneRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "google_dns_managed_zone",
+		RFunc: NewDNSManagedZone,
+	}
+}
+
+func NewDNSManagedZone(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	return &schema.Resource{
+		Name: d.Address,
+		CostComponents: []*schema.CostComponent{
+			{
+				Name:            "Managed zone",
+				Unit:            "months",
+				UnitMultiplier:  1,
+				MonthlyQuantity: decimalPtr(decimal.NewFromInt(1)),
+				ProductFilter: &schema.ProductFilter{
+					VendorName:    strPtr("gcp"),
+					Region:        strPtr("global"),
+					Service:       strPtr("Cloud DNS"),
+					ProductFamily: strPtr("Network"),
+					AttributeFilters: []*schema.AttributeFilter{
+						{Key: "description", Value: strPtr("ManagedZone")},
+					},
+				},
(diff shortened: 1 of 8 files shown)


internal/providers/terraform/google/dns_record_set.go
```diff
@@ -0,0 +1,45 @@
...
+						{Key: "description", Value: strPtr("DNS Query (port 53)")},
+					},
+				},
+				PriceFilter: &schema.PriceFilter{
+					StartUsageAmount: strPtr("0"),
+				},
+			},
+		},
+	}
+}
```
robh007, line 45: Would you not want to have the queries against the overall domain as appose to the individual records. Just thinking about how many records could be created. In this context if you generate a query usage per record set the output could get quite verbose. Domains could have lots of records. Maybe there's no right or wrong way.

internal/providers/terraform/google/dns_managed_zone.go
```diff
@@ -0,0 +1,39 @@
...
+					VendorName:    strPtr("gcp"),
+					Region:        strPtr("global"),
+					Service:       strPtr("Cloud DNS"),
+					ProductFamily: strPtr("Network"),
+					AttributeFilters: []*schema.AttributeFilter{
+						{Key: "description", Value: strPtr("ManagedZone")},
+					},
+				},
+				PriceFilter: &schema.PriceFilter{
+					StartUsageAmount: strPtr("0"),
```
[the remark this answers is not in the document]
alikhajeh1: Yeah I wanted to find a way to support it but each `dns_managed_zone` represents 1 hosted zone and it seemed odd to have two tiers for 1 zone as that zone itself has 1 price depending on the total number of domains. We might need something that counts the total number of zones and shows their cost in a combined way but we don't have that yet. So I went with ["Most expensive price tier is used."](https://github.com/infracost/docs/pull/39/files#diff-d26623f10c55215e3a0d65a5a799124e4b9349a3034d7f16ea606a91f677dc6cR48). We need to revisit these tiered resources when we have a better way to handle this.

internal/providers/terraform/google/dns_record_set.go
alikhajeh1: I initially put the queries cost component in the dns_managed_zone as you mentioned, but @aliscott pointed out that users might have the zone managed outside of terraform (this is why he had split the route_53 zone and query pricing). I'm ok changing this (and the route_53 resource) but leaning towards leaving it as is as so far we've tried to map each TF resource to a set of prices 🤷

PR #370, opened 2021-01-26
Add GKE support

aliscott:


internal/providers/terraform/google/container_cluster.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/shopspring/decimal"
+)
+
+func GetContainerClusterRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "google_container_cluster",
+		RFunc: NewContainerCluster,
+		Notes: []string{
+			"Sustained use discounts are applied to monthly costs, but not to hourly costs.",
+			"Costs associated with non-standard Linux images, such as Windows and RHEL are not supported.",
```
alikhajeh1, line 19: Are Windows/RHEL prices available in the Cloud Pricing API? Not suggesting any changes, just wanted to know in case users ask.

internal/providers/terraform/google/container_cluster.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	subResources := make([]*schema.Resource, 0)
+
+	if !d.Get("remove_default_node_pool").Bool() {
+		zones := int64(zoneCount(d.RawValues, ""))
+
+		countPerZone := int64(3)
+
+		if d.Get("initial_node_count").Type != gjson.Null {
```
alikhajeh1, line 40: why not `d.Get("initial_node_count").Exists()`? to catch cases where it exists but is an empty string?

alikhajeh1 (approved):
awesome!

internal/providers/terraform/google/container_cluster.go
aliscott: `.Exists()` is a weird one. It only checks if the value is present in the JSON, which it is always is in this case (https://github.com/tidwall/gjson/issues/104). It doesn't check for blank values. Checking `.Type != gjson.Null` checks for non-existent and null values, which catches this case okay - this one is a number so don't get empty strings.

internal/providers/terraform/google/container_cluster.go
aliscott: Yeah the license prices are in the Cloud Pricing API. To add them we'd need to somehow map the images used by the instance to the license that applies.

internal/providers/terraform/google/container_cluster.go
alikhajeh1: We have other places that do `if d.Get("iops").Exists() && d.Get("iops").Type != gjson.Null` - just checking the `.Exists()` isn't needed?

PR #373, opened 2021-01-28
Add better flag checking

aliscott:


internal/providers/terraform/provider.go
```diff
@@ -48,10 +48,22 @@ func (p *terraformProvider) ProcessArgs(c *cli.Context) error {
 	p.planFlags = c.String("tfflags")
 	p.usageFile = c.String("usage-file")
 
+	if p.useState && (p.jsonFile != "" || p.planFile != "") {
```
aliscott, line 51: Wasn't sure if this one should be an error or not. By default it will ignore the `--tfjson` or `--tfplan` args which seems a little odd.

alikhajeh1 (approved):

PR #374, opened 2021-01-28
Add support for Google PubSub resources

alikhajeh1:
Google pubsub has 1 price that applies for both the ingestion and delivery of messages. I show the same price with a slightly different name in the subscription and topic resources. I only show the retained and snapshot prices in the subscription as they apply on a per-subscription basis so it seems best to only show them there.

```
  NAME                                          MONTHLY QTY  UNIT         PRICE    HOURLY COST  MONTHLY COST
  google_pubsub_subscription.my_subscription
  ├─ Message delivery data                            7.416  TiB          40.0000       0.4064      296.6400
  ├─ Retained acknowledged message storage              605  GiB-months    0.2700       0.2238      163.3500
  └─ Snapshot message backlog storage                  70.6  GiB-months    0.2700       0.0261       19.0620
  Total                                                                                 0.6562      479.0520

  google_pubsub_topic.my_topic
  └─ Message ingestion data                           7.416  TiB          40.0000       0.4064      296.6400
  Total                                                                                 0.4064      296.6400
```

My initial attempt (see first commit) went down a rabbit hole of trying to model the costs using message sizes, retention windows etc but it got too complicated, and might confuse users who want to enter values based on their last week/month's usage data.


aliscott (approved):

aliscott:
LGTM!

PR #379, opened 2021-01-31
chore: remove generate-docs command

alikhajeh1:
We now maintain https://www.infracost.io/docs/supported_resources as the source of truth, the generate-docs command was also muddying the commit history

@sinabakh can you please review?


sinabakh:
@alikhajeh1 LGTM, by "We now maintain ... as the source of truth" you mean we are editing it manually for each release?

alikhajeh1:
thanks @sinabakh! yep, I maintain [this page](https://www.infracost.io/docs/supported_resources/) as each new release includes only a few resources so it's quick, and I want to ensure the Notes column is kept uptodate etc.

PR #392, opened 2021-02-04
Feat diff functionality in JSON output

### LINKED ISSUE #99: Adding diff functionality to JSON output
### Current aim
The aim of this ticket is to update the Infracost JSON output so it includes the following fields, which can then be used in the future to build table/HTML outputs that show the diff of what's about to be applied.

```
existingState:
  resources:
    ...
  totalHourlyCost
  totalMonthlyCost

plannedState:
  resources:
    ...
  totalHourlyCost
  totalMonthlyCost

diff:
  resources:
    ...
  totalHourlyCost
  totalMonthlyCost

timeGenerated
resourceSummary
``` 

### Original ticket description

Options to support:
- tfplan
- tfjson
- tfdirs
- tfstate (after #90)

So it would be 4*4=16 possible combinations to invoke it.

User can invoke it like
```
infracost diff --tfdir first/path --tfdir second/path
infracost diff --tfdir first/path --tfplan second/plan
...
```

So the flag presence order would be important. What other kind of implementations would work better??

sinabakh:
Fix #99


sinabakh:
It's currently a POC of what could be done, there is no diff functionality but the new structure and output are applied with backward compatibility.

@aliscott  @alikhajeh1  It would be great if you could take a look at what is done before I go any deeper so we can minimize the re-work and deliver as soon as possible.
Do you agree with the whole structure and where it is going? Also, do you have any suggestions regarding the naming of new entities?

internal/schema/resources_state.go
```diff
@@ -0,0 +1,6 @@
+package schema
+
+// ResourcesState is the state of a list of resources.
+type ResourcesState struct {
```
aliscott, line 4: Should this match the struct name in `output` (`ResourceState`)?

Also, I think this should have fields for `TotalHourlyCost` and `TotalMonthlyCost`, etc.

internal/schema/resource.go
```diff
@@ -23,9 +23,11 @@ type Resource struct {
 	Tags           map[string]string
 }
 
-func CalculateCosts(resources []*Resource) {
-	for _, r := range resources {
-		r.CalculateCosts()
+func CalculateCosts(state *State) {
+	for _, resources := range [][]*Resource{state.ExistingState.Resources, state.PlannedState.Resources} {
```
aliscott, line 27: Maybe add a `.AllResources()` method to State.

aliscott (commented):
Looks great so far 👌 👌. Just a heads up: I'm working on some changes that touch a lot of the codebase and will almost certainly conflict with this - but don't worry I'll help handle any conflicts when they come up.

internal/schema/resources_state.go
sinabakh: Yes, that's the output representation of this struct.

I was thinking about that, since we used to calculate the hourly and monthly costs on the output phase I didn't do it but if you agree I think it could be a field in this struct and be filled after the cost calculation phase and before generating the output.

internal/schema/resource.go
sinabakh: Makes total sense 👍🏼

internal/schema/resources_state.go
aliscott: Ah right, yeah make sense. I think your suggestion of doing it at the calculation phase is good.

sinabakh:
The last commit adds the actual functionality of the diff, but the code is not in its best form. I had to refactor the code over 3-4 times and still believe it's not as elegant as it should be. I will work on adding the final touches, heavy commenting, and completing unit tests in the coming 1-2 days.

I highly suggest against reviews in this stage.

However, here is a sample Infracost JSON output that you can take a look at for suggestions:
https://pastebin.com/iFuX6U4A

internal/providers/terraform/parser.go
```diff
@@ -78,21 +78,18 @@ func (p *Parser) createResource(d *schema.ResourceData, u *schema.UsageData) *sc
...
+	var vals gjson.Result
+	if parseExisting {
+		vals = parsed.Get("prior_state.values.root_module")
+		// TODO: Does the key differ in state file?
+	} else {
+		vals = parsed.Get("planned_values.root_module")
+		if !vals.Exists() {
+			vals = parsed.Get("values.root_module")
+		}
 	}
```
aliscott, line 93: > // TODO: Does the key differ in state file?

Yes the `values.root_module` is used in the state file. So should be like this:

```suggestion
	if parseExisting {
		vals = parsed.Get("prior_state.values.root_module")
		if !vals.Exists() {
			vals = parsed.Get("values.root_module")
		}
	} else {
		vals = parsed.Get("planned_values.root_module")
	}
```

aliscott:
This is looking solid. I'm still testing it but here's some initial thoughts after having played with it a bit.

**1. Output when all resources are new (no existing-state)**

Currently shows:

```json
  "existingState": {
    "resources": [],
    "totalHourlyCost": null,
    "totalMonthlyCost": null
  },
```

I think it's better if we change the costs to 0 to make it easier to consume, e.g:

```json
  "existingState": {
    "resources": [],
    "totalHourlyCost": 0,
    "totalMonthlyCost": 0
  },
```

**2. Output when all resources are removed**

Same as above here for the costs in `plannedState`.

**3. When using `--terraform-use-state` flag**

In this case there is no plannedState. We can get the current state from `values.root_module` (see https://github.com/infracost/infracost/pull/392/files#r574035259). I think we should set the `plannedState` and `diff` keys to null or remove them from the JSON. Let me know if I can help test/reproduce this.

This is what happens currently:

```
  "resources": {
    # RESOURCES
  },
  "existingState": {
    "resources": [],
    "totalHourlyCost": null,
    "totalMonthlyCost": null
  },
  "plannedState": {
    # RESOURCES
  }
  "diff": {
     # RESOURCES
  }
```

And here's what I think it should be

```
  "resources": {
    # RESOURCES
  },
  "existingState": {
    # RESOURCES
  },
  "plannedState": null
  "diff": null
```

**4. Table display logic**

This should be the same logic as for the to-be-deprecated `resources` key in the JSON to be backwards-compatible for now. Basically it should use `plannedState` to generate the table. If the `plannedState` is null (i.e. if the user has used `--terraform-use-state`) then it should used `existingState` to build the table.

sinabakh:
Happy to hear that. I'm a bit worried about some edge cases that might cause nil exceptions so your tests are really appreciated.
Regarding your comments:

**1. Output when all resources are new (no existing-state)**
That makes total sense, I will fix it tomorrow. Tbh, I will try to review other parts to see if there is a need for changes likes this so we can discuss it.

**2. Output when all resources are removed**
Same as above

**3. When using --terraform-use-state flag**
Ok, it seems a bit more complicated. Here are my ideas about each part of your suggestion:
* setting `plannedState` to null: I see your point that there is no actual planned state. But it will make our logic so complex. In current implementation what is PRESENT is always called the planned state and the prior state is called existing. So it might be better if we set `existingState` to null.
* nulling `diff` and `existingState/plannedState`: I think it's a tradeoff between preferring nulls or empty state (as it is). I would really like to know your preference regarding it. I generally prefer empty state but in this case, the problem is it implies that there was no prior state whereas there might actually be but we don't know of it because of the input file structure. So we have to somehow save this info on our output to be able to notify users. So I'm ok with nulling them.

PR #393, opened 2021-02-04
feat: add Atlantis integration script

alikhajeh1:


aliscott (approved):
LGTM

magichair:
❤️  Looks good, yeah the first tricky case I hit was when default branch plan files don't exist yet (a new directory on the "current branch") but it looks like you handle that if the awk of the output is empty or 0.

Issue #360, opened 2021-01-25
Add support for Google KMS


aliscott:
Resource name: Google KMS
Link to pricing information: https://cloud.google.com/kms/pricing
Terraform resources: google_kms*

ekifel:
Hello! I want to pick it up :)

alikhajeh1:
Released in v0.7.20!

PR #398, opened 2021-02-08
Adding support for Google KMS resource

ekifel:
Issue #360 

```
  NAME                                          MONTHLY QTY  UNIT            PRICE   HOURLY COST  MONTHLY COST
  google_kms_crypto_key.my_keys                                                                                 
  ├─ Key versions                                    10,000  months          0.0600       0.8219      600.0000  
  └─ Operations                                         100  10k operations  0.0300       0.0041        3.0000  
  Total                                                                                   0.8260      603.0000  
```


internal/providers/terraform/google/kms_crypto_key.go
```diff
@@ -0,0 +1,184 @@
...
+				},
+			},
+			PriceFilter: &schema.PriceFilter{
+				EndUsageAmount: strPtr("2000"),
+			},
+		})
+
+		if tierTwoKeys.GreaterThan(decimal.NewFromInt(0)) {
+			costComponents = append(costComponents, &schema.CostComponent{
+				Name:            keyDescript,
```
alikhajeh1, line 76:
```suggestion
			Name:            "Key versions (over 2K)",
```

Without this it wasn't clear :)

internal/providers/terraform/google/kms_crypto_key.go
```diff
@@ -0,0 +1,184 @@
...
+				Unit:            "keys",
```
alikhajeh1, line 77:
```suggestion
				Unit:            "months",
```

internal/providers/terraform/google/kms_crypto_key.go
```diff
@@ -0,0 +1,184 @@
...
+					},
+				},
+				PriceFilter: &schema.PriceFilter{
+					StartUsageAmount: strPtr("2000"),
+				},
+			})
+		}
+	} else {
+		costComponents = append(costComponents, &schema.CostComponent{
+			Name:            keyDescript,
```
alikhajeh1, line 96: same comment on name and unit as above

internal/providers/terraform/google/kms_crypto_key.go
```diff
@@ -0,0 +1,184 @@
...
+				ProductFamily: strPtr("ApplicationServices"),
+				AttributeFilters: []*schema.AttributeFilter{
+					{Key: "description", ValueRegex: strPtr("/" + keyDescript + "/")},
+				},
+			},
+		})
+	}
+
+	costComponents = append(costComponents, &schema.CostComponent{
+		Name:            operationDesctipt,
```
alikhajeh1, line 113:
```suggestion
		Name:            "Operations",
```

internal/providers/terraform/google/kms_crypto_key_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/infracost/infracost/internal/testutil"
+)
+
+func TestKMSCryptoKey(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode")
+	}
+
+	tf := `
+				resource "google_kms_crypto_key" "my_keys" {
```
alikhajeh1, line 17: the indentation looks odd here, we can treat these TF code snippets like other code when indenting

infracost-usage-example.yml
```diff
@@ -170,6 +171,10 @@ resource_usage:
   google_dns_record_set.my_record_set:
     monthly_queries:  1000000 # Monthly DNS queries.
 
+  google_kms_crypto_key.my_keys:
+    monthly_key_versions: 10000 # Number of uses crypto key versions per month
```
alikhajeh1, line 175:
```suggestion
    key_versions: 10000 # Number of key versions.
```

From the contributing guide's "Usage file notes" section, number 5 would apply here since each key version is used continuously in the month. You'll need to tweak the resource go file to use the new name too.
5. For resources that are continuous in time, do not use prefixes, e.g. use `instances`, `subscriptions`, `storage_gb`. For non-continuous resources, prefix with `monthly_` so users knows what time interval to estimate for, e.g. `monthly_log_lines`, `monthly_requests`.

infracost-usage-example.yml
```diff
@@ -170,6 +171,10 @@ resource_usage:
...
+    monthly_key_operations: 1000000 # Number of operations with crypto keys per month
```
alikhajeh1, line 176:
```suggestion
    monthly_key_operations: 1000000 # Monthly number of key operations.
```

alikhajeh1 (approved):
@ekifel this was an awesome PR! it's clear that you've read the contributing guide in details. I left my suggestions :)

internal/providers/terraform/google/kms_crypto_key.go
```diff
@@ -0,0 +1,184 @@
...
+				tierOneKeys = tierLimit
+				tierTwoKeys = monthlyKeys.Sub(tierLimit)
+			} else {
+				tierOneKeys = *monthlyKeys
+			}
+		}
+
+		costComponents = append(costComponents, &schema.CostComponent{
+			Name:            keyDescript + " (first 2K)",
+			Unit:            "keys",
```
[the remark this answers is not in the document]
aliscott: Sounds look a good idea.

ekifel:
@alikhajeh1 this is ready for review/merge

internal/providers/terraform/google/kms_crypto_key.go
```diff
@@ -0,0 +1,183 @@
...
+
+	algorithm := "GOOGLE_SYMMETRIC_ENCRYPTION"
+	protectionLevel := "SOFTWARE"
+
+	if d.Get("version_template").Type != gjson.Null {
+		algorithm = d.Get("version_template.0.algorithm").String()
+		protectionLevel = d.Get("version_template.0.protection_level").String()
+	}
+
+	monthlyKeys := decimal.Zero
```
alikhajeh1, line 31: This should default to `*decimal.Decimal`, since if usage-file.key_versions is not set AND rotation_period is not set (as it's optional in terraform), then this is shown:
```
resource "google_kms_crypto_key" "my_keys" {
  name            = "crypto-key-example"
  key_ring        = ""
}

results in:

  google_kms_crypto_key.my_keys
	├─ Key versions                                        0  months          0.0600       0.0000        0.0000
```
But `-` should be shown for the QTY and costs as we don't know the usage, so shouldn't guess the cost.

alikhajeh1 (changes requested):
Thanks for this! Just one tiny issue left before we can merge

internal/providers/terraform/google/kms_crypto_key.go
alikhajeh1: fixed this in the latest commit

alikhajeh1:
See https://github.com/infracost/infracost/pull/398/commits/059237d1b2b051eb2c66897081544310f597655f for my changes

PR #407, opened 2021-02-09
fix(terraform): parse region from values if exists

aliscott:
S3 buckets ARNs look like `arn:aws:s3:::bucket1` so they don't include a region key. In this case S3 buckets also pass a "region" value in the response so we can parse that out and use it as the region.


alikhajeh1 (approved):
Love a good inconsistency :D

PR #426, opened 2021-02-15
Feat: ECS on EC2

### LINKED ISSUE #79: Add support for ECS with EC2 launch type
Infracost currently supports ECS with Fargate launch type, but not EC2 launch type.

sinabakh:
Fix #79

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -158,6 +158,9 @@ resource_usage:
     monthly_log_lines: 5000000          # Monthly number of real-time log lines.
     custom_ssl_certificates: 3          # Number of dedicated IP custom SSL certificates.
 
+  aws_autoscaling_group.my_asg:
+    instances_count: 15 # Number of instances in the autoscaling group.
+  
   #
   # Terraform GCP resources
   #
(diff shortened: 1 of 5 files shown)


infracost-usage-example.yml
```diff
@@ -158,6 +158,9 @@ resource_usage:
     monthly_log_lines: 5000000          # Monthly number of real-time log lines.
     custom_ssl_certificates: 3          # Number of dedicated IP custom SSL certificates.
 
+  aws_autoscaling_group.my_asg:
+    instances_count: 15 # Number of instances in the autoscaling group.
```
aliscott, line 162: We use `instances` elsewhere and in the contributing docs (https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md#usage-file-notes), so I think we should keep this consistent.

aliscott (approved):
Looks good! Thanks @sinabakh.

infracost-usage-example.yml
sinabakh: Unfortunately, I didn't exactly get your suggestion. I believe `count` itself does not make much sense in this use case.

aliscott: `instances_count` -> `instances`

We use `instances` elsewhere as well.

sinabakh: Ahh sorry, my bad. Will fix it.

PR #436, opened 2021-02-17
feat(google): fix free IAM resources

alikhajeh1:
fix(google): these resources were mentioned on their parent page
e.g. there is no google_kms_key_ring_iam resource, there are three:
google_kms_key_ring_iam_binding
google_kms_key_ring_iam_member
google_kms_key_ring_iam_policy

@ekifel can you plz review this?


ekifel (approved):
@alikhajeh1 , I checked it, everything correct

PR #442, opened 2021-02-18
fix(flags): Adding no-color support to report

lastlee:
`--no-color` flag on `infracost report` returns the following error:

```
flag provided but not defined: -no-color
```

This fixes an issue with the configuration option to behave like `--show-skipped`.


aliscott:
Thanks @lastlee!

Issue #444, opened 2021-02-18
AWS Resource free of charge


dawidmalina:
Terraform resources:
- aws_resourcegroups_group

We can make it supported to remove the noise.

dawidmalina:
Hehe @alikhajeh1 you were quicker ;)

alikhajeh1:
Done via https://github.com/infracost/infracost/pull/445/files (sharing the link in case you prefer to send a PR if it saves you time vs creating the issue first for adding free resources)

alikhajeh1:
@dawidmalina doh! sorry about that - I didn't see the open PR, you added it in **exactly the right place**. As the community gets bigger maybe we should update the contributing guide to encourage people to add a comment on issues they're working on to avoid that woopsie 😹

alikhajeh1:
Tweaked it [here](https://github.com/infracost/infracost/commit/efe69e72eaaae607278aa033b12a6bfee12833db)

PR #446, opened 2021-02-18
AWS Resource free of charge #444

dawidmalina:
Not sure is the only place to add free resources. I was fallowing up this change: https://github.com/infracost/infracost/pull/422/files


dawidmalina:
Already added by @alikhajeh1

Issue #451, opened 2021-02-19
google_storage_bucket returns no price for EU


alikhajeh1:
The following resource returns no prices, but according to [this](https://cloud.google.com/storage/pricing) the EU multi-region has prices:
```
  NAME                                                       MONTHLY QTY  UNIT            PRICE   HOURLY COST  MONTHLY COST

  google_storage_bucket.my_storage_bucket
  ├─ Storage (standard)                                                -  GiB-months      0.0000            -             -
  ├─ Object adds, bucket/object list (class A)                         -  10k operations  0.0000            -             -
  └─ Object gets, retrieve bucket/object metadata (class B)            -  10k operations  0.0000            -             -
  Total                                                                                                     -             -

  OVERALL TOTAL (USD)                                                                                  0.0000        0.0000
```

Terraform resource:
```
resource "google_storage_bucket" "test" {
  name          = "test"
  location      = "EU"
  force_destroy = false
}
```

CLI logs:
```
time="2021-02-19T08:28:44Z" level=debug msg="Falling back to default region (us-central1) for google_storage_bucket.my_storage_bucket"
time="2021-02-19T08:28:44Z" level=debug msg="Loading usage data from Infracost provider resources"
time="2021-02-19T08:28:44Z" level=info msg="starting: Calculating cost estimate"
time="2021-02-19T08:28:44Z" level=debug msg="Getting pricing details from https://pricing.api.infracost.io/graphql for google_storage_bucket.my_storage_bucket"
time="2021-02-19T08:28:45Z" level=warning msg="No products found for google_storage_bucket.my_storage_bucket Storage (standard), using 0.00"
time="2021-02-19T08:28:45Z" level=warning msg="Multiple prices found for google_storage_bucket.my_storage_bucket Object adds, bucket/object list (class A), using the first price"
time="2021-02-19T08:28:45Z" level=warning msg="Multiple prices found for google_storage_bucket.my_storage_bucket Object gets, retrieve bucket/object metadata (class B), using the first price"
```

sinabakh:
I'm looking into it.

alikhajeh1:
Released in v0.8, migration guide is [here](https://www.infracost.io/docs/guides/v0.8_migration).

PR #456, opened 2021-02-19
feat: add google stroage bucket data transfer costs

### LINKED ISSUE #385: Add Google data transfer cost for Cloud Storage
The #350 PR already had some of this so the aim of this ticket is to move those to a separate resource like we did with AWS data transfer and add the requires usage-file params too

sinabakh:
Fix #385

### DIFF
--- a/internal/providers/terraform/google/storage_bucket.go
+++ b/internal/providers/terraform/google/storage_bucket.go
@@ -39,6 +39,9 @@ func NewStorageBucket(d *schema.ResourceData, u *schema.UsageData) *schema.Resou
 	return &schema.Resource{
 		Name:           d.Address,
 		CostComponents: components,
+		SubResources: []*schema.Resource{
+			networkEgress(d, u),
+		},
 	}
 }
 
--- a/internal/providers/terraform/google/storage_bucket.go
+++ b/internal/providers/terraform/google/storage_bucket.go
@@ -8,6 +8,17 @@ import (
 	"github.com/shopspring/decimal"
 )
 
+type storageBucketEgressRegionData struct {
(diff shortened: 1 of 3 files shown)


internal/providers/terraform/google/storage_bucket.go
```diff
@@ -194,142 +205,121 @@ func dataRetrieval(d *schema.ResourceData, u *schema.UsageData) *schema.CostComp
 	}
 }
 
-// TODO: Move to a separate resource, similar to what we did for AWS, also needs other data transfer prices
-
-// infracost-usage-example.yml for this
```
alikhajeh1, line 199: this should go back into that file

alikhajeh1:
Just FYI on this bug: https://github.com/infracost/infracost/issues/451

internal/providers/terraform/google/storage_bucket.go
sinabakh: A separate file for storage bucket or a separate file for whole google data transfer and a function for storage bucket?

alikhajeh1: sorry, I meant the infracost-usage-example.yml file should be updated as part of this PR to add the keys/descriptions (I had moved the values here to avoid confusing users)

sinabakh: Sure, I missed committing it.

alikhajeh1:
Thanks @sinabakh! I tweaked and merged this. It might be good if you take a quick look at the `func CalculateTierBuckets` in the master branch and see if this resource could use it? might help DRY some code.

Issue #472, opened 2021-02-25
Add support for Google Monitoring


sinabakh:
Pricing page:
https://cloud.google.com/stackdriver/pricing

ekifel:
@alikhajeh1 Hello, can I pick it?

alikhajeh1:
Thanks @ekifel 🕺

alikhajeh1:
@ekifel it might be good to look at the other 7 `google_monitoring_*` resources as part of this issue, we can still do 1 PR per resource but use the same GH issue.

alikhajeh1:
Released v0.8.0

PR #487, opened 2021-03-03
feat(google): Add support for Cloud Monitoring resource

ekifel:
Issue #472 

With usage-file: 
```
  NAME                                               MONTHLY QTY  UNIT      PRICE   HOURLY COST  MONTHLY COST
  google_monitoring_metric_descriptor.my_monitoring                                                            
  ├─ Monitoring data (first 100K)                        100,000  mebibyte  0.2580      35.3425   25,800.0000  
  ├─ Monitoring data (next 150K)                         150,000  mebibyte  0.1510      31.0274   22,650.0000  
  ├─ Monitoring data (over 250K)                         250,000  mebibyte  0.0610      20.8904   15,250.0000  
  └─ Monitoring API calls                              1,000,000  calls      1e-05       0.0137       10.0000  
  Total                                                                                 87.2740   63,710.0000 
```

Without usage-file: 
```
  NAME                                               MONTHLY QTY  UNIT      PRICE   HOURLY COST  MONTHLY COST
  google_monitoring_metric_descriptor.my_monitoring                                                              
  ├─ Monitoring data (first 100K)                              -  MiB-months  0.2580            -             -  
  └─ Monitoring API calls                                      -  calls        1e-05            -             -  
  Total                                                                                         -             -  
```


alikhajeh1:
@ekifel heads-up on my latest commit, I updated the resource to:
1. use MB for the units, I looked into this a while ago and GCP's MiB is the same as what most people think of when they see MB (it's very confusing as AWS use MiB to mean something different). Anyhow, the contributing guide's `Usage file notes` section mentions the units that we want to be used (so things are consistent).
2. added unit multiplier, to match the pricing page
3. removed the word "monitoring" from the monthly_api_calls in the usage-file as the resource has "monitoring" in it, so it's clear what the API calls are for, but "data" is generic so I kept that one as is.

PR #517, opened 2021-03-14
chore: remove duplicated `-y` for `apt-get` in Dockerfile

PeterDaveHello:
There were two `-y` for `apt-get` 😅


alikhajeh1:
Thanks @PeterDaveHello!

PR #532, opened 2021-03-24
Add 'post_condition' arg in CI script

ekifel:
Part of [Issue](https://github.com/infracost/infracost-gh-action/issues/22)

New argument: post_condition
Could be:
post_condition:'{"always":"true"}' - always post comments
post_condition:'{"has_diff":"true"}' - post comments if it has output
post_condition:'{"percentage_threshold":"2"}' - post comment if absolute_percent more than percentage_threshold

### DIFF
--- a/scripts/ci/atlantis_diff.sh
+++ b/scripts/ci/atlantis_diff.sh
@@ -14,7 +14,27 @@ if [ "$atlantis_debug" = "true" ] || [ "$atlantis_debug" = "True" ] || [ "$atlan
 fi
 
 process_args () {
+  # Validate post_condition
+  if ! echo "$post_condition" | jq empty; then
+    echo "Error: post_condition contains invalid JSON"
+  fi
+
   # Set defaults
+  if [ ! -z "$percentage_threshold" ] && [ ! -z "$post_condition" ]; then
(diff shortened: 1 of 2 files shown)


alikhajeh1:
@aliscott what were the steps you ran to test this script locally? 

@ekifel great start! here are my suggestions:
1. see `post_condition` as a new parameter so we don't break this for existing users.
2. if someone uses `percentage_threshold`, output this `Warning: The percentage_threshold parameter is deprecated and will be removed in v0.9.0, please use post_condition='{"percentage_threshold":"2"}'`.
3. if someone sets `percentage_threshold` && `post_condition`, then only use `post_condition` and output this `Warning: The percentage_threshold parameter is deprecated, using post_condition instead`.

Also, https://www.shellcheck.net/ might be useful too

aliscott:
> @aliscott what were the steps you ran to test this script locally?

I'm just running `scripts/ci/diff.sh examples/terraform` just now and then echoing out some vars.

alikhajeh1:
@ekifel sorry this took a while to review, testing it isn't quick... I've pushed a commit that fixes a few things, feel free to review.

PR #533, opened 2021-03-24
Sync usage data

### LINKED ISSUE #530: Auto-generate usage-file from Terraform project
Currently users have to write this out manually and everyone loves automation to help them.

sinabakh:
Fix #530

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -65,7 +65,7 @@ func runMain(cmd *cobra.Command, cfg *config.Config) error {
 
 		cfg.Environment.SetProjectEnvironment(provider.Type(), projectCfg)
 
-		u, err := usage.LoadFromFile(projectCfg.UsageFile)
+		u, err := usage.LoadFromFile(projectCfg.UsageFile, cfg.SyncUsageFile)
 		if err != nil {
 			return err
 		}
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -28,7 +28,7 @@ func addRunFlags(cmd *cobra.Command) {
 
 	cmd.Flags().Bool("show-skipped", false, "Show unsupported resources, some of which might be free")
 
-	cmd.Flags().Bool("sync-usage-file", false, "Experimental: Autofill the missing usage data and update the usage file")
+	cmd.Flags().Bool("sync-usage-file", false, "Autofill usage-file with missing resources (experimental)")
 }
 
 func runMain(cmd *cobra.Command, cfg *config.Config) error {
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -28,7 +28,7 @@ func addRunFlags(cmd *cobra.Command) {
 
 	cmd.Flags().Bool("show-skipped", false, "Show unsupported resources, some of which might be free")
 
-	cmd.Flags().Bool("sync-usage-file", false, "Autofill usage-file with missing resources (experimental)")
+	cmd.Flags().Bool("sync-usage-file", false, "Sync usage-file with missing resources, needs usage-file too (experimental)")
 }
 
 func runMain(cmd *cobra.Command, cfg *config.Config) error {
@@ -233,8 +233,7 @@ func loadRunFlags(cfg *config.Config, cmd *cobra.Command) error {
 
 func checkRunConfig(cfg *config.Config) error {
 	if cfg.Format == "json" && cfg.ShowSkipped {
-		ui.PrintWarning("The show skipped option is not needed with JSON output as that always includes them.\n")
-		return nil
+		ui.PrintWarning("show-skipped is not needed with JSON output format as that always includes them.\n")
 	}
 
 	return nil
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -77,7 +77,10 @@ func runMain(cmd *cobra.Command, cfg *config.Config) error {
 		}
 
 		projects = append(projects, project)
-		usage.SyncUsageData(project, u, projectCfg.UsageFile)
+		err = usage.SyncUsageData(project, u, projectCfg.UsageFile)
+		if err != nil {
+			return err
+		}
 
 		if !cfg.IsLogging() {
 			fmt.Fprintln(os.Stderr, "")
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -27,6 +27,8 @@ func addRunFlags(cmd *cobra.Command) {
 	cmd.Flags().String("terraform-workspace", "", "Terraform workspace to use. Applicable when path is a Terraform directory")
 
 	cmd.Flags().Bool("show-skipped", false, "Show unsupported resources, some of which might be free")
+
+	cmd.Flags().Bool("sync-usage-file", false, "Autofill the missing usage data and update the usage file")
 }
 
 func runMain(cmd *cobra.Command, cfg *config.Config) error {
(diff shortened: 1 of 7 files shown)


sinabakh:
The core functionality is ready!
Here's a sample scenario:
TF:
```
resource "google_storage_bucket" "my_storage_bucket" {
  ...
}

resource "google_storage_bucket" "my_storage_bucket_2" {
  ...
}
```

Initial usage data file:
```yml
version: 0.1
resource_usage:
  google_storage_bucket.my_storage_bucket:
    storage_gb: 150                   # Total size of bucket in GB.
    monthly_class_b_operations: 20000 # Monthly number of class B operations (object gets, retrieve bucket/object metadata).
    monthly_egress_data_transfer_gb:  # Monthly data transfer from Cloud Storage to the following, in GB:
      same_continent: 550  # Same continent.
  google_storage_bucket.my_storage_bucket_removed:
    storage_gb: 150                   # Total size of bucket in GB.
    monthly_class_a_operations: 40000 # Monthly number of class A operations (object adds, bucket/object list).
    monthly_data_retrieval_gb: 500    # Monthly amount of data retrieved in GB.
    monthly_egress_data_transfer_gb:  # Monthly data transfer from Cloud Storage to the following, in GB:
      same_continent: 550  # Same continent.
      worldwide: 12500     # Worldwide excluding Asia, Australia.
```

Updated usage data file using `--sync-usage-file` flag:
```yml
version: 0.1
resource_usage:
  google_storage_bucket.my_storage_bucket:
    monthly_class_a_operations: 0
    monthly_class_b_operations: 20000
    monthly_data_retrieval_gb: 0
    monthly_egress_data_transfer_gb:
      asia: 0
      australia: 0
      china: 0
      same_continent: 550
      worldwide: 0
    storage_gb: 150
  google_storage_bucket.my_storage_bucket_2:
    monthly_class_a_operations: 0
    monthly_class_b_operations: 0
    monthly_data_retrieval_gb: 0
    monthly_egress_data_transfer_gb:
      asia: 0
      australia: 0
      china: 0
      same_continent: 0
      worldwide: 0
    storage_gb: 0
```


it needs a way to load the reference usage example in the binary, more code reviews, real-world tests, and possibly a few changes. Here are some potential issues:

1. The usage file is overridden per project so if multiple projects use the same usage file then we will face data loss.

2. If the usage file is not existing, it will raise an error. We can possibly change it so if the `--sync-usage-file` is passed then an empty state file would be created.

3. The usage file gets overridden so it's easy to lose some WIP or uncommitted work.

sinabakh:
@aliscott I played with golang `embed`. The problem is it only supports files in the current package and its subdirectories. The only option we have is to copy the example file to the `internal/usage` directory when building the package. Or we have to find other ways of achieving the same result. I will try to see if there are other options available for us.

aliscott:
> 1. The usage file is overridden per project so if multiple projects use the same usage file then we will face data loss.

Good point. I wonder if we can detect this and add a "confirm overwrite" prompt. Probably not needed initially though.

> 2. If the usage file is not existing, it will raise an error. We can possibly change it so if the --sync-usage-file is passed then an empty state file would be created.

I think this case should create the file.

> 3. The usage file gets overridden so it's easy to lose some WIP or uncommitted work.

If we only add resources that have no existing data to the usage file does that solve this?

> The only option we have is to copy the example file to the internal/usage directory when building the package. Or we have to find other ways of achieving the same result.

Yeah, I think it should be fine to copy the file somewhere as part of the `make build` script. We will need to update the homebrew install as well though: https://github.com/Homebrew/homebrew-core/blob/master/Formula/infracost.rb#L19-L23

sinabakh:
@aliscott
1. Agreed, I think we will let it override the file for the first release.
2. I will make it to do it, but will leave a warning log that the specified path did not exist and will be created.
3. I believe part of the value of this feature is when we add more usage data options to existing resources, so let's say the users sees that by upgrading the infracsot some new fields were added to its existing resources usages.

FYI, I solved the embed problem by using a more appropriate workaround (in my opinion =)) )

There's also another problem:
4. We have a useful feature that will show a cost component is dependent on usage data (when the usage data is nill), by adding this feature cause all usage keys will have a 0 value we can't distinguish between real 0s (specified by the user) and the ones we prefilled. It's one of the common pitfalls of Golanf defaults and I can't think of any quick workarounds.

sinabakh:
The functionality is done, it would be great if you could test it.

alikhajeh1:
Re #4, I think that's ok for now as the we're guiding new users to use the usage-file if they want to see costs. They can also run `infracost breakdown/diff` without `--usage-file` too. Remember this is just one iteration of many... next we might start to fetch usage data from other sources

aliscott:
@sinabakh thanks have been playing around with this and it works really well! Here's some more thoughts: 

5. When `--sync-usage-file` is specified but `--usage-file` isn't we should show a warning.

6. There's at least one usage value that has a string value (`aws_instance > operating_system`). At the moment this sets the value to "0" and overrides it on each run. @alikhajeh1, @sinabakh  any thoughts what we should do about this? We could either remove this functionality, have some way of defaulting this to "linux" or an empty string.

7. At the moment when you remove a resource then `--sync-usage-file` deletes the information for that resource from the usage file. I think this is okay, but wondering if we should add a prompt to confirm? What do you think @alikhajeh1 @sinabakh. I guess this is similar to point 1 above, but we should be clear if this flag is going to be interactive or not going forward as that makes a difference with how people integrate it.

8. `--sync-usage-file` should be marked as `(experimental)` in the help text for just now.

sinabakh:
Thanks @aliscott 

Regarding 7, I would prefer not to remove the deleted resources by default and add another flag for it like `sync-deleted-resources` or sth like this.

Here's my idea of changes for this to be initially released:

A) When --sync-usage-file is specified but --usage-file isn't we should show a warning
B) Fix string default values (I will think about how to handle it)
C) The result of discussion on 7
D) Mark this flag experimental

Any additional tasks?

Btw, there's a minor problem: The spinner removes all messages produced in this stage. I don't know whether it's a valid issue or sth is wrong with my terminal env.

sinabakh:
@aliscott Regarding 6, I changed the code a bit to handle string values. Since empty strings make no sense I used the value in the example file as the default value so we might want to change them to the most common possible value.

alikhajeh1:
Re #6, defaulting to the value in the usage-file is good, they're already set good defaults.
Re #7, `sync-usage-file` deleting resources that have been deleted from the TF state is fine as otherwise the deleted resources need to be removed manually to cleanup. Those resources won't make a different to the output, nor fail command runs, and given this is experimental, we can change the behavior so go with whatever is fastest.

cmd/infracost/run.go
```diff
@@ -27,6 +27,8 @@ func addRunFlags(cmd *cobra.Command) {
 	cmd.Flags().String("terraform-workspace", "", "Terraform workspace to use. Applicable when path is a Terraform directory")
 
 	cmd.Flags().Bool("show-skipped", false, "Show unsupported resources, some of which might be free")
+
+	cmd.Flags().Bool("sync-usage-file", false, "Experimental: Autofill the missing usage data and update the usage file")
```
alikhajeh1, line 31:
```suggestion
	cmd.Flags().Bool("sync-usage-file", false, "Autofill usage-file with missing resources (experimental)")
```

ccom-dev-platform:
Is this functionality supported by the CI/CD integrations yet? Didn't see any docs regarding it!

dat-intellis:
I stumbled upon this, and it doesn't seem to support any usage-files within configuration files. Specifically, static configuration files: https://www.infracost.io/docs/features/config_file/#static-config-file-project-parameters 

(I happen to think the usage-file keyword should be project related, but alas : ) )

PR #550, opened 2021-04-04
chore(deps): install unzip instead of zip package in Dockerfile

PeterDaveHello:
`zip` is actually not needed in the Dockerfile, but install `zip` by default will also install the suggested `unzip` package, so that the later commands can use `unzip` without a problem, but `zip` really doesn't need to be installed here 😄 

This will help to speed up the builder stage a little bit 😀


alikhajeh1:
Thanks @PeterDaveHello - well spotted!

PR #551, opened 2021-04-04
chore: reorder Dockerfile instruction to minimize image size

PeterDaveHello:
Change the file permision in the builder stage, save an image layer and
the duplicated file caused by permission change.

Result as below, saved the size of `infracost` the binary itself(about 16MB):

```
REPOSITORY      TAG        IMAGE ID         CREATED              SIZE
infracost       after      0a932e916231     3 minutes ago        638MB
infracost       before     37edb87395bb     8 minutes ago        654MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,7 @@ WORKDIR /app
 COPY . .
 RUN make deps
 RUN NO_DIRTY=true make build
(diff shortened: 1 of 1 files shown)


alikhajeh1:
Every little counts 📉

PR #556, opened 2021-04-06
fix: rename costcomponets

Pauluson:


Pauluson:
@alikhajeh1 this is ready to review

Issue #528, opened 2021-03-19
Support resource arrays in usage file


aliscott:
Terraform resources can be created using `for_each` or `count`. This creates an array of resources. We could support setting usage values for all elements of the array by using a wildcard like this:

```yaml
resource_usage:
  aws_cloudwatch_log_group.loggroup[*]:
    ...
```

I think expected functionality would be that the usage value should apply individually to each element of the array. But an alternative is that the usage value is the aggregated using across all elements of the array.

Relevant code: https://github.com/infracost/infracost/blob/master/internal/providers/terraform/parser.go#L101-L105

Unit tests can be added to cover the new functionality.
Also add a new integration test to cover this. This can be added in a file called `parser_test.go` (separate from parser_internal_test.go) within a package called `parser_test`.

ekifel:
@alikhajeh1 hey, I want to work on it

alikhajeh1:
Released today in v0.8.4 - thanks @ekifel!

PR #562, opened 2021-04-08
feat(usage): add support resource arrays in usage-file

ekifel:
Issue #528 

Example: 
Usage-file: 
```
 aws_cloudwatch_log_group.my_log_group[*]:
    storage_gb: 1000               
    monthly_data_ingested_gb: 1000 
    monthly_data_scanned_gb: 200  
``` 

Terraform config: 
```
resource "aws_cloudwatch_log_group" "my_log_group" {
  count = 2
  name              = "log-group${count.index}"
}
```

Output: 
```
  NAME                                          MONTHLY QTY  UNIT  PRICE   HOURLY COST  MONTHLY COST  

  aws_cloudwatch_log_group.my_log_group[0]                                                            
  ├─ Data ingested                                    1,000  GB    0.5000       0.6849      500.0000  
  ├─ Archival Storage                                 1,000  GB    0.0300       0.0411       30.0000  
  └─ Insights queries data scanned                      200  GB    0.0050       0.0014        1.0000  
  Total                                                                         0.7274      531.0000  
                                                                                                      
  aws_cloudwatch_log_group.my_log_group[1]                                                            
  ├─ Data ingested                                    1,000  GB    0.5000       0.6849      500.0000  
  ├─ Archival Storage                                 1,000  GB    0.0300       0.0411       30.0000  
  └─ Insights queries data scanned                      200  GB    0.0050       0.0014        1.0000  
  Total                                                                         0.7274      531.0000
```


internal/providers/terraform/parser.go
```diff
@@ -99,7 +99,12 @@ func (p *Parser) parseJSONResources(parsePrior bool, baseResources []*schema.Res
 	p.stripDataResources(resData)
 
 	for _, d := range resData {
-		if r := p.createResource(d, usage[d.Address]); r != nil {
+		usageAddress := usage[d.Address]
+
+		if arrayUsageAddress := usage[fmt.Sprintf("%s[*]", strings.Split(d.Address, "[")[0])]; arrayUsageAddress != nil {
```
aliscott, line 104: Would this work with counts in modules? If the module itself uses `count` or `for_each` and the resource has the same then I think the address comes through as `module.my_module[0].aws_instance.my_instance[2]`

aliscott (commented):
Awesome work! The tests look great.

infracost-usage-example.yml
```diff
@@ -14,6 +14,16 @@ resource_usage:
   # module.lambda_function.aws_lambda_function.this[0]:
   #   monthly_requests: 20000
   #   request_duration_ms: 600
+  #
+  # You can use [*] for array of resources to apply usage-values individually to each element of the array.
+  # Specifying an array element have priority.
```
alikhajeh1, line 19: If this what we mean here?
```
If usage for an array element (e.g. this[0]) and [*] exist, only the array element's usage will be applied to that resource.
```

ekifel: Yes it is. Should I fix it to your describe?

alikhajeh1: Thanks, that would be good.

aliscott (approved):

internal/providers/terraform/parser_test.go
```diff
@@ -0,0 +1,300 @@
+package terraform
```
aliscott, line 1: We already have `parser_internal_test.go`, I think this is left from when we had integration tests for the parser. Do you think we should move all the tests from there into this `parser_test.go` file as well?

internal/providers/terraform/parser.go
```diff
@@ -100,7 +100,16 @@ func (p *Parser) parseJSONResources(parsePrior bool, baseResources []*schema.Res
 	p.stripDataResources(resData)
 
 	for _, d := range resData {
-		if r := p.createResource(d, usage[d.Address]); r != nil {
+		var usageData *schema.UsageData
+
+		if ud := usage[d.Address]; ud != nil {
+			usageData = ud
+		} else if strings.HasSuffix(d.Address, "]") {
+			if arrayUsageData := usage[fmt.Sprintf("%s*]", d.Address[:len(d.Address)-2])]; arrayUsageData != nil {
```
aliscott, line 108: Does this work when count >= 10?

aliscott (commented):
@ekifel, looking good. Just a couple more questions on this.

internal/providers/terraform/parser.go
ekifel: Do you mean 'count' as lenght of d.Address? If yes, sure. I checked it in situations with arrays of modules, it was over ~20-30 symbols and it works fine.

internal/providers/terraform/parser_test.go
ekifel: I think we should, I don't see any reasons to split that tests to two files. Should I move them?

aliscott: Yes that would be great, thanks.

internal/providers/terraform/parser.go
aliscott: I mean where `count = 11` in the resource. So the resource would have `aws_instance.my_instance[10]` as the address.

ekifel: ohh, missunderstanding, forgot about that. I will add checking an array lenght before it.

ekifel: Fixed! Ready to review

alikhajeh1:
@ekifel I tested this, love it!

I found one array case where it's not working though, can you please take a look?
Given this terraform resource:
```
resource "google_container_cluster" "with_node_pools_zonal" {
  name     = "with-node-pools"
  location = "us-central1-a"

  node_pool {
    node_config {
      machine_type = "n1-standard-16"
    }
  }

  node_pool {
    node_config {
      machine_type  = "n1-standard-16"
    }
  }
}
```

And this usage-file:
```
version: 0.1
resource_usage:
  google_container_cluster.with_node_pools_zonal:
    node_pool[*]:
      nodes: 8
```

If I do `node_pool[0]` it works as expected. It doesn't apply the usage-file wildcard.

PR #564, opened 2021-04-09
fix: sanitize path in error

aliscott:


cmd/infracost/run.go
```diff
@@ -46,14 +47,14 @@ func runMain(cmd *cobra.Command, cfg *config.Config) error {
...
-			return errors.New(m)
+			return events.NewError(errors.New(m), "Could not detect path type")
 		}
 
 		if cmd.Name() == "diff" && provider.Type() == "terraform_state_json" {
 			m := "Cannot use Terraform state JSON with the infracost diff command.\n\n"
 			m += fmt.Sprintf("Use the %s flag to specify the path to one of the following:\n", ui.PrimaryString("--path"))
 			m += " - Terraform plan JSON file\n - Terraform directory\n - Terraform plan file"
-			return errors.New(m)
+			return events.NewError(errors.New(m), "Cannot use Terraform state JSON with the infracost diff command")
```
alikhajeh1, line 57: It might help to offer an alternative? e.g. use the plan JSON instead?

alikhajeh1 (approved):

cmd/infracost/run.go
aliscott: We do above. This is just the shortened error for reporting.

Issue #535, opened 2021-03-26
Differentiate path-not-found vs undetectable path type


alikhajeh1:
The following two runs produce the same error, which confuses CI/CD users as they think the `path` param file/dir is present and something is wrong with the CLI. Usually in CI/CD, the issue is that the file cannot be found, because the path is wrong or, in the case of CircleCI, a previous task produced the plan.json file but this hasn't been attached to the infracost orb.

```
$ infracost breakdown --path nonexistentfile
Error: Could not detect path type for nonexistentfile
...

$ echo abc > plan.json
$ infracost breakdown --path plan.json
Error: Could not detect path type for plan.json
```

It would be clearer if the first error was "Error: No such file or directory plan.json" (similar to what `ls nonexistentdir` outputs).

Pauluson:
i want to do it

alikhajeh1:
Released in v0.8.4

PR #565, opened 2021-04-09
feat: add error message to breakdown/cli

Pauluson:
#535 
`<pre>make -s run ARGS=&quot;breakdown --path plan&quot;
<font color="#EF2929">Error:</font> No such file or directory plan

Use the <font color="#34E2E2">--path</font> flag to specify the path to one of the following:
 - Terraform plan JSON file
 - Terraform directory
 - Terraform plan file
 - Terraform state JSON file
</pre>`
`<pre> make -s run ARGS=&quot;breakdown --path plan.json&quot;
<font color="#EF2929">Error:</font> Could not detect path type for plan.json

Use the <font color="#34E2E2">--path</font> flag to specify the path to one of the following:
 - Terraform plan JSON file
 - Terraform directory
 - Terraform plan file
 - Terraform state JSON file
</pre>`

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -38,7 +38,7 @@ func runMain(cmd *cobra.Command, cfg *config.Config) error {
 	for _, projectCfg := range cfg.Projects {
 		provider, err := providers.Detect(cfg, projectCfg)
 
-		if provider == nil {
+		if err != nil {
 			m := fmt.Sprintf("%s\n\n", err)
 			m += fmt.Sprintf("Use the %s flag to specify the path to one of the following:\n", ui.PrimaryString("--path"))
 			m += " - Terraform plan JSON file\n - Terraform directory\n - Terraform plan file"
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -13,6 +13,7 @@ import (
 	"github.com/infracost/infracost/internal/ui"
 	"github.com/infracost/infracost/internal/usage"
 	"github.com/pkg/errors"
+
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


cmd/infracost/run.go
```diff
@@ -38,6 +38,18 @@ func runMain(cmd *cobra.Command, cfg *config.Config) error {
...
 		if provider == nil {
+			m := fmt.Sprintf("No such file or directory %s\n\n", ui.DisplayPath(projectCfg.Path))
+			m += fmt.Sprintf("Use the %s flag to specify the path to one of the following:\n", ui.PrimaryString("--path"))
+			m += " - Terraform plan JSON file\n - Terraform directory\n - Terraform plan file"
+
+			if cmd.Name() != "diff" {
+				m += "\n - Terraform state JSON file"
+			}
+
+			return errors.New(m)
```
aliscott, line 49:
```suggestion
			return events.NewError(errors.New(m), "No such file or directory ")
```
This sanitizes the path out of any error reported upstream, while still keeping the original error message to show the user.

aliscott (commented):
I think we may need to update the `providers.Detect` function to return an error as well. I don't think this distinguishes between the file not existing and the path type not being detected.

Currently if `provider == nil` it can be either. So maybe we need the `providers.Detect` method to return different errors for `Could not detect path type` and `No such file or directory`.

cmd/infracost/run.go
Pauluson: ah, okey, i understand

cmd/infracost/run.go
```diff
@@ -35,10 +36,10 @@ func runMain(cmd *cobra.Command, cfg *config.Config) error {
 	projects := make([]*schema.Project, 0)
 
 	for _, projectCfg := range cfg.Projects {
-		provider := providers.Detect(cfg, projectCfg)
+		provider, err := providers.Detect(cfg, projectCfg)
 
 		if provider == nil {
```
aliscott, line 41:
```suggestion
		if err != nil {
```

aliscott (approved):
LGTM!

PR #585, opened 2021-04-15
Update dockerfile

alikhajeh1:


Dockerfile
```diff
@@ -2,15 +2,15 @@ FROM golang:1.16 as builder
...
-ARG TERRAGRUNT_VERSION=0.28.8
+ARG TERRAGRUNT_VERSION=0.28.22
 
 # Set Environment Variables
 SHELL ["/bin/bash", "-c"]
 ENV HOME /app
 ENV CGO_ENABLED 0
 
 # Install Packages
-RUN apt-get update -q && apt-get -y install unzip jq
```
alikhajeh1, line 13: didn't look like we use `jq` in the builder part

Dockerfile
```diff
@@ -45,7 +45,17 @@ RUN chmod +x /app/build/infracost
 # Application
 FROM alpine:3.13 as app
 # Tools needed for running diffs in CI integrations
-RUN apk --no-cache add ca-certificates openssl openssh-client curl git jq
```
alikhajeh1, line 48: the old diff.sh script used git but not anymore, should we keep git for `scripts/get_version.sh` though?

aliscott: `scripts/get_version.sh` is only used by make which we don't include here anyway and don't think we'll need.

aliscott (approved):

aliscott:
LGTM

PR #594, opened 2021-04-19
feat(azure): Add support for Azure MariaDB database

ekifel:
Issue #573 

Output example: 

```
 Name                                               Quantity  Unit            Monthly Cost
 azurerm_mariadb_server.my_server                                                          
 ├─ Compute (MO_Gen5_16)                                 730  hours              $1,379.41 
 ├─ Storage                                                5  GB-months              $0.57 
 └─ Additional backup storage                          2,000  GB-months            $200.00 
                                                                                           
 azurerm_mariadb_server.without_usage                                                      
 ├─ Compute (GP_Gen5_4)                                  730  hours                $255.79 
 ├─ Storage                                               40  GB-months              $4.60 
 └─ Additional backup storage                  Cost depends on usage: $0.10 per GB-months 
```

### DIFF
--- a/internal/providers/terraform/azure/mariadb_server.go
+++ b/internal/providers/terraform/azure/mariadb_server.go
@@ -2,6 +2,7 @@ package azure
 
 import (
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/infracost/infracost/internal/schema"
@@ -21,15 +22,26 @@ func NewAzureMariaDBServer(d *schema.ResourceData, u *schema.UsageData) *schema.
 
 	region := d.Get("location").String()
 	sku := d.Get("sku_name").String()
-	tier := strings.Split(sku, "_")[0]
-	family := strings.Split(sku, "_")[1]
-	cores := strings.Split(sku, "_")[2]
+	var tier, family, cores string
+	if s := strings.Split(sku, "_"); len(s) == 3 {
+		tier = strings.Split(sku, "_")[0]
+		family = strings.Split(sku, "_")[1]
+		cores = strings.Split(sku, "_")[2]
+	} else {
+		log.Println("MariaDB sku format is not recognised. Please follows the tier_family_cores pattern (for example: GP_Gen5_4)")
+		return nil
+	}
 
 	tierName := map[string]string{
 		"B":  "Basic",
 		"GP": "General Purpose",
 		"MO": "Memory Optimized",
 	}[tier]
+
+	if tierName == "" {
+		log.Println("MariaDB tier is not recognised. Please use B_ for Basic, GP_ for General Purpose, MO_ for Memory Optimized")
+		return nil
+	}
 
 	productNameRegex := fmt.Sprintf("/%s - Compute %s/", tierName, family)
 	skuName := fmt.Sprintf("%s vCore", cores)
@@ -82,10 +94,8 @@ func NewAzureMariaDBServer(d *schema.ResourceData, u *schema.UsageData) *schema.
 	}
 
 	skuName = "Backup LRS"
-	if d.Get("geo_redundant_backup_enabled").Exists() {
-		if d.Get("geo_redundant_backup_enabled").String() == "true" {
-			skuName = "Backup GRS"
-		}
+	if d.Get("geo_redundant_backup_enabled").Exists() && d.Get("geo_redundant_backup_enabled").Bool() {
+		skuName = "Backup GRS"
 	}
 
 	costComponents = append(costComponents, &schema.CostComponent{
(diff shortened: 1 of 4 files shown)


internal/providers/terraform/azure/mariadb_server.go
```diff
@@ -0,0 +1,112 @@
...
+
+	var backupStorageGB *decimal.Decimal
+
+	if u != nil && u.Get("additional_backup_storage_gb").Exists() {
+		backupStorageGB = decimalPtr(decimal.NewFromInt(u.Get("additional_backup_storage_gb").Int()))
+	}
+
+	skuName = "Backup LRS"
+	if d.Get("geo_redundant_backup_enabled").Exists() {
+		if d.Get("geo_redundant_backup_enabled").String() == "true" {
```
aliscott, line 86:
```suggestion
	if d.Get("geo_redundant_backup_enabled").Exists() && d.Get("geo_redundant_backup_enabled").Bool() {
```

internal/providers/terraform/azure/mariadb_server.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func NewAzureMariaDBServer(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	var costComponents []*schema.CostComponent
+
+	region := d.Get("location").String()
+	sku := d.Get("sku_name").String()
+	tier := strings.Split(sku, "_")[0]
+	family := strings.Split(sku, "_")[1]
+	cores := strings.Split(sku, "_")[2]
```
aliscott, line 26: I think we should check the length of the split here so we don't have the possibility of an Index Out of Range error. Maybe we should log if the sku format is not recognised.

internal/providers/terraform/azure/mariadb_server.go
```diff
@@ -0,0 +1,112 @@
...
+
+	tierName := map[string]string{
+		"B":  "Basic",
+		"GP": "General Purpose",
+		"MO": "Memory Optimized",
+	}[tier]
```
aliscott, line 32: Same here, I think we should log if the tier is not recognised

aliscott (approved):
LGTM

internal/providers/terraform/azure/mariadb_server.go
ekifel: Are u sure that we need to do this? As Terraform don't give user an opportunity to set wrong sku. 
Example:

```
Running terraform plan

Error: expected sku_name to be one of [B_Gen5_1 B_Gen5_2 GP_Gen5_2 GP_Gen5_4 GP_Gen5_8 GP_Gen5_16 GP_Gen5_32 MO_Gen5_2 MO_Gen5_4 MO_Gen5_8 MO_Gen5_16], got MO_Gen5
```

aliscott: Nice! It seems they do a lot more validation at the plan stage with Azure. I think it's still best to add this sort of validation on our end. I guess it's pulling those SKUs from the Azure API and we don't know if and when those will change.

internal/providers/terraform/azure/mariadb_server.go
```diff
@@ -0,0 +1,122 @@
...
+
+	region := d.Get("location").String()
+	sku := d.Get("sku_name").String()
+	var tier, family, cores string
+	if s := strings.Split(sku, "_"); len(s) == 3 {
+		tier = strings.Split(sku, "_")[0]
+		family = strings.Split(sku, "_")[1]
+		cores = strings.Split(sku, "_")[2]
+	} else {
+		log.Println("MariaDB sku format is not recognised. Please follows the tier_family_cores pattern (for example: GP_Gen5_4)")
```
aliscott, line 31: log.Warnf("
```suggestion
		log.Warnf("Unrecognised MariaDB SKU format for resource %s: %s", d.Address, sku)
```

I think we should output the address and sku.

internal/providers/terraform/azure/mariadb_server.go
```diff
@@ -0,0 +1,122 @@
+package azure
+
+import (
+	"fmt"
+	"log"
```
aliscott, line 5:
```suggestion
	log "github.com/sirupsen/logrus"
```

internal/providers/terraform/azure/mariadb_server.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	tierName := map[string]string{
+		"B":  "Basic",
+		"GP": "General Purpose",
+		"MO": "Memory Optimized",
+	}[tier]
+
+	if tierName == "" {
+		log.Println("MariaDB tier is not recognised. Please use B_ for Basic, GP_ for General Purpose, MO_ for Memory Optimized")
```
aliscott, line 42:
```suggestion
		log.Warnf("Unrecognised MariaDB tier prefix for resource %s: %s", d.Address, tierName)
```

aliscott (approved):

aliscott (approved):

Issue #529, opened 2021-03-19
Support AWS EC2 RI prices


alikhajeh1:
Currently Infracost supports on-demand prices. We can add the following options to [aws_instance](https://github.com/infracost/infracost/blob/master/infracost-usage-example.yml#L118-L119) in the infracost usage-file to enable users to switch to see RI prices:
- reserved_instance_type: standard or convertible
- reserved_instance_term: 1_year or 3_year
- reserved_instance_payment_option: no_upfront, partial_upfront, all_upfront

The ability to define wildcards in the usage-file (e.g. `aws_instance.*` or `aws_instance.app1*`) would save users from having to define that for all instances. A similar thing could be applied to Azure instances that also have [RI prices](https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/).

See [this reddit thread](https://www.reddit.com/r/Terraform/comments/m81h38/infracost_diff_git_diff_but_for_cloud_costs/grhzcpb/?utm_source=reddit&utm_medium=web2x&context=3) for background - thanks to u/sumthingcool for the discussion.

alikhajeh1:
I think we can add the above usage file parameters to the following supported resources:
- aws_instance
- aws_eks_node_group
- aws_autoscaling_group (which has aws_launch_configuration as a sub-resource)

That would mean we support EC2 RI prices anywhere that uses them.

ekifel:
@alikhajeh1 Good. Can I work on it?

alikhajeh1:
This was released in v0.8.5

PR #602, opened 2021-04-21
feat(aws): add support for AWS EC2 RI prices

ekifel:
Issue #529 

Output examples:

```
 Name                                                         Quantity  Unit            Monthly Cost 
                                                                                                     
 aws_autoscaling_group.my_asg                                                                        
 └─ aws_launch_configuration.lc1                                                                     
    ├─ Instance usage (Linux/UNIX, reserved, t3.medium)         10,950  hours                $285.80 
    ├─ EC2 detailed monitoring                                     105  metrics               $31.50 
    ├─ CPU credits                                       Cost depends on usage: $0.05 per vCPU-hours 
    ├─ root_block_device                                                                             
    │  └─ Storage (general purpose SSD, gp2)                       150  GB-months             $15.00 
    ├─ ebs_block_device[0]                                                                           
    │  └─ Storage (general purpose SSD, gp2)                       150  GB-months             $15.00 
    └─ ebs_block_device[1]                                                                           
       └─ Storage (general purpose SSD, gp3)                       150  GB-months             $12.00 
                                                                                                                                                                                        
 aws_eks_node_group.my_instance                                                                      
 ├─ Instance usage (Linux/UNIX, reserved, t3.medium)               730  hours                  $9.05 
 ├─ CPU credits                                                    730  vCPU-hours            $36.50 
 └─ Storage (general purpose SSD, gp2)                              20  GB-months              $2.00 
                                                                                                     
 aws_instance.my_instance                                                                            
 ├─ Instance usage (Linux/UNIX, reserved, t3.medium)               730  hours                  $0.00 
 ├─ CPU credits                                          Cost depends on usage: $0.05 per vCPU-hours 
 └─ root_block_device                                                                                
    └─ Storage (general purpose SSD, gp2)                            8  GB-months              $0.80 
                                                                                                     
```

aws_instance has `reserved_instance_payment_option: all_upfront` usage-option, so it has $0.00 Monthly Cost

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -46,6 +46,7 @@ resource_usage:
 
   aws_autoscaling_group.my_asg:
     instances: 15 # Number of instances in the autoscaling group.
+    operating_system: linux # Override the operating system of the instance, can be: linux, windows, suse, rhel.
     reserved_instance_type: standard # Offering class for Reserved Instances. Can be: convertible, standard.
     reserved_instance_term: 1_year # Term for Reserved Instances. Can be: 1_year, 3_year.
     reserved_instance_payment_option: no_upfront # Payment option for Reserved Instances. Can be: no_upfront, partial_upfront, all_upfront.
@@ -137,6 +138,7 @@ resource_usage:
     monthly_infrequent_access_write_gb: 100 # Monthly infrequent access write requests in GB.
 
   aws_eks_node_group.my_instance:
+    operating_system: linux # Override the operating system of the instance, can be: linux, windows, suse, rhel.
     reserved_instance_type: standard # Offering class for Reserved Instances. Can be: convertible, standard.
     reserved_instance_term: 1_year # Term for Reserved Instances. Can be: 1_year, 3_year.
     reserved_instance_payment_option: partial_upfront # Payment option for Reserved Instances. Can be: no_upfront, partial_upfront, all_upfront.
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -44,6 +44,12 @@ resource_usage:
     message_size_kb: 32               # Average size of the messages sent to the Websocket API Gateway in KB. Messages are metered in 32 KB increments, maximum size is 128KB.
     monthly_connection_mins: 10000000 # Monthly total connection minutes to Websockets.
 
+  aws_autoscaling_group.my_asg:
+    instances: 15 # Number of instances in the autoscaling group.
+    reserved_instance_type: standard # Offering class for Reserved Instances. Can be: convertible, standard.
+    reserved_instance_term: 1_year # Term for Reserved Instances. Can be: 1_year, 3_year.
+    reserved_instance_payment_option: no_upfront # Payment option for Reserved Instances. Can be: no_upfront, partial_upfront, all_upfront.
+
   aws_cloudwatch_event_bus.my_events:
     monthly_custom_events: 1000000            # Monthly custom events published. Each 64 KB chunk of payload is billed as 1 event.
     monthly_third_party_events: 2000000       # Monthly third-party and cross-account events published. Each 64 KB chunk of payload is billed as 1 event.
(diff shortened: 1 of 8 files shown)


internal/providers/terraform/aws/eks_node_group.go
```diff
@@ -70,14 +70,28 @@ func NewEKSNodeGroup(d *schema.ResourceData, u *schema.UsageData) *schema.Resour
...
+		RIType = u.Get("reserved_instance_type").String()
+		if u.Get("reserved_instance_term").Exists() {
+			RITerm = u.Get("reserved_instance_term").String()
+		}
+		if u.Get("reserved_instance_payment_option").Exists() {
+			RIPaymentOption = u.Get("reserved_instance_payment_option").String()
+		}
+	}
+
+	if RIType != "" {
```
alikhajeh1, line 91: Should this also be checking that RITerm and RIPaymentOption are set? otherwise the price lookup might return multiple prices if those are not set?

internal/providers/terraform/aws/instance.go
```diff
@@ -91,6 +91,22 @@ func computeCostComponent(d *schema.ResourceData, u *schema.UsageData, purchaseO
...
+		reservedIType = u.Get("reserved_instance_type").String()
+		if u.Get("reserved_instance_term").Exists() {
+			reservedTerm = u.Get("reserved_instance_term").String()
+		}
+		if u.Get("reserved_instance_payment_option").Exists() {
+			reservedPaymentOption = u.Get("reserved_instance_payment_option").String()
+		}
+	}
+
+	if reservedIType != "" {
```
alikhajeh1, line 106: same comment as above

alikhajeh1 (commented):
Thanks! I added a few suggestions

infracost-usage-example.yml
```diff
@@ -130,14 +136,22 @@ resource_usage:
...
+
   aws_elasticache_cluster.my_redis_snapshot:
     snapshot_storage_size_gb: 10000 # Size of Redis snapshots in GB. 
 
   aws_elb.my_elb:
     monthly_data_processed_gb: 10000 # Monthly data processed by a Classic Load Balancer in GB.
 
-  aws_instance.windows_instance:
+  aws_instance.my_instance:
     operating_system: linux # Override the operating system of the instance, can be: linux, windows, suse, rhel.
```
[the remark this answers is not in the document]
ekifel: I investigated AWS documentation, these resources supports EC2 instances with different OS. So, I think yes, we should add operation_system to them too and set Linux as default OS. Do you agree?

internal/providers/terraform/aws/eks_node_group.go
ekifel: Yes, sure. Will add checks for them too

infracost-usage-example.yml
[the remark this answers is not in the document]
alikhajeh1: Thanks! That makes sense

ekifel:
@alikhajeh1, this is ready to review!

alikhajeh1:
Thanks @ekifel! This is almost done, can you please check my latest commit? I refactored it so it follows this bit of the contributing guide:
- to conditionally set values based on the Terraform resource values, first check `d.Get("value_name").Type != gjson.Null` like in the [google_container_cluster](https://github.com/infracost/infracost/blob/f7b0594c1ee7d13c0c37acb8edfb36dde223471a/internal/providers/terraform/google/container_cluster.go#L38-L42) resource. In the past we used `.Exists()` but this only checks that the key does not exist in the Terraform JSON, not that the key exists and is set to null.

Can you please take a look at the TODO comment I added in the code? Currently if I run this with the following usage-file, I see `Instance usage (..., reserved, ...)` but the on-demand price is used. Once this TODO is done, we should see a warning when running with `--log-level warn` and it'll show (on-demand) in the output. This is similar to what is done for the `operating_system` property too.
```
resource_usage:
  aws_instance.web_app:
    operating_system: linux
    reserved_instance_payment_option: partial_upfront
    reserved_instance_term: a
    reserved_instance_type: standard
```

ekifel:
@alikhajeh1, this is ready to review!

PR #608, opened 2021-04-23
fix(docker): add git to docker image

### LINKED ISSUE #607: Git removed from Docker image which breaks remote modules
In the PR #585 Git was removed from the list of installed packages but it is required by Terraform to pull remote modules

```
    Error: Failed to download module
    Could not download module "private" (main.tf:1) source code from
    "git@github.com:private/repo": error
    downloading
    'ssh://git@github.com/private/repo': git
    must be available and on the PATH
```
I think maybe removing git was unintentional without realizing Terraform depended on it

keslerm:
This fixes #607 to allow terraform to pull remote git modules

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,7 +45,7 @@ RUN chmod +x /app/build/infracost
 # Application
 FROM alpine:3.13 as app
 # Tools needed for running diffs in CI integrations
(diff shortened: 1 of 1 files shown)


aliscott:
Thanks @keslerm!

Issue #509, opened 2021-03-10
infracost diff html format


sushant-pradhan:
Hello team, I was wondering if we can enable html format for diff command. The console output looks good but if I have to send the diff output over email, the console output doesn't really go well. It will be better if we have a html equivalent like we have for breakdown command

infracost v0.8.1

alikhajeh1:
@sushant-pradhan hah, funny you mention that as I posted [a comment](https://github.com/infracost/infracost/issues/394#issuecomment-795914531) a few hours ago asking for exactly this sort of feedback.

alikhajeh1:
@sushant-pradhan who would you email the HTML reports to? fellow team mates? other teams? managers? clients? This context helps us as we think about how the output of Infracost can be shared with others

sushant-pradhan:
@alikhajeh1 we plan to send to the team mates who can review the delta and approve the plan. It something similar to the PR comments we add from GH actions.

alikhajeh1:
@sushant-pradhan we're working on a dashboard so team members can view run details, plz email me at ali.hosseini@infracost.io if this is something that might interest your team!

alikhajeh1:
@sushant-pradhan as part of our [Terraform Cloud integration](https://www.infracost.io/docs/iac_tools/terraform_cloud_enterprise#terraform-cloud-run-tasks), we've developed a way to generate sharable, pre-authenticated, HTML reports (with both the breakdown and the diff format) that could be passed around teams. I've emailed you an example as we're thinking of developing that further so going to close this issue for now.

PR #611, opened 2021-04-23
feat: add jenkins support

sinabakh:
The `jenkins_diff.sh` does relatively the same task as the other diff scripts, here is a list of changes:
* `path` variable is changed to `iac_path`
* Remove support for deprecated flags
* Remove `post_condition`
* Add `fail_condition` with the `percentage_threshold` field to fail the pipeline if the change is over the threshold
* Add a function to wrap the diff output in a simple HTML page
* Add few utility functions to make the script work with Jenkins
* Diff-output is always printed in the output
* Save the diff output HTML in a file for further use

Here's a sample `Jenkinsfile` config:
```Jenkinsfile
pipeline {
    agent any
    stages {
        stage('infracost-breakdown') {
            agent {
                docker {
                  image 'infracost-local'
                  args '--user=root --entrypoint='
                }
            }
            environment {
              INFRACOST_API_KEY = credentials('jenkins-infracost-api-key')
              IAC_PATH = 'terraform'
              FAIL_CONDITION = '{"percentage_threshold": 10}'
            }

            steps {
                sh '/scripts/ci/jenkins_diff.sh'

                publishHTML (target: [
                    	allowMissing: false,
      		    	alwaysLinkToLastBuild: false,
      			keepAll: true,
      			reportDir: './',
      			reportFiles: 'infracost_diff_output.html',
      			reportName: "Infracost Diff Output"
    		])

            }
        }
    }
}
```

Here's the pipeline's "Output consone":
Screen Shot 2021-04-23 at 9 46 53 PM

And the output get's its own sub-menu:
Screen Shot 2021-04-22 at 1 19 09 AM

Screen Shot 2021-04-23 at 9 49 46 PM

### DIFF
--- a/scripts/ci/jenkins_diff.sh
+++ b/scripts/ci/jenkins_diff.sh
@@ -1,0 +1,184 @@
+#!/bin/sh -le
+
+# This is an Jenkins-specific script that runs infracost on the current branch.
+# Usage docs: https://www.infracost.io/docs/integrations/
+
+fix_env_vars () {
+    # Jenkins has problems with envs case sensivity
+    iac_path=${iac_path:-$IAC_PATH}
+    terraform_plan_flags=${terraform_plan_flags:-$TERRAFORM_PLAN_FLAGS}
+    terraform_workspace=${terraform_workspace:-$TERRAFORM_WORKSPACE}
+    usage_file=${usage_file:-$USAGE_FILE}
+    config_file=${config_file:-$CONFIG_FILE}
+    fail_condition=${fail_condition:-$FAIL_CONDITION}
+}
+
+process_args () {
+  # Set variables based on the order for GitHub Actions, or the env value for other CIs
+  iac_path=${1:-$iac_path}
+  terraform_plan_flags=${2:-$terraform_plan_flags}
+  terraform_workspace=${3:-$terraform_workspace}
+  usage_file=${4:-$usage_file}
+  config_file=${5:-$config_file}
+  fail_condition=${7:-$fail_condition}
+
+  # Validate fail_condition
+  if ! echo "$fail_condition" | jq empty; then
+    echo "Error: fail_condition contains invalid JSON"
+  fi
+
+  # Set defaults
+  if [ ! -z "$fail_condition" ] && [ "$(echo "$fail_condition" | jq '.percentage_threshold')" != "null" ]; then
+    fail_percentage_threshold=$(echo "$fail_condition" | jq -r '.percentage_threshold')
+  fi
+  fail_percentage_threshold=${fail_percentage_threshold:--1}
+  INFRACOST_BINARY=${INFRACOST_BINARY:-infracost}
+
+  # Export as it's used by infracost, not this script
+  export INFRACOST_LOG_LEVEL=${INFRACOST_LOG_LEVEL:-info}
+  export INFRACOST_CI_DIFF=true
+
+  if [ ! -z "$GIT_SSH_KEY" ]; then
+    echo "Setting up private Git SSH key so terraform can access your private modules."
+    mkdir -p .ssh
+    echo "${GIT_SSH_KEY}" > .ssh/git_ssh_key
+    chmod 600 .ssh/git_ssh_key
+    export GIT_SSH_COMMAND="ssh -i $(pwd)/.ssh/git_ssh_key -o 'StrictHostKeyChecking=no'"
+  fi
+
+}
+
+build_breakdown_cmd () {
+  breakdown_cmd="${INFRACOST_BINARY} breakdown --no-color --format json"
+
+  if [ ! -z "$iac_path" ]; then
+    breakdown_cmd="$breakdown_cmd --path $iac_path"
+  fi
+  if [ ! -z "$terraform_plan_flags" ]; then
+    breakdown_cmd="$breakdown_cmd --terraform-plan-flags \"$terraform_plan_flags\""
+  fi
+  if [ ! -z "$usage_file" ]; then
+    breakdown_cmd="$breakdown_cmd --usage-file $usage_file"
+  fi
+  if [ ! -z "$config_file" ]; then
+    breakdown_cmd="$breakdown_cmd --config-file $config_file"
+  fi
+  echo "$breakdown_cmd"
+}
+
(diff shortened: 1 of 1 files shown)


sinabakh:
It would be great if you could take a quick look at it for suggestions or changes before I write the actual documentation.
@alikhajeh1 @aliscott

alikhajeh1:
Thanks @sinabakh! Will review this today :)

alikhajeh1:
This is great! To keep the style of fail_condition similar to post_condition, should we also have this option:
`'{"never": true}': never fail the Jenkins build. This is the default behavior.`

See [this comment](https://github.com/infracost/infracost-gh-action/issues/22#issuecomment-797536663) and the following 2 comments about the pros/cons of why we picked that style.

sinabakh:
Thanks for the review @alikhajeh1. Regarding the suggestion, I think it can't be applied to such a config. I mean, posting a comment all the time makes sense but failing a build is not something that somebody wants unless a specific condition is met. So adding the `never` is somehow a duplication.

Let me know if I'm missing a point or it would helo maintain the same standard.

alikhajeh1:
Good point! What would be the behavior if `fail_condition`:
1. is not defined? I guess we'd never fail the build so there's no point in having an explicit `never` option.
2. is set to `'{"percentage_threshold": 0}'`, we'd fail the build if any cost diff is present, so it enables a safety net to be present if the cost spikes a lot or drops a lot (e.g. someone deletes something).

sinabakh:
That's exactly how you described it.

scripts/ci/jenkins_diff.sh
```diff
@@ -0,0 +1,184 @@
...
+}
+
+process_args () {
+  # Set variables based on the order for GitHub Actions, or the env value for other CIs
+  iac_path=${1:-$iac_path}
+  terraform_plan_flags=${2:-$terraform_plan_flags}
+  terraform_workspace=${3:-$terraform_workspace}
+  usage_file=${4:-$usage_file}
+  config_file=${5:-$config_file}
+  fail_condition=${7:-$fail_condition}
```
alikhajeh1, line 23: I think this can be deleted since env vars are used to run the script (not params passed into the script like GH Action).

scripts/ci/jenkins_diff.sh
```diff
@@ -0,0 +1,184 @@
...
+  # Validate fail_condition
+  if ! echo "$fail_condition" | jq empty; then
+    echo "Error: fail_condition contains invalid JSON"
+  fi
+
+  # Set defaults
+  if [ ! -z "$fail_condition" ] && [ "$(echo "$fail_condition" | jq '.percentage_threshold')" != "null" ]; then
+    fail_percentage_threshold=$(echo "$fail_condition" | jq -r '.percentage_threshold')
+  fi
+  fail_percentage_threshold=${fail_percentage_threshold:--1}
```
alikhajeh1, line 34: hmm I wasn't expecting this to default to -1 since for [other CI integrations](https://github.com/infracost/infracost-atlantis/#post_condition) this is the absolute percentage threshold:
"For example, set to 1 to post a comment if the cost estimate changes by more than plus or minus 1%."

Could we do something similar here? I'm actually not sure how useful the drops in cost posting a comment or failing the build are but I lean towards keeping thins consistent until we have more feedback 🤷‍♂️

scripts/ci/jenkins_diff.sh
```diff
@@ -0,0 +1,184 @@
...
+    breakdown_cmd="$breakdown_cmd --usage-file $usage_file"
+  fi
+  if [ ! -z "$config_file" ]; then
+    breakdown_cmd="$breakdown_cmd --config-file $config_file"
+  fi
+  echo "$breakdown_cmd"
+}
+
+build_output_cmd () {
+  breakdown_path=$1
```
alikhajeh1, line 70: @aliscott I saw this line in diff.sh and atlantis_diff.sh too, is it redundant since it's not used in this function or am I missing something?

scripts/ci/jenkins_diff.sh
```diff
@@ -0,0 +1,184 @@
...
+  msg="${msg}Previous monthly cost: $(format_cost $past_total_monthly_cost)\n"
+  msg="${msg}New monthly cost: $(format_cost $total_monthly_cost)\n"
+  msg="${msg}\n"
+  msg="${msg}Infracost output:\n"
+  msg="${msg}\n"
+  msg="${msg}$(echo "    ${diff_output//$'\n'/\\n    }" | sed "s/%/%%/g")\n"
+  printf "$msg"
+}
+
+build_msg_html () {
```
alikhajeh1, line 114: Maybe add a comment so we remember to switch this once https://github.com/infracost/infracost/issues/509 is released

alikhajeh1 (commented):
@sinabakh thanks for chipping away at this! it's in good shape, just a few minor comments

scripts/ci/jenkins_diff.sh
sinabakh: Oh I see, I actually had a different intention. Since setting the default value to 0 will collide with "Fail if there was any change in diff (setting the threshold to 0)", I set its default value as -1. Do you have any suggestions to enhance it?

alikhajeh1: So we'd want to handle 2 cases:
1. never fail the build. Default behavior. Maybe we can do that by checking if `fail_percentage_threshold` is **not present**?
2. fail the build if the costs increase **or** decrease by X%, can be done using the fail_percentage_threshold

sinabakh: Makes sense, done.

scripts/ci/jenkins_diff.sh
sinabakh: Sure.

scripts/ci/jenkins_diff.sh
sinabakh: Removed it from this script but left the Atlantis for another pr.

scripts/ci/jenkins_diff.sh
sinabakh: Added a todo comment

sinabakh:
@alikhajeh1 All done.

PR #619, opened 2021-04-29
Test/add golden file tests

tim775:


aliscott:
Had a play with this and think it is a great improvement over our current integration tests! I think this makes writing the tests a lot easier, and means we can validate against actual prices listed by the cloud vendors.

I'm a +1 for removing the current integration tests and using these going forward. @alikhajeh1 I would recommend having a play with this as well to see what you think.

Some things that need done before we switch:
1. The PR needs rebased
2. Existing integration tests should be removed.
3. The contributing docs should be changed to add an example of how to add/update these.
4. Should we add a make target for updating the golden tests / maybe even for updating a single one?

alikhajeh1:
Nice work! I also like the ability to quickly see the output from old resources as that helps when adding new resources that have similarities.

If I understand it correctly (plz correct me if I'm wrong):
1. we'd still get failing tests if the production dataset shows a different price vs the golden files. This seems ok as we can verify/update the golden file when prices are reduced.
2. we can still run tests with `INFRACOST_LOG_LEVEL=warn` to catch these issues: "Multiple products found", "No products found for" or "No prices found for".
3. we usually have tests for when a usage-file is not present, to ensure that we're showing the "cost depends on usage" message and not a 0-cost line-item. Should we:
   a) ensure that our TF test files have a resource with no usage data so we can test the output?
   b) or could we automate that by always running the same TF test file twice, once with and once without usage-data, but then we'd have to provide to golden files to check?
   c) or leave things as they are in this PR and take extra care to review for that case in code reviews?

aliscott:
>  1. we'd still get failing tests if the production dataset shows a different price vs the golden files. This seems ok as we can verify/update the golden file when prices are reduced.

I think this is nicer than what we had before, since we get a heads up when prices change. I also don't anticipate them changing regularly.

tim775:
> we'd still get failing tests if the production dataset shows a different price vs the golden files. This seems ok as we can verify/update the golden file when prices are reduced.

Correct.  This could get pretty annoying if prices are changing frequently, but I'm guessing if that happens it will be limited to particular resources and we can revert to the old style tests or figure out something else clever.

> we can still run tests with INFRACOST_LOG_LEVEL=warn to catch these issues: "Multiple products found", "No products found for" or "No prices found for".

Correct.

> we usually have tests for when a usage-file is not present, to ensure that we're showing the "cost depends on usage" message and not a 0-cost line-item. Should we:
a) ensure that our TF test files have a resource with no usage data so we can test the output?
b) or could we automate that by always running the same TF test file twice, once with and once without usage-data, but then we'd have to provide to golden files to check?
c) or leave things as they are in this PR and take extra care to review for that case in code reviews?

My vote would be for c (which I believe is the same for current tests) and make it clear in the contributing docs.  If a lot of PRs are sent back to for missing a non-usage test then maybe think about a or b.

tim775:
@aliscott I made the changes you requested.  Only a couple dozen resources are using the new tests, but I suggest we merge this PR now and create a separate issue for converting the rest of the resources.

aliscott (approved):

Issue #561, opened 2021-04-07
Add Azure App Service Plan and Environment


aliscott:
The initial work has already been done by @DanielMabbett's here: https://github.com/infracost/infracost/pull/457/commits/c0d00adbac5a1bf89ac6984f5eae9db9fce0016e

Azure pricing page: https://azure.microsoft.com/en-gb/pricing/details/app-service/

Resources:

`azurerm_app_service_plan`

* Instance usage (S1) - unit is hours. Quantity is based on `capacity`. We should also allow the user to pass in a `instances` usage variable that overrides this capacity.

`azurerm_app_service_environment` - This is for the Isolated Service Plans.

* Stamp fee - unit is hour. This only applies to v1 instances.
* Instance usage (I1) - unit is hours

Not supported:

* Dev/test pricing - this doesn't seem to be supported by Terraform.

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

PR #621, opened 2021-04-30
feat(azure):Add Azure App Service Plan and Environment 

NeverovDS:
Issue: #561 
Output:

```
azurerm_app_service_environment.my_service                                     
 ├─ Stamp fee                                          730  hours     $1,043.81 
 └─ Instance usage (I1)                                730  hours       $219.00 
                                                                                
 azurerm_app_service_plan.example                                               
 └─ Instance usage (S1)                              3,650  hours       $365.00 
                                                                                
 PROJECT TOTAL                                                        $2,067.81
```

### DIFF
--- a/internal/providers/terraform/azure/app_service_environment.go
+++ b/internal/providers/terraform/azure/app_service_environment.go
@@ -1,0 +1,101 @@
+package azure
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/infracost/infracost/internal/schema"
+
+	"github.com/shopspring/decimal"
+)
+
+func GetAzureRMAppIsolatedServicePlanRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_app_service_environment",
+		RFunc: NewAzureRMAppIsolatedServicePlan,
+		ReferenceAttributes: []string{
+			"resource_group_name",
+		},
+		Notes: []string{
+			"Costs associated with running an app service plan in Azure",
+		},
+	}
+}
+
+func NewAzureRMAppIsolatedServicePlan(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	return &schema.Resource{
+		Name:           d.Address,
+		CostComponents: AppIsolatedServicePlanCostComponent(d, u),
+	}
+}
+func AppIsolatedServicePlanCostComponent(d *schema.ResourceData, u *schema.UsageData) []*schema.CostComponent {
+
+	tier := d.Get("pricing_tier").String()
+	group := d.References("resource_group_name")
+	location := group[0].Get("location").String()
+	stampFeeTiers := []string{"I1", "I2", "I3"}
+	productName := "Isolated Plan"
+	purchaseOption := "Consumption"
+	costComponents := make([]*schema.CostComponent, 0)
+	os := "Linux"
+	if u != nil && u.Get("operating_system").Exists() {
+		os = strings.ToLower(u.Get("operating_system").String())
+	}
+	if os == "Linux" {
+		productName += " - Linux"
+	}
+
+	if Contains(stampFeeTiers, tier) == bool(true) {
+		costComponents = append(costComponents, &schema.CostComponent{
+			Name:           "Stamp fee",
+			Unit:           "hours",
+			UnitMultiplier: 1,
+			HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
+			ProductFilter: &schema.ProductFilter{
+				VendorName:    strPtr("azure"),
+				Region:        strPtr(location),
+				Service:       strPtr("Azure App Service"),
+				ProductFamily: strPtr("Compute"),
+				AttributeFilters: []*schema.AttributeFilter{
+					{Key: "productName", Value: strPtr("Azure App Service " + productName)},
+					{Key: "skuName", Value: strPtr("Stamp")},
+				},
+			},
+			PriceFilter: &schema.PriceFilter{
+				PurchaseOption: strPtr(purchaseOption),
+			},
+		})
+	}
+
+	costComponents = append(costComponents, &schema.CostComponent{
+		Name:           fmt.Sprintf("Instance usage (%s)", tier),
+		Unit:           "hours",
+		UnitMultiplier: 1,
+		HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
+		ProductFilter: &schema.ProductFilter{
+			VendorName:    strPtr("azure"),
+			Region:        strPtr(location),
+			Service:       strPtr("Azure App Service"),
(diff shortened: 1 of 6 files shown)


internal/providers/terraform/azure/app_service_environment.go
```diff
@@ -0,0 +1,101 @@
...
+func GetAzureRMAppIsolatedServicePlanRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_app_service_environment",
+		RFunc: NewAzureRMAppIsolatedServicePlan,
+		ReferenceAttributes: []string{
+			"resource_group_name",
+		},
+		Notes: []string{
+			"Costs associated with running an app service plan in Azure",
+		},
```
alikhajeh1, line 21:
```suggestion
```

This isn't needed since the notes are only to document cases that could not be estimated.

internal/providers/terraform/azure/app_service_environment.go
```diff
@@ -0,0 +1,101 @@
...
+
+func NewAzureRMAppIsolatedServicePlan(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	return &schema.Resource{
+		Name:           d.Address,
+		CostComponents: AppIsolatedServicePlanCostComponent(d, u),
+	}
+}
+func AppIsolatedServicePlanCostComponent(d *schema.ResourceData, u *schema.UsageData) []*schema.CostComponent {
+
+	tier := d.Get("pricing_tier").String()
```
alikhajeh1, line 33: `pricing_tier` [is optional](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service_environment#pricing_tier) in terraform, so this should default to `I1` according to the terraform docs and a check should be done like `if d.Get("pricing_tier").Type != gjson.Null` to set its value

internal/providers/terraform/azure/app_service_environment.go
```diff
@@ -0,0 +1,101 @@
...
+	group := d.References("resource_group_name")
+	location := group[0].Get("location").String()
```
alikhajeh1, line 35: @aliscott [location](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service_environment#location) is mentioned as an Attribute for this resource in Terraform, does that mean this code can just do `d.Get("location").String()` or does it have to get it from `resource_group_name` if it's set and otherwise get it from the `subnet_id` (since that's required).

internal/providers/terraform/azure/app_service_environment.go
```diff
@@ -0,0 +1,101 @@
...
+	stampFeeTiers := []string{"I1", "I2", "I3"}
+	productName := "Isolated Plan"
+	purchaseOption := "Consumption"
+	costComponents := make([]*schema.CostComponent, 0)
+	os := "Linux"
+	if u != nil && u.Get("operating_system").Exists() {
```
alikhajeh1, line 41:
```suggestion
	if u != nil && u.Get("operating_system").Type != gjson.Null {
```

Search the contributing guide for `gjson.Null` to see why we do it this way.

internal/providers/terraform/azure/app_service_environment.go
aliscott: I think it we will have to get it from `resource_group_name` or the `azurerm_virtual_network` of the subnet if we can't find it in the resource group, since attributes are only known after apply.

internal/providers/terraform/azure/app_service_environment.go
```diff
@@ -0,0 +1,101 @@
...
+		os = strings.ToLower(u.Get("operating_system").String())
+	}
+	if os == "Linux" {
```
alikhajeh1, line 44: Does this work if I set the operating system to "linux" (small l) in the usage file?

internal/providers/terraform/azure/app_service_environment_test.go
```diff
@@ -0,0 +1,216 @@
...
+	  value = "1"
+	}
+  }
+`
+	usage := schema.NewUsageMap(map[string]interface{}{
+		"azurerm_app_service_environment.example": map[string]interface{}{
+			"operating_system": "windows",
+		},
+		"azurerm_app_service_environment.example2": map[string]interface{}{
+			"operating_system": "windows",
```
alikhajeh1, line 170: maybe change this test to be "linux" so it checks that case too?

internal/providers/terraform/azure/app_service_plan.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/shopspring/decimal"
+)
+
+func GetAzureRMAppServicePlanRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_app_service_plan",
+		RFunc: NewAzureRMAppServicePlan,
+		Notes: []string{
+			"Costs associated with running an app service plan in Azure",
+		},
```
alikhajeh1, line 17:
```suggestion
```

This isn't needed

internal/providers/terraform/azure/app_service_plan.go
```diff
@@ -0,0 +1,130 @@
...
+
+func AppServicePlanCostComponent(d *schema.ResourceData) []*schema.CostComponent {
+	sku := d.Get("sku.0.size").String()
+	skuRefactor := ""
+	purchaseOption := "Consumption"
+	capacity := d.Get("sku.0.capacity").Int()
+	location := d.Get("location").String()
+	productName := "Standard Plan"
+
+	if sku == "P1v2" {
```
alikhajeh1, line 36: would these if conditions be better as a switch statement?

internal/providers/terraform/azure/app_service_plan.go
```diff
@@ -0,0 +1,130 @@
...
+		productName = "Basic Plan"
+	}
+
+	os := "Windows"
+	if d.Get("kind").Exists() {
+		os = d.Get("kind").String()
+	}
+	if os != "Windows" {
+		productName += " - Linux"
+	}
```
alikhajeh1, line 105: Looking at the Note underneath kind in [this doc](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service_plan#sku), it seems like Windows can also have the value "App". So can you possibly handle that case too?

Also, can you please test if "windows" with small W is accepted by Terraform and if it comes through here as a small W? if so, then maybe always change to lower-case during the if statement to avoid bugs?

alikhajeh1 (changes requested):
@NeverovDS Awesome first PR! I added my suggestions

internal/providers/terraform/azure/app_service_environment.go
NeverovDS: My bad, correcting

PR #626, opened 2021-05-02
chore: refactor virtual_cpu_count to vcpu_count so it matches the cloud

alikhajeh1:
provider terminology


alikhajeh1:
@robh007 just FYI on this

PR #629, opened 2021-05-04
fix: fix workspace param in diff scripts

alikhajeh1:


scripts/ci/jenkins_diff.sh
```diff
@@ -105,13 +109,14 @@ build_msg () {
...
   # TODO: change it to infracost output once https://github.com/infracost/infracost/issues/509 is resolved.
   msg=$1
-  html="<!DOCTYPE html>\n<html>\n<body>\n<pre>"
-  html="${html}\n${msg}"
-  html="${html}</pre>\n</body>\n</html>"
-  printf "$html"
+  html="<!DOCTYPE html><html><body><pre>"
+  html="${html} ${msg}"
+  html="${html}</pre></body></html>"
+  printf '%s' "$html"
```
alikhajeh1, line 115: @sinabakh just FYI on https://www.shellcheck.net, it caught an issue here that was causing this to fail when running on macos with zsh with this:
```
IAC_PATH=terraform FAIL_CONDITION='{"percentage_threshold": 10}' scripts/ci/jenkins_diff.sh
```

Issue #548, opened 2021-04-02
Update HTML output to be consistent with CLI output


aliscott:
We should update the HTML output to be consistent with the new CLI table output

* Support multiple projects. We should update it so it doesn't use the following deprecated output fields: `resources`, `totalMonthlyCost`, `totalHourlyCost`, and instead uses the `projects` field.
* Remove the file grouping logic we have - this can now be done with the multiple project
* Support the `--fields` flag (when it's merged).  This should work the same way the `table` output format works
* Keep the existing tags functionality so we continue showing the tag list under each resource
* Add the `Cost depends on usage: XXX`

Here's a wireframe for what the output would look like:

<img width="840" alt="Screenshot 2021-04-26 at 11 54 39" src="https://user-images.githubusercontent.com/158824/116521804-581ca880-a8cc-11eb-96c7-3db96c9b6b18.png">

**Future**
1. How do we show both the breakdown and diff in the output? See: https://github.com/infracost/infracost/issues/509
2. Support dynamic columns (as per https://github.com/infracost/infracost/issues/524)

alikhajeh1:
Initially let's just have the current HTML output format show a breakdown.

ekifel:
@aliscott, I want to work on it!

alikhajeh1:
Released in v0.9.0

PR #632, opened 2021-05-04
feat(output): update the HTML output

ekifel:
Issue #548 

Updated: 
- HTML output to be consistent with CLI output
- Warnings and `--help` descriptions

Added: 
- Support dynamic columns


internal/output/templates.go
```diff
@@ -157,14 +166,76 @@ iVBORw0KGgoAAAANSUhEUgAAAMAAAADACAMAAABlApw1AAAABGdBTUEAALGPC/xhBQAAAAFzUkdCAK7O
...
+        <td class="price">{{.CostComponent.Price | formatPrice }}</td>
+      {{end}}
+      {{if contains .Fields "hourlyCost"}}
+        <td class="hourly-cost">{{.CostComponent.HourlyCost | formatCost2DP}}</td>
+      {{end}}
+      {{if contains .Fields "monthlyCost"}}
+        <td class="monthly-cost">{{.CostComponent.MonthlyCost | formatCost2DP}}</td>
+      {{end}}
+    {{else}}
+      <td colspan="{{len .Fields}}" class="usage-cost">Cost depends on usage: {{.CostComponent.Price | formatPrice}} per {{.CostComponent.Unit}}</td>
```
aliscott, line 186: Should we make the text color `#6b7280` here so it's a little faded like in the CLI?

internal/output/templates.go
```diff
@@ -157,14 +166,76 @@ iVBORw0KGgoAAAANSUhEUgAAAMAAAADACAMAAABlApw1AAAABGdBTUEAALGPC/xhBQAAAAFzUkdCAK7O
...
+  {{if contains .Fields "monthlyCost"}}
+    <td class="monthly-cost">Monthly Cost</td>
+  {{end}}
+{{end}}
+
+{{define "projectBlock"}}
+  {{$fields := .Options.Fields}}
+  <p class="project-name">Project: {{.Project.Path}}</p>
+  <div class="table-border">
+    <table>
```
aliscott, line 214: Maybe we should add a sensible set width to the table so the tables are the same width.

aliscott (commented):
This looks awesome! Thanks @ekifel.

internal/output/templates.go
```diff
@@ -38,8 +38,8 @@ a {
 }
 
 table {
+  border: 1px solid gray;
```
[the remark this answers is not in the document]
ekifel: Because I changed it while trying to figure out with one problem and forgot about it. I'll bring it back

aliscott:
Thanks @ekifel! 😄  Merging this into the v0.9 branch which we will use going forward for v0.9 features.

PR #633, opened 2021-05-04
feat(azure):Add Azure App Service Plan and Environment #561

NeverovDS:
Output:
```
azurerm_app_service_environment.my_service                                     
 ├─ Stamp fee                                          730  hours       $991.34 
 └─ Instance usage (I1)                                730  hours       $208.05 
                                                                                
 azurerm_app_service_plan.example                                               
 └─ Instance usage (PC2)                               730  hours       $118.59 
                                                                                
 PROJECT TOTAL                                                        $1,757.98 ``
```

### DIFF
--- a/internal/providers/terraform/azure/app_service_plan.go
+++ b/internal/providers/terraform/azure/app_service_plan.go
@@ -1,0 +1,114 @@
+package azure
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/infracost/infracost/internal/schema"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/shopspring/decimal"
+)
+
+func GetAzureRMAppServicePlanRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_app_service_plan",
+		RFunc: NewAzureRMAppServicePlan,
+	}
+}
+
+func NewAzureRMAppServicePlan(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	sku := d.Get("sku.0.size").String()
+	skuRefactor := ""
+	os := "Windows"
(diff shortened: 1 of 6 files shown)


internal/providers/terraform/azure/app_service_plan.go
```diff
@@ -0,0 +1,114 @@
...
+	switch sku[2:] {
+	case "v2":
+		skuRefactor = sku[:2] + " " + sku[2:]
+		productName = "Premium v2 Plan"
+	case "v3":
+		skuRefactor = sku[:2] + " " + sku[2:]
+		productName = "Premium v3 Plan"
+	}
+
+	switch sku {
```
alikhajeh1, line 37: Can this code be made a little more future-proof so when Azure add PC5 or S4 or B4, it continues to work? For example, if the first two letters are PC, then use that + the 3rd letter for the skuRefactor. If it starts with "B" then do something similar, same when it starts with "S"

internal/providers/terraform/azure/app_service_plan.go
```diff
@@ -0,0 +1,114 @@
...
+		os = strings.ToLower(d.Get("kind").String())
+	}
+	if os == "app" {
+		os = "windows"
+	}
+	if os != "windows" {
+		productName += " - Linux"
+	}
+	if sku[:2] == "PC" && os == "linux" {
+		log.Warnf("Skipping resource %s.This tariff plan is not supported for the Linux operating system", d.Address)
```
alikhajeh1, line 78: Is the plan not available in the pricing DB? If so, then there is no need to log a warning as users should never hit that as it's not a valid option in Terraform...

alikhajeh1 (commented):
Good stuff! I left a few minor suggestions

PR #634, opened 2021-05-05
test: Speed up tests (reduce terraform init time) by using a single t…

tim775:
…emp dir per integration test suite run


tim775:
This isn't ready for merge.  I create a PR to make sure the tests are faster in CI.

Issue #598, opened 2021-04-20
Displaying double negative on percentage decreases


swade1987:
See below ...

```
Monthly cost change for /home/atlantis/.atlantis/repos/XXX.tfplan
    Amount:  -$281 ($914 -> $633)
    Percent: --31%
```
Looks like a bug in this function: https://github.com/infracost/infracost/blob/master/internal/output/diff.go#L281

aliscott:
Fixed in https://github.com/infracost/infracost/pull/636, thanks @swade1987 and @ekifel.

alikhajeh1:
@swade1987 this was shipped in v0.8.7 (we've also upgraded the infracost-atlantis image to the latest version of Atlantis and the Infracost 0.8.7 CLI version)

PR #636, opened 2021-05-05
feat(fix): Double negative for percentage in diff output

ekifel:
Issue #598 

Cases:
```
Monthly cost change for .
Amount:  +$3,790 ($0.00 -> $3,790)
```

```
Monthly cost change for .
Amount:  +$2,790 ($1,000 -> $3,790)
Percent: +279%
```

```
Monthly cost change for .
Amount:  -$2,510 ($6,300 -> $3,790)
Percent: -40%
```

### DIFF
--- a/internal/output/diff.go
+++ b/internal/output/diff.go
@@ -284,9 +284,13 @@ func formatPercentChange(oldCost *decimal.Decimal, newCost *decimal.Decimal) str
 	}
 
 	p := newCost.Div(*oldCost).Sub(decimal.NewFromInt(1)).Mul(decimal.NewFromInt(100)).Round(0)
+	percentSym := ""
+	if p.IsPositive() {
+		percentSym = "+"
+	}
 
 	f, _ := p.Float64()
-	return fmt.Sprintf("%s%s%%", getSymForPercent(p), humanize.FormatFloat("#,###.", f))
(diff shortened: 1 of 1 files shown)


internal/output/diff.go
```diff
@@ -294,9 +294,5 @@ func getSym(d decimal.Decimal) string {
 		return "+"
 	}
 
-	if d.IsNegative() {
```
aliscott, line 297: @ekifel I'm not sure this is the correct fix. Looking at your example the Cost change should also be showing a negative not a +.

ekifel: Woops, I did not notice that

ekifel: @aliscott, fixed. I updated PR comment with all of cases

internal/output/diff.go
```diff
@@ -300,3 +300,10 @@ func getSym(d decimal.Decimal) string {
 
 	return ""
 }
+
+func getSymForPercent(d decimal.Decimal) string {
```
aliscott, line 304: I would just put this logic in the `formatPercentChange` function. Having a function called `getSym` and `getSymForPercent` might be a bit misleading, because the `getSymForPercent` is really for displaying a symbol assuming the number isn't absolute.

aliscott (approved):

PR #641, opened 2021-05-06
Add more AWS golden tests

aliscott:


internal/providers/terraform/aws/testdata/acmpca_certificate_authority_test/acmpca_certificate_authority_test.golden
```diff
@@ -0,0 +1,19 @@
+
+ Name                                                     Monthly Qty  Unit      Monthly Cost 
+                                                                                              
+ aws_acmpca_certificate_authority.private_ca                                                  
+ ├─ Private certificate authority                                   1  months         $400.00
```
tim775, line 5: Should there be a test case for a ca with usage privded?

tim775 (commented):
Looks good except that we're missing "without usage" test cases for several resources

internal/providers/terraform/aws/testdata/acmpca_certificate_authority_test/acmpca_certificate_authority_test.golden
aliscott: The following two resources have usage values.

aliscott:
> Looks good except that we're missing "without usage" test cases for several resources

Ah maybe I misinterpreted the discussion about non-usage tests. I thought [option c here](https://github.com/infracost/infracost/pull/619#issuecomment-831310848) meant we shouldn't bother with the non-usage tests. I'll update these to have non-usage ones for each that has usage values.

PR #649, opened 2021-05-10
fix(aws): Fix transit gateway data usage to use MonthlyQuantity for '…

tim775:
…monthly_data_processed_gb'


aliscott:
@tim775 thanks!

Issue #627, opened 2021-05-02
Add support for Azure Functions


alikhajeh1:
Pricing web page: https://azure.microsoft.com/en-gb/pricing/details/functions/
Terraform resource: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/function_app

The `app_service_plan_id` can be followed to see what type of pricing applies:
1. If the `app_service_plan.kind=elastic` OR `app_service_plan.sku.tier=ElasticPremium` then `Premium plan` pricing applies so the following cost components should be shown. The instance type can be fetched from `azurerm_app_service_plan.sku.size` and the supported values are [EP1, 2 and 3](https://docs.microsoft.com/en-gb/azure/azure-functions/functions-premium-plan?tabs=portal#available-instance-skus), we can have a hard-coded map of the CPU/Memory of those 3 types. The `instances` usage file value can be used to calculate the cost.
- Name "vCPU", units "vCPU-hours"
- Name "Memory", units "GB-hours"

2. If the app_service_plan.kind is FunctionApp then `Consumption` pricing applies so the following cost components should be shown. The monthly_executions, execution_duration_ms, and memory_mb usage file values can be used to calculate the cost. 
- Name "Execution time", units "GB-seconds"
- Name "Executions", units "requests"

The usage file should have these params, the top 3 can be used in Consumption and the bottom 2 can be used in Premium plan pricing:
```
    monthly_executions: 100000 # Monthly executions to the function. Only applicable for Consumption plan.
    execution_duration_ms: 500 # Average duration of each execution in milliseconds. Only applicable for Consumption plan.
    memory_mb: 128 # Average amount of memory consumed by function in MB. Only applicable for Consumption plan.
    instances: 2 # Number of instances. Only applicable for Premium plan.
```

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

PR #622, opened 2021-05-01
Feat/make units and quantities understandable

tim775:

PR #653, opened 2021-05-11
feat(azure): Add support for Azure Functions 

NeverovDS:
Issue #627 
Outputs: golden test file.

### DIFF
--- a/internal/providers/terraform/azure/function_app.go
+++ b/internal/providers/terraform/azure/function_app.go
@@ -60,14 +60,14 @@ func NewAzureRMAppFunction(d *schema.ResourceData, u *schema.UsageData) *schema.
 		"EP3": 14.0,
 	}
 
-	appServicePlanId := d.References("app_service_plan_id")
+	appServicePlanID := d.References("app_service_plan_id")
 
-	if len(appServicePlanId) > 0 {
-		kind = strings.ToLower(appServicePlanId[0].Get("kind").String())
-		skuTier = strings.ToLower(appServicePlanId[0].Get("sku.0.tier").String())
+	if len(appServicePlanID) > 0 {
+		kind = strings.ToLower(appServicePlanID[0].Get("kind").String())
+		skuTier = strings.ToLower(appServicePlanID[0].Get("sku.0.tier").String())
 	}
-	if appServicePlanId[0].Get("sku.0.size").Type != gjson.Null {
-		skuSize = appServicePlanId[0].Get("sku.0.size").String()
+	if appServicePlanID[0].Get("sku.0.size").Type != gjson.Null {
+		skuSize = appServicePlanID[0].Get("sku.0.size").String()
 	}
 
 	if val, ok := skuMapCPU[skuSize]; ok {
--- a/internal/providers/terraform/azure/function_app.go
+++ b/internal/providers/terraform/azure/function_app.go
@@ -1,0 +1,188 @@
+package azure
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/infracost/infracost/internal/schema"
+	"github.com/shopspring/decimal"
+	"github.com/tidwall/gjson"
+)
+
+func GetAzureRMAppFunctionRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_function_app",
+		RFunc: NewAzureRMAppFunction,
+		ReferenceAttributes: []string{
+			"app_service_plan_id",
+		},
+	}
+}
+
+func NewAzureRMAppFunction(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	meterNameCPU := "vCPU Duration"
+	meterNameMemory := "Memory Duration"
+	meterNameExecTime := "Execution Time"
+	meterNameExecutions := "Total Executions"
+	memorySize := 1.0
+	executionTime := 1.1
+	executions := 1.1
+	kind := "kind"
+	skuTier := "tier"
+	skuSize := "EP"
+	skuMemory := 3.0
+	skuCPU := 1
+	instances := 1.0
+	location := "someus"
+
+	if d.Get("location").Type != gjson.Null {
+		location = d.Get("location").String()
+	}
+	if u != nil && u.Get("monthly_executions").Type != gjson.Null {
+		executions = u.Get("monthly_executions").Float()
+	}
+	if u != nil && u.Get("execution_duration_ms").Type != gjson.Null {
+		executionTime = u.Get("execution_duration_ms").Float() / 1000
+		fmt.Println(executionTime)
+	}
+	if u != nil && u.Get("memory_mb").Type != gjson.Null {
+		memorySize = u.Get("memory_mb").Float() / 1000
+	}
+
+	skuMapCPU := map[string]int{
+		"EP1": 1,
+		"EP2": 2,
+		"EP3": 4,
+	}
+	skuMapMemory := map[string]float64{
+		"EP1": 3.5,
+		"EP2": 7.0,
+		"EP3": 14.0,
+	}
+
+	app_service_plan_id := d.References("app_service_plan_id")
+
+	if len(app_service_plan_id) > 0 {
+		kind = strings.ToLower(app_service_plan_id[0].Get("kind").String())
+		skuTier = strings.ToLower(app_service_plan_id[0].Get("sku.0.tier").String())
+	}
+	if app_service_plan_id[0].Get("sku.0.size").Type != gjson.Null {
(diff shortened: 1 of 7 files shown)


internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+	}
+}
+
+func NewAzureRMAppFunction(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	meterNameCPU := "vCPU Duration"
+	meterNameMemory := "Memory Duration"
+	meterNameExecTime := "Execution Time"
+	meterNameExecutions := "Total Executions"
+	memorySize := 1.0
```
alikhajeh1, line 27: These never change in this file so you can just hard-code them in the functions that use them instead of defining variables and passing them around. e.g. on line 115, the string can be set directly without using any variables. That makes the code easier to read.

internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+	executionTime := 1.1
+	executions := 1.1
```
alikhajeh1, line 29: These seem like odd default values, why not `var executionTime, executions float64`?

internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+	kind := "kind"
```
alikhajeh1, line 30: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service_plan#kind defaults to "Windows"

internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+	skuTier := "tier"
+	skuSize := "EP"
+	skuMemory := 3.0
+	skuCPU := 1
+	instances := 1.0
+	location := "someus"
```
alikhajeh1, line 36: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/function_app#location is a required parameter so you can just do `location = d.Get("location").String()`, there's no need to check for nil or have a default value...

internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+	meterNameCPU := "vCPU Duration"
+	meterNameMemory := "Memory Duration"
+	meterNameExecTime := "Execution Time"
+	meterNameExecutions := "Total Executions"
+	memorySize := 1.0
+	executionTime := 1.1
+	executions := 1.1
+	kind := "kind"
+	skuTier := "tier"
+	skuSize := "EP"
```
alikhajeh1, line 32: These are also required parameters in app_service_plan, see my comment below on line 36 as the same applies here.

internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+	skuMemory := 3.0
+	skuCPU := 1
+	instances := 1.0
```
alikhajeh1, line 35: These should default to nil so if the user does not define them in the usage file, then we show "cost depends on usage" in the output, not 0 cost.

internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+		PriceFilter: &schema.PriceFilter{
+			PurchaseOption: strPtr("Consumption"),
+		},
+	}
+}
+func AppFunctionConsumptionExecutionTimeCostComponent(executions, executionTime float64, memorySize float64, location, meterName string) *schema.CostComponent {
+	return &schema.CostComponent{
+
+		Name:           "Execution time",
+		Unit:           "GB Seconds",
```
alikhajeh1, line 149:
```suggestion
		Unit:           "GB-seconds",
```

internal/providers/terraform/azure/function_app.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+	}
+}
+func AppFunctionConsumptionExecutionsCostComponent(executions float64, location, meterName string) *schema.CostComponent {
+	return &schema.CostComponent{
+
+		Name:            "Executions",
+		Unit:            "requests",
+		UnitMultiplier:  1,
+		MonthlyQuantity: decimalPtr(decimal.NewFromFloat(executions / 10)),
```
alikhajeh1, line 173: @aliscott / @tim775 could we allow UnitMultiplier to be a float to address the following issue? I've only seen a few resources have this issue so it's fine if we don't change anything just now...

Azure's pricing page mentions "$0.20 per million executions", however, their pricing API returns `"unit" : "10", "USD" : "0.000002"` (i.e. they return the price per 10 requests) 🤦 

So when I enter 1,000,000 in the usage file, I see the following, the price and monthly cost is correct, but I entered 1M requests not 100K.

```
 └─ Executions                                 $0.000002      100,000  requests          $0.20
```

If we use UnitMultiplier 100,000 (as that's how we can represent the Azure pricing API figure in nicer terms), then we see the following, the price and costs map to what the Azure pricing page shows but I entered 1M...
```
 └─ Executions                                 $0.20            1  100k requests        $0.20
```

@NeverovDS no action needed for this comment.

alikhajeh1 (changes requested):
Thanks @NeverovDS! I added my comments

internal/providers/terraform/azure/function_app.go
aliscott: Wouldn't we want to set `UnitMultiplier = 100k`, `Unit = 1M requests`?

As a side note, I think the unit here should be "executions" since that's what Azure show on their pricing page.

alikhajeh1: @aliscott that shows the following, which also seems confusing despite showing the correct price/cost:
```
 └─ Executions                                 $0.20            1  100k 1M requests        $0.20
```

Regarding the units, Azure's pricing calculator shows "requests". I went with that as it can help distinguish requests to executions from the "Execution time" cost component.

aliscott: Okay, I think with @tim775's changes in https://github.com/infracost/infracost/pull/622 then `UnitMultiplier = 100k`, `Unit = 1M requests` will work as expected, but will let him confirm.

internal/providers/terraform/azure/testdata/function_app_test/function_app_test.golden
```diff
@@ -0,0 +1,23 @@
+
+ Name                                Monthly Qty  Unit        Monthly Cost 
+                                                                           
+ azurerm_app_service_plan.elastic                                          
+ └─ Instance usage (EP2)                     730  hours              $0.00
```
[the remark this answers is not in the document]
NeverovDS: @alikhajeh1 app_service_plan here just to take the kind you want for the function_app.App_service_plan itself does not support these kind values, so no product was found.

internal/providers/terraform/azure/function_app.go
alikhajeh1: @NeverovDS in that case plz use `UnitMultiplier = 100000` and `Unit = 1M requests` in this PR

internal/providers/terraform/azure/testdata/function_app_test/function_app_test.golden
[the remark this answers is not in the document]
alikhajeh1: cool - thanks for clarifying! Can you please also add two test cases here for when the usage file doesn't have the resources (one for each of the function types). That should show "cost depends on usage"

internal/providers/terraform/azure/testdata/function_app_test/function_app_test.tf
```diff
@@ -0,0 +1,79 @@
+provider "azurerm" {
+  skip_provider_registration = true
+  features {}
+  subscription_id             = "84171726-c002-4fc7-924a-b6ff82c59677"
+  client_id                   = "2c517011-ceed-4166-9fc3-d1c6e077d0fa"
+  client_secret               = "KhWvOLIR6Zkq4Yf9I_Ceg_jiOJRfCqlMHw"
+  tenant_id                   = "b0b4f93e-59ae-44ff-b39a-e51e7e2b16e9"
```
alikhajeh1, line 7: @NeverovDS please do not commit secrets into git :)

@aliscott can you please rotate these creds?

alikhajeh1:
Thanks @NeverovDS - I'll merge this as soon as the tests passed, plz took a look at my most recent commit for some minor changes I did.

internal/providers/terraform/azure/testdata/function_app_test/function_app_test.tf
NeverovDS: @alikhajeh1  sorry!!this will not happen again!

PR #656, opened 2021-05-12
test(aws): convert integration tests to golden

ekifel:


internal/providers/terraform/aws/testdata/efs_file_system_test/efs_file_system_test.tf
```diff
@@ -0,0 +1,23 @@
+provider "aws" {
```
aliscott, line 1: We should add a test without usage

aliscott (approved):
Looks good. I think one non-usage test is missing.

internal/providers/terraform/aws/testdata/efs_file_system_test/efs_file_system_test.tf
ekifel: Added!

PR #658, opened 2021-05-12
feat(aws): MQ broker

### LINKED ISSUE #495: add support for AWS ActiveMQ
Provide support for the following resources.

```
aws_mq_broker
aws_mq_configuration
```

https://aws.amazon.com/amazon-mq/pricing/

sinabakh:
Fix #495


sinabakh:
Here's a sample output:
```
aws_mq_broker.my_aws_mq_broker_activemq_single                                                      
 ├─ Instance (ActiveMQ, mq.t2.micro)                               730  hours                 $21.90 
 └─ Storage (ActiveMQ, Single-AZ)                                    7  GB-months              $0.70 
```

However, there are two points that I'm not sure about:
1. How one can specify the number of nodes in a cluster in terraform config? Since I couldn't find a way I added the `nodes_count` to the usage file.
2. The data transfer fee seemed like the general transfer rate to me so I skipped it in favor of our global transfer data

Pricing page:
https://aws.amazon.com/amazon-mq/pricing/

Terraform page:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/mq_broker

internal/providers/terraform/aws/mq_broker.go
```diff
@@ -0,0 +1,105 @@
...
+				{Key: "brokerEngine", Value: strPtr(engine)},
+				{Key: "deploymentOption", Value: strPtr(deploymentOption)},
+			},
+		},
+	}
+}
+
+func storage(region, engine string, isActiveStandby bool, storageSizeGB, nodesCount *decimal.Decimal) *schema.CostComponent {
+	if storageSizeGB != nil {
+		freeSizeGB := decimal.NewFromInt(5)
```
alikhajeh1, line 71: This only applies for EFS, and it's not clear if the first 5GB is free for all instances in the first cluster, or just the first instance, or just the first AmazonMQ cluster that's created in an AWS account... So I wonder if it's better to just remove this?

internal/providers/terraform/aws/testdata/mq_broker_test/mq_broker_test.tf
```diff
@@ -0,0 +1,111 @@
...
+  skip_get_ec2_platforms      = true
+  skip_region_validation      = true
+  access_key                  = "mock_access_key"
+  secret_key                  = "mock_secret_key"
+}
+
+resource "aws_security_group" "my_aws_security_group" {
+}
+
+resource "aws_mq_configuration" "my_aws_mq_configuration" {
```
alikhajeh1, line 15: Might be good to also include ebs and efs storage_type options explicitly to test the prices are ok.

internal/providers/terraform/aws/mq_broker.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func NewMQBroker(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := d.Get("region").String()
+	engine := d.Get("engine_type").String()
+	instanceType := d.Get("host_instance_type").String()
+	deploymentMode := d.Get("deployment_mode").String()
+	isActiveStandby := false
+	if deploymentMode == "ACTIVE_STANDBY_MULTI_AZ" {
```
alikhajeh1, line 23: Does `CLUSTER_MULTI_AZ` need to be taken into account too?

alikhajeh1 (commented):
Thanks @sinabakh! I added my suggestions

sinabakh:
@alikhajeh1 Jan all your comments made sense. While debugging the problems I came up with some interesting facts:
* The prices in API contain the number of instances with them. The RabbitMQ's cluster mode always has 3 instances.
* Storage prices are not multiplied and need to always be multiplied by the number of instances.

I did some refactoring in logics and I think it's ready.

alikhajeh1:
Thanks @sinabakh - awesome level of details :) One tiny suggestion:
The capital letters look distracting for the deployment mode (we don't do that elsewhere), maybe "single-instance", "active-standby", "cluster"?

sinabakh:
Thank you @alikhajeh1 for the review. Your comment makes sense so I fixed it.

PR #659, opened 2021-05-13
chore: remove deprecated fields in json output

### LINKED ISSUE #549: Remove deprecated JSON keys from output
Remove the following keys from the JSON output: https://github.com/infracost/infracost/blob/master/internal/output/output.go#L18-L20

The HTML output currently relies on these so this will need to be done with https://github.com/infracost/infracost/issues/548.

sinabakh:
Fix #549


alikhajeh1:
Thanks @sinabakh! I think @aliscott mentioned we want this in the 0.9 integration branch, see https://github.com/infracost/infracost/pull/632 as that was done first to enable this GH issue to be done, looks like `infracost:v0.9` is the integration branch/tag

sinabakh:
Thanks @alikhajeh1 , so my current branch is different. I will fix it.

sinabakh:
Closing in favor of https://github.com/infracost/infracost/pull/662

PR #662, opened 2021-05-13
Remove few output deprecations

### LINKED ISSUE #549: Remove deprecated JSON keys from output

sinabakh:
Fix #549 

Changes:
* Remove deprecated fields
* Remove deprecated table format


cmd/infracost/run.go
```diff
@@ -154,9 +154,6 @@ func runMain(cmd *cobra.Command, cfg *config.Config) error {
 	case "diff":
 		b, err = output.ToDiff(r, opts)
 		out = fmt.Sprintf("\n%s", string(b))
-	case "table_deprecated":
```
aliscott, line 157: We will need to remove the [root_deprecated](https://github.com/infracost/infracost/blob/master/cmd/infracost/root_deprecated.go) command as well since it uses this format. Happy for that to be done in a separate PR though. What do you think?

aliscott (approved):
LGTM thanks @sinabakh!

sinabakh:
@aliscott I removed few more deprecations as you mentioned. Since it initially contained more changes than the original issue I pushed the codes on this branch.
However, the commits are atomic and can be merged without a squash.

aliscott (approved):
This is awesome! Thanks @sinabakh

Issue #648, opened 2021-05-09
Add support for Azure Container Registry


alikhajeh1:
Resources: azurerm_container_registry as the paid resource with these free resources: azurerm_container_registry_scope_map, azurerm_container_registry_token, azurerm_container_registry_webhook

Pricing: https://azure.microsoft.com/en-us/pricing/details/container-registry/

Cost components:
1. "Registry usage (sku_here)", units "days", set MonthlyQuantity to 30 so we assume a 30 day month. sku defaults to Classic according to the terraform-provider-azurerm code (which is the same as Basic).

2. For Premium sku, if there is are georeplication_locations or georeplications, then add a "Geo replication (X locations)" cost component where X is the number of locations (use the word "location" if there is only 1), units "days", set MonthlyQuantity to 30 so we assume a 30 day month.

3. "Storage (over XGB)", units "GB-months", with the following entry in the usage file. X should be the included storage for that sku. If the usage file value is above the "Included storage" then apply the Additional storage cost, otherwise this should show as a usage-based cost component.
`storage_gb: 150 # Total size of registry in GB.`

4. "Build vCPU", units "seconds" (in the pricing API, they have `"meterName" : "Task vCPU Duration"`), with the following entry in the usage file, convert the hrs to seconds for the calculation.
`monthly_build_vcpu_hrs: 150 # Monthly container build hours across all vCPUs.`

Terraform doesn't seem to support Dedicated agent pool yet so no need to include that.

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

PR #664, opened 2021-05-14
feat(azure): Add support for Azure Container Registry 

NeverovDS:
Issue #648 
Output: golden test


internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,181 @@
...
+	if d.Get("georeplication_locations").Exists() {
+		nambersOflocations = len(d.Get("georeplication_locations").Array())
+	}
+	if d.Get("georeplications").Exists() {
+		nambersOflocations = len(d.Get("georeplications.0.location").Array())
+	}
+
+	costComponents := make([]*schema.CostComponent, 0)
+
+	if d.Get("georeplication_locations").Exists() || d.Get("georeplications").Exists() {
```
alikhajeh1, line 51: should this be like `d.Get("value_name").Type != gjson.Null`? basically anytime I see `Exists()`, I think of this from our contributing doc:
In the past we used `.Exists()` but this only checks that the key does not exist in the Terraform JSON, not that the key exists and is set to null.

internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,181 @@
...
+	return &schema.Resource{
+		Name:           d.Address,
+		CostComponents: costComponents,
+	}
+}
+
+func ContainerRegistryCostComponent(name, location, sku string) *schema.CostComponent {
+
+	return &schema.CostComponent{
+
```
alikhajeh1, line 88:
```suggestion
```

internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+		PriceFilter: &schema.PriceFilter{
+			PurchaseOption: strPtr("Consumption"),
+		},
+	}
+}
+func ContainerRegistryGeolocationCostComponent(name, location, sku string) *schema.CostComponent {
+
+	return &schema.CostComponent{
+
```
alikhajeh1, line 112:
```suggestion
```

internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,181 @@
...
+		Name:            name,
+		Unit:            "days",
+		UnitMultiplier:  1,
+		MonthlyQuantity: decimalPtr(decimal.NewFromInt(30)),
```
alikhajeh1, line 116: The Azure pricing page mentions "1.667 per replicated region" so we'll need to multiply this by the number of location

internal/providers/terraform/azure/testdata/container_registry_test/container_registry_test.golden
```diff
@@ -0,0 +1,33 @@
...
+ Name                                           Monthly Qty  Unit            Monthly Cost 
+                                                                                          
+ azurerm_container_registry.basic                                                         
+ ├─ Registry usage (Basic)                               30  days                   $5.00 
+ ├─ Storage (over 10 GB)                                140  GB-months             $14.00 
+ └─ Build vCPU                                      540,000  seconds               $54.00 
+                                                                                          
+ azurerm_container_registry.my_registry                                                   
+ ├─ Geo replication (1 location)                         30  days                  $50.00 
+ └─ Build vCPU                                      540,000  seconds               $54.00
```
alikhajeh1, line 11: The "Registry usage" cost component should always be applied, see https://azure.microsoft.com/en-us/pricing/details/container-registry/#pricing

alikhajeh1 (commented):
Thanks @NeverovDS! This is almost there, just a few minor issues

PR #668, opened 2021-05-14
feat(azure): Add support for Azure Container Registry

NeverovDS:
Issue #648

### DIFF
--- a/internal/providers/terraform/azure/container_registry.go
+++ b/internal/providers/terraform/azure/container_registry.go
@@ -1,0 +1,173 @@
+package azure
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/infracost/infracost/internal/schema"
+	"github.com/shopspring/decimal"
+	"github.com/tidwall/gjson"
+)
+
+func GetAzureRMContainerRegistryRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_container_registry",
+		RFunc: NewAzureRMContainerRegistry,
+	}
+}
+
+func NewAzureRMContainerRegistry(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	var nambersOflocations int
+	var storageGb, includedStorage, monthlyBuildVcpu *decimal.Decimal
+	var overStorage decimal.Decimal
+
+	sku := "Classic"
+	location := d.Get("location").String()
+
+	if d.Get("sku").Exists() {
+		sku = d.Get("sku").String()
+	}
+
+	switch sku {
+	case "Classic":
+		includedStorage = decimalPtr(decimal.NewFromFloat(10))
+	case "Basic":
+		includedStorage = decimalPtr(decimal.NewFromFloat(10))
+	case "Standard":
+		includedStorage = decimalPtr(decimal.NewFromFloat(100))
+	case "Premium":
+		includedStorage = decimalPtr(decimal.NewFromFloat(500))
+	}
+
+	if d.Get("georeplication_locations").Exists() {
+		nambersOflocations = len(d.Get("georeplication_locations").Array())
+	}
+	if d.Get("georeplications").Exists() {
+		nambersOflocations = len(d.Get("georeplications.0.location").Array())
+	}
+
+	costComponents := make([]*schema.CostComponent, 0)
+
+	if d.Get("georeplication_locations").Type != gjson.Null || d.Get("georeplications").Type != gjson.Null {
+		if nambersOflocations == 1 {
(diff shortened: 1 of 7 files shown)


internal/providers/terraform/azure/testdata/container_registry_test/container_registry_test.golden
```diff
@@ -0,0 +1,35 @@
...
+ ├─ Geo replication (1 location)                         30  days                  $50.00 
+ ├─ Registry usage (Premium)                             30  days                  $50.00 
+ └─ Build vCPU                                      540,000  seconds               $54.00 
+                                                                                          
+ azurerm_container_registry.overStorage                                                   
+ ├─ Registry usage (Basic)                               30  days                   $5.00 
+ ├─ Storage (over 10 GB)                                540  GB-months             $54.00 
+ └─ Build vCPU                                      540,000  seconds               $54.00 
+                                                                                          
+ azurerm_container_registry.twoLocations
```
alikhajeh1, line 19: "If the usage file value is above the "Included storage" then apply the Additional storage cost, otherwise this should show as a usage-based cost component."

So Storage should always be shown, if there is a usage file property then it should show costs, otherwise it should show "usage depend on cost..."

internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,173 @@
...
+		includedStorage = decimalPtr(decimal.NewFromFloat(10))
+	case "Basic":
+		includedStorage = decimalPtr(decimal.NewFromFloat(10))
+	case "Standard":
+		includedStorage = decimalPtr(decimal.NewFromFloat(100))
+	case "Premium":
+		includedStorage = decimalPtr(decimal.NewFromFloat(500))
+	}
+
+	if d.Get("georeplication_locations").Exists() {
```
alikhajeh1, line 42:
```suggestion
	if d.Get("georeplication_locations").Type != gjson.Null {
```

internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,173 @@
...
+		nambersOflocations = len(d.Get("georeplication_locations").Array())
+	}
+	if d.Get("georeplications").Exists() {
```
alikhajeh1, line 45:
```suggestion
	if d.Get("georeplications").Type != gjson.Null() {
```

internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,173 @@
...
+
+func NewAzureRMContainerRegistry(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	var nambersOflocations int
+	var storageGb, includedStorage, monthlyBuildVcpu *decimal.Decimal
+	var overStorage decimal.Decimal
+
+	sku := "Classic"
+	location := d.Get("location").String()
+
+	if d.Get("sku").Exists() {
```
alikhajeh1, line 27:
```suggestion
	if d.Get("sku").Type != gjson.Null() {
```

alikhajeh1 (commented):
Thanks @NeverovDS - I spotted one small issue, I'll merge after that's resolved

internal/providers/terraform/azure/testdata/container_registry_test/container_registry_test.golden
NeverovDS: @alikhajeh1 Storage should always be shown, if there is a usage file property and value above "Included strorage", right?

alikhajeh1: hmm good point, so there are 3 cases, here's what I think is nice:
1. no usage file: show "Storage (over XGB)" and "cost depends on usage"
2. usage file with value less than included storage. Don't show Storage cost as it's free.
3. usage file with value more than included storage. Show "Storage (over XGB)" and costs.

internal/providers/terraform/azure/container_registry.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	costComponents = append(costComponents, ContainerRegistryCostComponent(fmt.Sprintf("Registry usage (%s)", sku), location, sku))
+
+	if u != nil && u.Get("storage_gb").Type != gjson.Null {
+		storageGb = decimalPtr(decimal.NewFromFloat(u.Get("storage_gb").Float()))
+		if storageGb.GreaterThan(*includedStorage) {
+			overStorage = storageGb.Sub(*includedStorage)
+			storageGb = &overStorage
+			costComponents = append(costComponents, ContainerRegistryStorageCostComponent(fmt.Sprintf("Storage (over %s GB)", includedStorage), location, sku, storageGb))
```
alikhajeh1, line 67:
```suggestion
			costComponents = append(costComponents, ContainerRegistryStorageCostComponent(fmt.Sprintf("Storage (over %sGB)", includedStorage), location, sku, storageGb))
```

alikhajeh1:
@NeverovDS awesome job! This looked like an easy resource at first but it had surprisingly complicated pricing rules

Issue #591, opened 2021-04-17
Add Azure Kubernetes Service AKS


NissesSenap:
Pricing page: https://azure.microsoft.com/en-gb/pricing/details/kubernetes-service/

Make it possible to calculate how much a AKS cluster costs.

A basic graphql query to get info about AKS pricing.
```
query {
  products(
    filter: {
      vendorName: "azure",
      service: "Azure Kubernetes Service",
      region: "centralus",
      attributeFilters: []
    },
  ) {
    attributes { key, value }
    prices { USD }
  }
}
```

alikhajeh1:
Thanks for adding this! I haven't used [`azurerm_kubernetes_cluster`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster) or [`azurerm_kubernetes_cluster_node_pool`](azurerm_kubernetes_cluster_node_pool) before but it look similar to how we added the Google Kubernetes Engine service in that the cluster has a node pool, see `internal/providers/terraform/google/container_cluster.go` and `internal/providers/terraform/google/container_node_pool.go` (and their corresponding [usage-file](https://github.com/infracost/infracost/blob/master/infracost-usage-example.yml) params). For that, we needed re-used the cost components from the compute and disk resources. The [Azure price calculator](https://azure.microsoft.com/en-gb/pricing/calculator/?service=kubernetes-service) shows a similar thing (nodes and managed disks, both of which we support already but we'd need to refactor/re-use). We can start with the pay-as-you-go prices.

It looks like the cluster itself just has one cost component:
- If `sku_tier=Paid`, add the cost component name "Uptime SLA" with units "cluster-hours".

If this is the first time you're adding a resource to Infracost, you might prefer to add a simpler resource such as [this](https://github.com/infracost/infracost/issues/596) to go through the end-to-end process as an onboarding step.

aliscott:
Here's the pricing I've found for AKS.

We might want to add support for the following services before we tackle this:
* [Azure Load Balancers](https://azure.microsoft.com/en-us/pricing/details/load-balancer/)
* [Azure DNS](https://azure.microsoft.com/en-gb/pricing/details/dns/#pricing)
* [Azure Monitoring](https://azure.microsoft.com/en-gb/pricing/details/monitor/)

If not, then I've split this into initial support and future support, so we can add a baseline level of support for this initially.

### Initial support

**azurerm_kubernetes_cluster**

Cost components:
* `Uptime SLA`
[...]

PR #677, opened 2021-05-17
feat(azure): Add support for Azure Kubernetes Service

ekifel:
Issue #591 

Initial support.

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -477,7 +477,7 @@ resource_usage:
     default_node_pool:
       nodes: 2 # Node count for the default node pool.
 
-  azurerm_kubernetes_cluster_node_pool.my_node:
+  azurerm_kubernetes_cluster_node_pool.my_node_pool:
     nodes: 3 # Node count for the node pool.
   
   azurerm_container_registry.my_registry:
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -472,6 +472,13 @@ resource_usage:
     execution_duration_ms: 500 # Average duration of each execution in milliseconds. Only applicable for Consumption plan.
     memory_mb: 128 # Average amount of memory consumed by function in MB. Only applicable for Consumption plan.
     instances: 2 # Number of instances. Only applicable for Premium plan.
+
+  azurerm_kubernetes_cluster.my_cluster:
+    default_node_pool:
+      nodes: 2 # Node count for the default node pool.
+
+  azurerm_kubernetes_cluster_node_pool.my_node:
+    nodes: 3 # Node count for the node pool.
   
   azurerm_container_registry.my_registry:
     storage_gb: 150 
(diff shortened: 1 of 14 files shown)


ekifel:
@aliscott, Hey, this is ready for review!

internal/providers/terraform/azure/virtual_machine.go
```diff
@@ -56,3 +57,26 @@ func osDiskSubResource(region string, d *schema.ResourceData, u *schema.UsageDat
 		CostComponents: managedDiskCostComponents(region, diskData, monthlyDiskOperations),
 	}
 }
+
+func aksOSDiskSubResource(region string, diskSize int, u *schema.UsageData) *schema.Resource {
```
aliscott, line 61: I think this func should be moved to `kubernetes_cluster_node_pool.go`.

internal/providers/terraform/azure/kubernetes_cluster_node_pool.go
```diff
@@ -0,0 +1,64 @@
...
+			diskSize = int(d.Get("os_disk_size_gb").Int())
+		}
+		osDisk := aksOSDiskSubResource(location, diskSize, u)
+
+		if osDisk != nil {
+			subResources = append(subResources, osDisk)
+			schema.MultiplyQuantities(subResources[0], nodeCount)
+			mainResource.SubResources = subResources
+		}
+	}
```
aliscott, line 61: I wonder if we should separate out this logic into a function called `aksClusterNodePool`:
```go
nodeCount := decimal.NewFromInt(1)
if d.Get("node_count").Type != gjson.Null {
  nodeCount = decimal.NewFromInt(d.Get("node_count").Int())
}
if u != nil && u.Get("nodes").Exists() {
  nodeCount = decimal.NewFromInt(u.Get("nodes").Int())
}

r := aksClusterNodePool(d.Address, location, d.RawValues, nodeCount)
```


That way we can re-use the `aksClusterNodePool` function in `kubernetes_cluster.go` like this to avoid repeated logic.

```go
subResources = []*schema.Resource{
  aksClusterNodePool("default_node_pool", location, d.Get("default_node_pool.0"), nodeCount),
}
```

What do you think?

infracost-usage-example.yml
```diff
@@ -472,6 +472,13 @@ resource_usage:
     execution_duration_ms: 500 # Average duration of each execution in milliseconds. Only applicable for Consumption plan.
     memory_mb: 128 # Average amount of memory consumed by function in MB. Only applicable for Consumption plan.
     instances: 2 # Number of instances. Only applicable for Premium plan.
+
+  azurerm_kubernetes_cluster.my_cluster:
+    default_node_pool:
+      nodes: 2 # Node count for the default node pool.
+
+  azurerm_kubernetes_cluster_node_pool.my_node:
```
aliscott, line 480:
```suggestion
  azurerm_kubernetes_cluster_node_pool.my_node_pool:
```
```suggestion
  azurerm_kubernetes_cluster_node_pool.my_node:
```

aliscott (approved):
Thanks @ekifel this is brilliant! 😄

internal/providers/terraform/azure/kubernetes_cluster_node_pool.go
ekifel: @aliscott, thanks! I think it is a good case :) Did it, ready for review!

aliscott:
LGTM!

Issue #669, opened 2021-05-17
Add support for Azure Cosmos DB


alikhajeh1:
Terraform resources: azurerm_cosmosdb* (there are around 12 resources) but they all have the same pricing model mentioned in https://azure.microsoft.com/en-us/pricing/details/cosmos-db/#pricing so we can probably use the same logic between those resources

ekifel:
@alikhajeh1, Hey, I want to work on it

ekifel:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
@ekifel the master branch test logs are showing this, can you please check?
```
time="2021-05-26T17:14:14Z" level=warning msg="Multiple products found for azurerm_cosmosdb_cassandra_keyspace.serverless Provisioned throughput (serverless), using the first product"

```

PR #696, opened 2021-05-20
feat(azure): Add support for Azure CosmosDB

ekifel:
Issue #669 

Initial support
Added main logic and `azurerm_cosmosdb_cassandra_keyspace` resource


ekifel:
@alikhajeh1 this is ready for initial review :)

internal/providers/terraform/azure/testdata/cosmosdb_cassandra_keyspace_test/cosmosdb_cassandra_keyspace_test.golden
```diff
@@ -0,0 +1,72 @@
...
+ ├─ Analytical Storage (Japan West)               Cost depends on usage: $0.03 per GB               
+ ├─ Write operations (Japan West)                 Cost depends on usage: $0.05 per 10K operations   
+ ├─ Read operations (Japan West)                  Cost depends on usage: $0.005 per 10K operations  
+ ├─ Pereodic backup (West US)                     Cost depends on usage: $0.15 per GB               
+ ├─ Pereodic backup (Central US)                  Cost depends on usage: $0.15 per GB               
+ ├─ Pereodic backup (Japan West)                  Cost depends on usage: $0.16 per GB               
+ └─ Restored data                                 Cost depends on usage: $0.15 per GB               
+                                                                                                    
+ azurerm_cosmosdb_cassandra_keyspace.provisioned                                                    
+ ├─ Request units (provisioned, West US)                    3,650  hours                     $29.20
```
tim775, line 26: For provisioned I think the displayed unit  should be "RU/s x 100" and then set `UnitMultiplier: schema.HourToMonthUnitMultiplier,` so it displays as
```
├─ Request units (provisioned, West US)                    500  RU/s x 100                     $29.20 
```
@alikhajeh1 thoughts?

tim775 (changes requested):

internal/providers/terraform/azure/testdata/cosmosdb_cassandra_keyspace_test/cosmosdb_cassandra_keyspace_test.golden
```diff
@@ -0,0 +1,72 @@
+
+ Name                                                 Monthly Qty  Unit                Monthly Cost 
+                                                                                                    
+ azurerm_cosmosdb_cassandra_keyspace.autoscale                                                      
+ ├─ Request units (provisioned, West US)                   43,800  hours                    $350.40
```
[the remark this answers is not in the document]
alikhajeh1: If there is one price point, then we could consider it as one cost component. Also, if there's one usage file param then we can use that to calculate the total cost (and using the min for autoscale, "a minimum of 4,000 RU/s must be provisioned for autoscale provisioned throughput").

For the non-serverless options, is the name "Provisioned throughput (autoscale)"  and "Provisioned throughput (standard)" more suitable as it matches the Azure pricing webpage headings? With the units being "RU/s" as you mentioned. But if it's all just request units regardless of the type then it doesn't make a difference.

internal/providers/terraform/azure/testdata/cosmosdb_cassandra_keyspace_test/cosmosdb_cassandra_keyspace_test.golden
alikhajeh1: +1, that seems clearer to understand as it maps to the pricing webpage.

internal/providers/terraform/azure/testdata/cosmosdb_cassandra_keyspace_test/cosmosdb_cassandra_keyspace_test.golden
```diff
@@ -0,0 +1,72 @@
...
+ ├─ Pereodic backup (West US)                               1,000  GB                       $150.00 
+ ├─ Pereodic backup (Central US)                            1,000  GB                       $150.00 
+ ├─ Pereodic backup (Japan West)                            1,000  GB                       $160.00 
+ └─ Restored data                                           3,000  GB                       $450.00 
+                                                                                                    
+ azurerm_cosmosdb_cassandra_keyspace.serverless                                                     
+ ├─ Requests (serverless)                                      10  1M units                   $2.50 
+ ├─ Transactional storage (West US)                         1,000  GB                       $250.00 
+ ├─ Analytical Storage (West US)                            1,000  GB                        $30.00 
+ ├─ Write operations (West US)                    Cost depends on usage: $0.06 per 10K operations
```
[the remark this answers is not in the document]
alikhajeh1: "Analaytical write operations" (we try to only capitalize the first word, unless it's for an acronym), same with Analytical Storage -> Analytical storage

infracost-usage-example.yml
```diff
@@ -476,6 +476,13 @@ resource_usage:
     memory_mb: 128 # Average amount of memory consumed by function in MB. Only applicable for Consumption plan.
     instances: 2 # Number of instances. Only applicable for Premium plan.
 
+  azurerm_cosmosdb_cassandra_keyspace.my_cassandra:
+    storage_gb: 1000 # Total size of storage in GB.
+    monthly_serverless_request_units: 10000000 # Monthly number of serverless request units.
+    restored_data_gb: 3000 # Size of point-in-time restore data in GB.
+    monthly_write_operations: 1000000 # Monthly number of write analytical storage operations.
```
alikhajeh1, line 483: We try to keep the usage file param names consistent with the cost component names, so the operations ones would be:
monthly_analytical_storage_write_operations
monthly_analytical_storage_read_operations

infracost-usage-example.yml
```diff
@@ -476,6 +476,13 @@ resource_usage:
     memory_mb: 128 # Average amount of memory consumed by function in MB. Only applicable for Consumption plan.
     instances: 2 # Number of instances. Only applicable for Premium plan.
 
+  azurerm_cosmosdb_cassandra_keyspace.my_cassandra:
+    storage_gb: 1000 # Total size of storage in GB.
+    monthly_serverless_request_units: 10000000 # Monthly number of serverless request units.
+    restored_data_gb: 3000 # Size of point-in-time restore data in GB.
```
alikhajeh1, line 482: I think this should be monthly since it's a non-continuous resources, meaning it's the total amount of data that was restored in each month (whereas storage_gb is used continuously in time so it doesn't make sense to define it as as monthly).
```suggestion
    monthly_restored_data_gb: 3000 # Monthly total amount of point-in-time restore data in GB.
```

internal/providers/terraform/azure/testdata/cosmosdb_cassandra_keyspace_test/cosmosdb_cassandra_keyspace_test.golden
```diff
@@ -0,0 +1,44 @@
+
+ Name                                                                   Monthly Qty  Unit                      Monthly Cost 
+                                                                                                                            
+ azurerm_cosmosdb_cassandra_keyspace.autoscale                                                                              
+ ├─ Provisioned throughput (standard, West US)                                   60  RU/s x 100                     $350.40
```
tim775, line 5: I think this is still not correct because the cost is being calculated based on the max_throughput, not a usage parameter.

internal/providers/terraform/azure/testdata/cosmosdb_cassandra_keyspace_test/cosmosdb_cassandra_keyspace_test.golden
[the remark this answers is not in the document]
tim775: I think I vote for not worrying about the minimum/maximum.  Just ignore the value of `autoscale.throughput` entirely and make `"Provisioned throughput (autoscale)"` a usage based cost.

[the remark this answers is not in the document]
alikhajeh1: have you seen the cosmosdb calculation in https://azure.microsoft.com/en-gb/pricing/calculator/ ? would it help if we do the same thing but always assume 730 hours (so the user defines `max_request_units_per_second` and `max_request_units_utilization_percentage`)?

alikhajeh1:
Is this ready to merge? going to cut a release tag soon

internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go
```diff
@@ -89,48 +85,63 @@ func provisionedCosmosCostComponents(model modelType, throughputs *decimal.Decim
...
 	}
 
+	name := "Provisioned throughput"
+	if model == Autoscale {
+		name = fmt.Sprintf("%s (autoscale", name)
+
+		if u != nil && u.Get("max_request_units_per_second").Exists() {
+			throughputs = decimalPtr(decimal.NewFromInt(u.Get("max_request_units_per_second").Int()))
+
+			if u.Get("max_request_units_utilization_percentage").Exists() {
```
tim775, line 95: I like `max_request_units_utilization_percentage` that makes a lot of sense.  For `max_request_units_per_second` previously we were getting this value from `throughputs = decimalPtr(decimal.NewFromInt(d.Get("autoscale_settings.0.max_throughput").Int()))` what's the reason we can't still use that?

tim775: Also:
If `max_request_units_utilization_percentage` is not set, we should show the 'Monthly cost depends on usage...'.  Right now if the precentage is not set it shows the max possible cost.  That doesn't seem correct.

ekifel: No reason, my bad. Brought it back and refactored it to show `'Monthly cost depends on ...'` if `max_request_units_utilization_percentage` is not set.

tim775: @alikhajeh1 LGTM

alikhajeh1:
Thanks @ekifel and @tim775!

Issue #672, opened 2021-05-17
Add support for Azure Firewall


alikhajeh1:
Terraform resources: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/firewall. The other 5 azurerm_firewall* resources should be added as free resources
Pricing details: https://azure.microsoft.com/en-us/pricing/details/azure-firewall/

ekifel:
@alikhajeh1, Hey, I want to work on it! :)

alikhajeh1:
@ekifel I noticed that the `attributes.skuName` mongo field has "Secured Virtual Hub" and "Premium Secured Virtual Hub" options. Can you please send a new PR that checks if `azurerm_firewall.virtual_hub` is set? if so, then these two options can be used instead of the existing ones. That should cover the "Azure Firewall with Secured Virtual Hub" [pricing section](https://azure.microsoft.com/en-us/pricing/details/azure-firewall/#pricing) - the prices are all the same just now but this new logic would ensure that if the prices change, the code works as expected

PR #697, opened 2021-05-20
feat(azure): Add support for Azure Firewall with Secured Virtual Hub

ekifel:
Issue #672 

Details:
I compare `d.Get()` with empty array because defaulty exists empty array of virtual hub block: `"virtual_hub":[]` and it means that `d.Get("virtual_hub").Type` will never return `gjson.Null`

### DIFF
--- a/internal/providers/terraform/azure/firewall.go
+++ b/internal/providers/terraform/azure/firewall.go
@@ -24,6 +24,14 @@ func NewAzureFirewall(d *schema.ResourceData, u *schema.UsageData) *schema.Resou
 		skuTier = d.Get("sku_tier").String()
 	}
 
+	if v := d.Get("virtual_hub").String(); v != "[]" {
+		if skuTier == "Standard" {
+			skuTier = "Secured Virtual Hub"
+		} else {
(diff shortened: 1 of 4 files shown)


ekifel:
@alikhajeh1 this is ready for review!

internal/providers/terraform/azure/firewall.go
```diff
@@ -24,6 +24,14 @@ func NewAzureFirewall(d *schema.ResourceData, u *schema.UsageData) *schema.Resou
 		skuTier = d.Get("sku_tier").String()
 	}
 
+	if v := d.Get("virtual_hub").String(); v != "[]" {
```
tim775, line 27: I think add your explanation from the pr as a code comment here.  "I compare d.Get() with empty array because defaulty exists empty array of virtual hub block: "virtual_hub":[] and it means that d.Get("virtual_hub").Type will never return gjson.Null"

tim775 (commented):
Looks good!

tim775:
@alikhajeh1 LGTM

Issue #671, opened 2021-05-17
Add support for Azure Databricks


alikhajeh1:
Terraform resource: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/databricks_workspace
Pricing details: https://azure.microsoft.com/en-us/pricing/details/databricks/

We can support the Pay as you go pricing for DBUs only initially (not the VM or storage costs as those require other non-Azure terraform resources to be processed). Use the Terraform `sku` field to find which pricing applies, return `NoPrice=true,IsSkipped=true` if the sku is trial as those are free. There should be three cost components, with the following usage file params:
- "All-purpose compute DBUs", units "DBU-hours"
- "Jobs compute DBUs", units "DBU-hours"
- "Jobs light compute DBUs", units "DBU-hours"

Usage file should have (if any are not set, show usage depends on costs...):
```
monthly_all_purpose_compute_dbu: 500 # Monthly number of All-purpose Compute Databricks Units (DBUs).
monthly_jobs_compute_dbu: 1000 # Monthly number of Jobs Compute Databricks Units (DBUs).
monthly_jobs_light_compute_dbu: 2000 # Monthly number of Jobs Light Compute Databricks Units (DBUs).
```

ekifel:
@alikhajeh1, Hey, I want to work on it!

PR #698, opened 2021-05-20
feat(azure): Add support for Azure Databricks

ekifel:
Issue #671

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -469,6 +469,11 @@ resource_usage:
   #
   azurerm_app_service_environment.my_service:
      operating_system: linux # Override the operating system of the instance, can be: linux, windows.
+
+  azurerm_databricks_workspace.my_workspace:
+    monthly_all_purpose_compute_dbu: 500 # Monthly number of All-purpose Compute Databricks Units (DBUs).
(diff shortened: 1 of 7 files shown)


infracost-usage-example.yml
```diff
@@ -469,6 +469,11 @@ resource_usage:
   #
   azurerm_app_service_environment.my_service:
      operating_system: linux # Override the operating system of the instance, can be: linux, windows.
+
+  azurerm_databricks_workspace.my_workspace:
+    monthly_all_purpose_compute_dbu: 500 # Monthly number of All-purpose Compute Databricks Units (DBUs).
```
tim775, line 474: @alikhajeh1  I'm wondering if these should `..._dbu_hours`.  The azure pricing pages talks about DBUs in DBU-month, DBU-hour, DBU-sec.  

At a minimum we should update the comment to say that it is DBU-hour

alikhajeh1: good point, all three can be changed to be line this (we mention _hrs in the contributing doc for usage file params):
```
monthly_all_purpose_compute_dbu_hrs: 500 # Monthly number of All-purpose Compute Databricks Units in DBU-hours.
```

tim775:
@alikhajeh1 LGTM

PR #701, opened 2021-05-21
Test aws add data transfer test

sinabakh:
Changes:
* Add `transfer_data` golden file tests
* Fix a typo in usage file examples


alikhajeh1:
Thanks @sinabakh! since the typo meant that previous users were not able to use that usage file param, it feels like a non-breaking change...

Issue #702, opened 2021-05-22
Add support for AWS Step functions.


robh007:
Terraform providers a resource `aws_sfn_state_machine` for AWS Step functions. Infracost should be able to take the workflow type & then calculate the cost based on the state machine definition. 

Link to pricing page.
https://aws.amazon.com/step-functions/pricing/

praveenghuge:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
Thanks @praveenghuge, https://calculator.aws/#/createCalculator/StepFunctions has clues as to what our usage file params could be, the `type` terraform field can tell us which pricing applies.

alikhajeh1:
> then calculate the cost based on the state machine definition.

@robh007 do you mean we can parse the aws_sfn_state_machine.definition JSON to calculate something? I haven't used the resource but I see that @praveenghuge has added a usage file param (which is the same thing that the AWS calculator does).

robh007:
> > then calculate the cost based on the state machine definition.
> 
> @robh007 do you mean we can parse the aws_sfn_state_machine.definition JSON to calculate something? I haven't used the resource but I see that @praveenghuge has added a usage file param (which is the same thing that the AWS calculator does).

Yes - not sure how complex it would be. But I think you can take the state definition & the workflow type and work out how many transitions occur, you're billed based on these calculations. I'm happy to help if required.

alikhajeh1:
The basic form of this was merged as part of https://github.com/infracost/infracost/pull/704 / https://github.com/infracost/infracost/pull/720. Going to close this ticket, feel free to create another ticket with the advanced version details to calculate the transitions (I think there was some Slack discussion on it too that might be useful in the new ticket).

alikhajeh1:
Shipped in v0.8.7

PR #704, opened 2021-05-23
Add Support for Step function

praveenghuge:
https://github.com/infracost/infracost/issues/702


alikhajeh1:
Thanks @praveenghuge! This is a good start. I added some details below to help get it over the finish line - this is similar to what we did for Lambda. Plz lmk if you have any questions.

Usage file params (add to `infracost-usage-example.yml`):
```
aws_sfn_state_machine.my_sfn_state_machine:
  monthly_transitions: 1000 # Monthly number of state transitions. Only applicable for Standard Workflows.
  monthly_requests: 10000 # Monthly number of workflow requests. Only applicable for Express Workflows.
  memory_mb: 128 # Average amount of memory consumed by workflow in MB. Only applicable for Express Workflows.
  workflow_duration_ms: 500 # Average duration of workflow in milliseconds. Only applicable for Express Workflows.
```

Show this one cost component for Standard workflows:
- "Transitions", units "1K transitions", unit multiplier is 1000. If `monthly_transitions` not set, this should show "cost depends on usage". If `monthly_transitions` is <= 4000, this should return "IsSkipped: true, NoPrice: true" as it's a free resource. If `monthly_transitions` is more than 4000, this should show the cost.

Show these cost components for Express workflows:
Usage file logic: round-**up** memory_mb to next 64mb, round-**up** duration to nearest 100ms.

- "Requests", units "1M requests", unit multiplier is 1000000. If `monthly_requests` not set, this should show "cost depends on usage". Otherwise it should show costs.

Duration calculation (taken from [AWS calc](https://calculator.aws/#/createCalculator/StepFunctions) > Express Workflows pricing):
```
Round up by 64 MB (120) = 128 MB of memory consumed by each workflow
MB of memory consumed by each workflow rounded to 64 MB chunks: 128
Round up by 100 milliseconds (500) = 500 total duration (milliseconds)
Duration rounded up to nearest 100 milliseconds: 500
1 workflow requests x 0.001 seconds x 500 milliseconds = 0.50 seconds of execution
0.50 seconds of execution x 128 MB of memory consumed by each workflow = 64.00 MB-seconds
64.00 MB-seconds / 1024 GB = 0.06 GB-seconds
1 workflow requests x 0.000001 USD = 0.00 USD for monthly request charges
```

- "Duration (first 1K)", units "GB-seconds", unit multiplier is 1. If a usage param is missing and this can't be calculated, show all 3 duration cost components.
- "Duration (next 4K)", units "GB-seconds", unit multiplier is 1. When usage params are available, only show if this tier applies.
- "Duration (over 5K)", units "GB-seconds", unit multiplier is 1. When usage params are available, only show if this tier applies.

praveenghuge:
@alikhajeh1 there is no price available in database for memory_mb for step function express workflow.

alikhajeh1:
@praveenghuge thanks for the changes! Notice how in my previous comment I pasted the "Duration calculation" details for Express workflow - that can be coded in this PR so the cost calculation matches AWS' pricing rules. That way developers enter 3 simple params (monthly_requests, memory_mb and workflow_duration_ms) and we calculate the cost components.

alikhajeh1:
This PR was closed in favor of https://github.com/infracost/infracost/pull/720 (which has these commits except the last one)

Issue #675, opened 2021-05-17
Add support for Azure legacy VMs


alikhajeh1:
Terraform resources: azurerm_virtual_machine and azurerm_virtual_machine_scale_set
Pricing details: https://azure.microsoft.com/en-us/pricing/details/virtual-machines/windows/ and https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/

Ping @aliscott before starting to work on this as I think he worked on the non-legacy versions of these resources

aliscott:
What I've found so far:

To find if it's a windows or Linux VM we should assume Linux for default. If either `storage_os_disk.os_type == "Windows"` or `storage_image_reference.offer == "WindowsServer"` then we should use Windows. We can then use the existing `linuxVirtualMachineCostComponent` and `windowsVirtualMachineCostComponent` cost components.

We should also add a cost component for `Ultra disk reservation (if unattached)` using the existing `ultraSSDReservationCostComponent` function.

We should also have the following subresources:
* `storage_os_disk` is similar to the `osDiskSubResource`, using the `managedDiskCostComponents` function. We could either copy the logic from here or refactor the `managedDiskCostComponents` function to take `diskType` as an argument, since this resource uses `managed_disk_type` instead of `storage_account_type`.

* `storage_data_disk[x]` is an array of additional disks and should add more disks similar to `storage_os_disk`.

ekifel:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
@ekifel the master branch test logs are showing this, can you please check? 
```
time="2021-05-26T17:14:57Z" level=warning msg="No products found for storage_data_disk Disk operations, using 0.00"
time="2021-05-26T17:14:57Z" level=warning msg="No products found for storage_data_disk Disk operations, using 0.00"

```

PR #714, opened 2021-05-24
feat(azure): Add support for Azure Legacy VMs

ekifel:
Issue #675

### DIFF
--- a/internal/providers/terraform/azure/testdata/virtual_machine_scale_set_test/virtual_machine_scale_set_test.tf
+++ b/internal/providers/terraform/azure/testdata/virtual_machine_scale_set_test/virtual_machine_scale_set_test.tf
@@ -1,0 +1,213 @@
+provider "azurerm" {
+  skip_provider_registration = true
+  features {}
+}
+
+resource "azurerm_resource_group" "example" {
+  name     = "example-resources"
+  location = "West Europe"
+}
+
+resource "azurerm_virtual_network" "example" {
+  name                = "acctvn"
+  address_space       = ["10.0.0.0/16"]
+  location            = azurerm_resource_group.example.location
+  resource_group_name = azurerm_resource_group.example.name
+}
+
+resource "azurerm_subnet" "example" {
+  name                 = "acctsub"
+  resource_group_name  = azurerm_resource_group.example.name
+  virtual_network_name = azurerm_virtual_network.example.name
+  address_prefixes     = ["10.0.2.0/24"]
+}
+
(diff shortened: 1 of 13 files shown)


ekifel:
@aliscott, this is ready for review!

internal/providers/terraform/azure/testdata/virtual_machine_scale_set_test/virtual_machine_scale_set_test.tf
```diff
@@ -0,0 +1,213 @@
...
+  name                = "mytestscaleset-1"
+  location            = azurerm_resource_group.example.location
+  resource_group_name = azurerm_resource_group.example.name
+
+  # automatic rolling upgrade
+  automatic_os_upgrade = true
+  upgrade_policy_mode  = "Rolling"
+
+  # required when using rolling upgrade policy
+  health_probe_id = azurerm_lb_probe.example.id
```
aliscott, line 84: Can we just pass in a fake ID here so we don't need the LB resources in the golden test?

internal/providers/terraform/azure/testdata/virtual_machine_scale_set_test/virtual_machine_scale_set_test.tf
```diff
@@ -0,0 +1,213 @@
...
```
aliscott, line 155: As above, can we pass in a fake ID?

internal/providers/terraform/azure/virtual_machine_scale_set.go
```diff
@@ -0,0 +1,84 @@
...
+	if u != nil && u.Get("instances").Type != gjson.Null {
+		capacity = decimal.NewFromInt(u.Get("instances").Int())
+	}
+
+	os := "Linux"
+	if d.Get("os_profile_windows_config").Type != gjson.Null {
+		if d.Get("storage_profile_os_disk.0.os_type").String() == "Windows" {
+			os = "Windows"
+		}
+	}
```
aliscott, line 33: Looks like the `azurerm_virtual_machine_scale_set` resource has a `storage_profile_image_reference.0.offer` which can be set to "WindowsServer" as well, so I think we should also check that, like what we do with `storage_image_reference.0.offer` for the VM resource.

aliscott (commented):
Looks great, thanks for doing the scale set resource as well @ekifel.

ekifel:
@aliscott, I added fake parameters to avoid a lot of LB resources and added check for OS type

PR #715, opened 2021-05-24
fix(aws): fix a few issues in the Amazon MQ resource

alikhajeh1:
- Terraform string values are case-insensitive for most part, changed the
Regex lookups to be case-insensitive
- Simplified the logic and fix a bug that was causing a warning before “Multiple products found for aws_mq_broker.my_aws_mq_broker_activemq_single_standby_ebs Storage (ActiveMQ, EBS), using the first product”

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -179,7 +179,7 @@ resource_usage:
     monthly_data_processed_gb: 10 # Monthly data processed by the NAT Gateway in GB.
 
   aws_mq_broker.my_aws_mq_broker:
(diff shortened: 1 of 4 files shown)


alikhajeh1:
@sinabakh just FYI on this PR, the I noticed a warning in the test logs and it took me down a rabbit hole, AWS pricing is damn complicated...

aliscott (approved):
LGTM

sinabakh:
Thanks, @alikhajeh1. I tried hard to handle all cases but as you said it's complicated.

Issue #506, opened 2021-03-10
Add support AWS WAFv2


dawidmalina:
Terraform resources:
- aws_wafv2_rule_group
- aws_wafv2_web_acl

AWS pricing page: https://aws.amazon.com/waf/pricing/

Fallowing resources are free of charge:
- aws_wafv2_ip_set
- aws_wafv2_regex_pattern_set
- aws_wafv2_web_acl_association
- aws_wafv2_web_acl_logging_configuration

Price example
| Resource Type | Price |
| --- | --- |
| Web ACL | $5.00 per month (prorated hourly) |
| Rule | $1.00 per month (prorated hourly) |
| Request | $0.60 per 1 million requests |

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
Notice this bit at the bottom of the pricing table:
In addition, you will be charged $1.00 per month (prorated hourly) for each **rule group** or each **managed rule group** that you add to your web ACL"

The main resource is `aws_wafv2_web_acl` and it has the following cost components:
- "Web ACL usage", units "hours"

- "Rules", units "hours". To calculate the monthly quantity, sum these three numbers:
a) number of `rule` Terraform params, which have the rule.action set. This means it's a rule, and not a reference to cases b).
b) number of rule_group_rules and managed_rule_group_rules in the usage file: 
```
  rule_group_rules: 5          # Total number of Rule Group rules.
  managed_rule_group_rules: 10 # Total number of Managed Rule Group rules.  
```

- "Rule groups", units "hours", count the number of `rule.rule_group_reference_statement.arn` and use that as the monthly quantity. If that count is 0, don't show this cost component.
- "Managed rule groups", units "hours", count the number of `rule.managed_rule_group_statement.name` and use that as the monthly quantity. If that count is 0, don't show this cost component.

- "Requests", units "requests", usage file param: `monthly_requests: 1000000 # Monthly number of web requests received.`

In the future, we can make a change to the CLI parser so we could count the rules from the referenced resources (after that we can remove the 2 rule params from the usage file).

alikhajeh1:
Will be released in the next version (this week).

alikhajeh1:
@dawidmalina we added support for WAF and WAFv2 in v0.9.0 (released just now). Please let us know if you have any feedback on the output or notice any issues.

It was one of the more complicated resources - thanks to @NeverovDS for sticking with it!

PR #716, opened 2021-05-24
feat(aws): Add support AWS WAFv2

NeverovDS:
Init commit
Issue #506 
aws_wafv2_web_acl:
component : Rule groups
I can't get value rule.rule_group_reference_statement.arn(always 0) - I need advice what logic to implement this component


alikhajeh1:
Looks like the TF file has an issue in this PR? Have you tried running this to see if that works ok locally?
```
 INFRACOST_LOG_LEVEL=info make -s run ARGS="breakdown --path internal/providers/terraform/azure/your_project"
```

``` Terraform command failed with:
    Error: Too many action blocks
      on  line 0:
      (source code not available)
    No more than 1 "action" blocks are allowed
    Error: Too many action blocks
      on  line 0:
      (source code not available)
    No more than 1 "action" blocks are allowed

=== CONT  TestWafV2WebAclGoldenFile
    tftest.go:197: 
        	Error Trace:	tftest.go:197
        	            				wafv2_web_acl_test.go:15
```

alikhajeh1:
>  I can't get value rule.rule_group_reference_statement.arn(always 0) - I need advice what logic to implement this component

@NeverovDS can you please paste an example terraform block, or point me to which of the tests in this PR have the issue so I can reproduce/investigate it?

internal/providers/terraform/aws/testdata/waf_web_acl_test/.terraform.lock.hcl
```diff
@@ -0,0 +1,20 @@
+# This file is maintained automatically by "terraform init".
```
alikhajeh1, line 1: this file should not be checked in (I updated the .gitignore in master earlier today)

alikhajeh1 (commented):
>  I can't get value rule.rule_group_reference_statement.arn(always 0) - I need advice what logic to implement this component

I added some temp logging code in the `internal/providers/terraform/aws/wafv2_web_acl.go` file to demonstrate how we can fetch all the rules from Terraform and loop through them to do the counting.

PR #718, opened 2021-05-25
fix(azure): ignore EP service plans as they don’t have a cost

alikhajeh1:
Their costs are handled in azurerm_function_app


alikhajeh1:
@NeverovDS can you please review this PR? I noticed a few issues with case-sensitive comparisons, and also added logic to ignore the EP types as they don't have costs in app_service_plans

PR #726, opened 2021-05-26
docs: improve testing notes

alikhajeh1:


alikhajeh1:
@tim775 / @aliscott can you plz review this? the `-update` option is very nice but we should still double check things manually, and/or in code reviews

CONTRIBUTING.md
```diff
@@ -71,7 +71,7 @@ make test
 
 #### Integration tests
 
-The entire test suite can take >20 mins to run, so we recommend against running them all locally. These will run on GitHub actions.
+The entire test suite can take ~7 mins to run, so we recommend against running them all locally. These will run on GitHub actions.
```
tim775, line 74: Do we still recommend against running the whole thing locally?  It takes 2-3 minutes on my machine.

alikhajeh1:
> "Please verify the calculations with the cloud cost calculators and include the output in the PR"

Good idea, there are two things that contributors should verify: 
- a) the price shown in the CLI matches the cloud pricing webpage for the tests they ran; using `--fields=price,...` helps - maybe that field should be in the golden files too?
- b) the cost calculation is correct. The latter can always be done manually, and for certain resources in the cloud vendor cost calcs (sometimes they don't have the service or cost component from the pricing web page).

> Should the build just fail if there are warnings?

+1, maybe we can save the test logs to a file in GH Actions and grep them for warnings as an initial cut to fail the build? We could also output the file at the end so the logs are actually visible in the GH Action.

alikhajeh1:
@ekifel re the "azurerm_mssql" test warning logs, can we do anything to avoid those? Those are legit warnings, right? so the build shouldn't fail on those?

ekifel:
@alikhajeh1, yeah, it happening because we have duplicate cost components in base. They differ only by SKU parameter, so we can't avoid these warnings without changes in db.

alikhajeh1:
Thanks for confirming @ekifel - I guess for those cases we can have an ignore-list in the GH Action steps so they don't fail the build.

alikhajeh1:
We've fixed all warnings in the master branch, @ekifel I added a warning log for the above case so it's clear for future maintainers: https://github.com/infracost/infracost/commit/90e2fa34cf91811b78b169fdc663ad20b8f48070

Going to merge this PR. We can create issues to implement the above ideas if this issue pops-up again...

PR #727, opened 2021-05-26
fix(test): fixed no/multiple products warnings in tests

ekifel:
Fixes of these cases: 

https://github.com/infracost/infracost/issues/591#issuecomment-848972038
https://github.com/infracost/infracost/issues/675#issuecomment-848976179
https://github.com/infracost/infracost/issues/669#issuecomment-848968240


ekifel:
It is initial fixes of that cases tests. 
Support of managed disk types and instance types needs in more investigation.

Issue #531, opened 2021-03-24
Add support and documentation for Azure DevOps CICD


anmoel:
Hi, 

we also use Azure DevOps for some Projects. Can you add an Azure DevOps extension to the marktplace (https://marketplace.visualstudio.com/) or document how to integrate it?

Thanks in advance
André

alikhajeh1:
Hey @anmoel - thanks for creating this issue, it's something we should support.

I've not used Azure DevOps before and happy to learn. Do you know if it's possible to run docker tasks in Azure pipelines? That's how most of the current CI/CD integrations work, e.g. see the [GitLab one](https://gitlab.com/infracost/infracost-gitlab-ci/-/blob/master/infracost.yml). If so, then it should be fairly simple to get this working.

alikhajeh1:
@anmoel we shipped initial support for Azure DevOps Pipelines, can you possibly review https://github.com/infracost/infracost-azure-devops/ and let me know if it works ok for you?

PR #728, opened 2021-05-27
Add CI/CD integration for Azure DevOps

alikhajeh1:
Fixes https://github.com/infracost/infracost/issues/531 

This works with Azure DevOps Pipelines that use either: the built-in Azure DevOps git repos or GitHub repos. If users ask for it, we can add support for the other options (e.g. bitbucket repos).

@sinabakh can you possibly review this PR?

### DIFF
--- a/internal/config/environment.go
+++ b/internal/config/environment.go
@@ -95,6 +95,8 @@ func ciScript() string {
 		return "ci-diff"
 	} else if IsTruthy(os.Getenv("INFRACOST_CI_ATLANTIS_DIFF")) {
 		return "ci-atlantis-diff"
+	} else if IsTruthy(os.Getenv("INFRACOST_CI_JENKINS_DIFF")) {
+		return "ci-jenkins-diff"
 	}
 
 	return ""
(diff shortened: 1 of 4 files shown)


scripts/ci/diff.sh
```diff
@@ -93,12 +93,12 @@ build_output_cmd () {
 format_cost () {
   cost=$1
     
-  if [ -z "$cost" ] | [ "${cost}" == "null" ]; then
```
alikhajeh1, line 96: I made these changes because https://www.shellcheck.net/ was flagging them - they were also causing warnings in the bash version that Ubuntu runs (with our Azure DevOps integration uses).

internal/config/environment.go
```diff
@@ -107,6 +109,8 @@ func ciPlatform() string {
 		return "gitlab_ci"
 	} else if IsTruthy(os.Getenv("CIRCLECI")) {
 		return "circleci"
+	} else if IsTruthy(os.Getenv("SYSTEM_COLLECTIONURI")) {
+		return fmt.Sprintf("azure_devops_%s", os.Getenv("BUILD_REPOSITORY_PROVIDER"))
 	} else {
```
sinabakh, line 114: Since you touched and fixed few Jenkins-related things. Here's an environment to detect Jenkins: `JENKINS_HOME` has false positives but I think we can ignore them.

sinabakh:
Apart from the comment, LGTM.

internal/config/environment.go
alikhajeh1: Done

PR #731, opened 2021-05-27
feat(aws): Add support AWS WAFv2

NeverovDS:
Issue #506


NeverovDS:
@alikhajeh1 ready to check

ekifel:
Hey @alikhajeh1! Will happy to review this

internal/providers/terraform/aws/waf_web_acl.go
```diff
@@ -0,0 +1,149 @@
...
+		Unit:            unit,
+		UnitMultiplier:  unitMultiplier,
+		MonthlyQuantity: quantity,
+		ProductFilter: &schema.ProductFilter{
+			VendorName:    strPtr("aws"),
+			Region:        strPtr(region),
+			Service:       strPtr("awswaf"),
+			ProductFamily: strPtr("Web Application Firewall"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "usagetype", Value: strPtr(usagetype)},
```
ekifel, line 122: Now anytime we see `Value` or `ValueRegex` with a variable, we should take care of case-sensitive lookups to follow [contribution guide](https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md#usage-file-guidelines).
```suggestion
				{Key: "usagetype", Value: strPtr(fmt.Sprintf("/%s/i", usagetype))},
```

internal/providers/terraform/aws/waf_web_acl.go
```diff
@@ -0,0 +1,149 @@
...
+	rules = d.Get("rules").Array()
+	var count int
+	for _, val := range rules {
+		types := val.Get("type").String()
+		if types == "GROUP" {
+			count++
+		}
+	}
+	ruleforGroup = count
+	if ruleforGroup > 0 {
```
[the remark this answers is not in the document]
ekifel: If we showing `Rule groups` cost component only when count > 0, it would be nice if you'll add a test to cover these case (and similar cases below)

internal/providers/terraform/aws/wafv2_web_acl.go
```diff
@@ -0,0 +1,158 @@
...
+			"months",
+			"USE1-RuleV2",
+			1,
+			&sumForRules,
+		))
+	}
+
+	if d.Get("rule.0.statement.0.rule_group_reference_statement").Type != gjson.Null {
+		counter := 0
+		log.Warnf(">>>> processing resource=%s", d.Address)
```
ekifel, line 60: I'm pretty sure that user doesn't want to see this in his logs :)
```suggestion

```

ekifel (approved):
Thanks @NeverovDS, it was complicated resource! The calculations is correct, just some minor code-changes.

internal/providers/terraform/aws/waf_web_acl.go
ekifel: To match the complete value we can use `^` and `$` symbols: ValueRegex: strPtr(fmt.Sprintf("/^%s$/i", usagetype))},

NeverovDS:
@ekifel ready to review

ekifel:
@alikhajeh1 LGTM!

PR #736, opened 2021-05-28
chore: bump Terraform and Terragrunt versions

alikhajeh1:
This increases the compressed size of our docker image from ~260MB to ~320MB. Not the end of the world but we should probably stop adding other dependencies soon?


aliscott (approved):
Yeah I think we should look at moving to separate tags for different Terraform versions soon.

PR #739, opened 2021-05-28
fix(aws): remove hours from units as we already take that into account

alikhajeh1:
CR = change-record, where we calculate it as follows:
averageStatements.Mul(decimal.NewFromInt(730)).Mul(changeRecords).Mul(windowHours)
So I think it’s clearer if we remove the hours and call it “change-records” as we’re already taking into account the hours


tim775:
@alikhajeh1 need to `-update` but otherwise LGTM

Issue #707, opened 2021-05-24
Add support for Azure Automation


alikhajeh1:
Pricing details: https://azure.microsoft.com/en-us/pricing/details/automation/
Terraform resources: azurerm_automation* (there's quite a lot of resources but most are free)

tim775:
@alikhajeh1 I had trouble wrapping my head around this, but I think the best option is:

For `azurerm_automation_account `:
- A `monthly_job_run_ mins` usage parameter displayed as `Job run time (over 500 mins)` with unit `minutes`.
- A `monthly_watcher_hours` usage parameter displayed as `Watchers (over 744 hours)` with unit `hours`.
- A `monthly_non_azure_configuration_node_count` usage parameter displayed as `Non-azure configuration nodes (over 5)` with unit `nodes`.

All other resources would be free.

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
@NeverovDS maybe go for #709 until we finalize the details on this one? I've not had a chance to look into Tim's comments

NeverovDS:
@alikhajeh1 Sure!

alikhajeh1:
@tim775 how about this?

1. `azurerm_automation_job_schedule` (since a runbook by itself doesn't seem to run):
    - Cost component "Job run time" with unit "minutes", usage file param `monthly_job_run_mins: 0 # Monthly number of job run minutes.`.
    - Cost component "Watchers" with unit "hours", usage file param `monthly_watcher_hours: 0 # Monthly number of watcher hours.`

2. `azurerm_automation_dsc_configuration` and `azurerm_automation_dsc_nodeconfiguration` each have:
    - Cost component "Non-azure config nodes" with unit "nodes", and usage file param `non_azure_config_node_count: 0 # Number of non-Azure configuration nodes.`

3. `azurerm_automation_account` would have all 3 of the above cost components and their usage file params to show users that the account itself has these prices. But by default since the usage file params are 0, it doesn't show costs.

All other resources would be free.

tim775:
That seems good except I can't say if a 'watcher' makes sense on a job schedule or not.  

I guess I wasn't sure how a "one-time" run of a runbook would be represented in tf but I think have the ability to put `monthly_job_run_mins` on `azurerm_automation_account` is sufficient to handle that.

alikhajeh1:
[...]

PR #770, opened 2021-06-02
feat(azure):Add support for Azure Automation

NeverovDS:
Issue #707


NeverovDS:
@ekifel ready to check

internal/providers/terraform/azure/testdata/automation_dsc_configuration_test/automation_dsc_configuration_test.golden
```diff
@@ -0,0 +1,20 @@
...
+ └─ Non-azure config nodes                                        0  nodes                    $0.00 
+                                                                                                    
+ PROJECT TOTAL                                                                               $30.00 
+
+----------------------------------
+To estimate usage-based resources use --usage-file, see https://infracost.io/usage-file
+
+1 resource type wasn't estimated as it's not supported yet.
+Please watch/star https://github.com/infracost/infracost as new resources are added regularly.
+1 x azurerm_automation_account
```
ekifel, line 20: Looks like it needs in update, as azurerm_automation_account should not be skipped

internal/providers/terraform/azure/testdata/automation_dsc_nodeconfiguration_test/automation_dsc_nodeconfiguration_test.tf
```diff
@@ -0,0 +1,118 @@
...
+  content_embedded        = "configuration test {}"
+}
+
+resource "azurerm_automation_dsc_nodeconfiguration" "zeroNodes" {
+  name                    = "test.localhost"
+  resource_group_name     = azurerm_resource_group.example.name
+  automation_account_name = azurerm_automation_account.example.name
+  depends_on              = [azurerm_automation_dsc_configuration.example]
+
+  content_embedded = <<mofcontent
```
ekifel, line 32: I think it would be better to use this (or "fake"), to avoid a lot of useless strings in that case
```suggestion
  content_embedded = "configuration test {}" 
```

internal/providers/terraform/azure/testdata/automation_job_schedule_test/automation_job_schedule_test.tf
```diff
@@ -0,0 +1,43 @@
...
+    vmname        = "TF-VM-01"
+  }
+}
+resource "azurerm_automation_job_schedule" "fiveMinutes" {
+  resource_group_name     = azurerm_resource_group.example.name
+  automation_account_name = "tf-automation-account"
+  schedule_name           = "hour"
+  runbook_name            = "Get-VirtualMachine"
+
+  parameters = {
```
ekifel, line 28: This is not very important, but we can avoid `optional` config parameters (like this block), on which the cost does not depend for reduce the number of lines of code

ekifel (approved):
Thanks @NeverovDS! Almost done, just some minor changes

PR #771, opened 2021-06-02
fix(azure): handle cases where references are not available

alikhajeh1:
This might happen if the reference is defined in another module.


aliscott (approved):
LGTM

Issue #625, opened 2021-05-01
bug(aws): root block device calculation - autoscaling launch template.


robh007:
Currently when using a launch template with ```aws_autoscaling_group``` storage is being calculated twice in some scenarios. Infracost by default calculates a 8gb root volume is calculated, however with a launch template you don't have a root_block_device attribute in terraform. Launch templates use a ``block_device_mapping`` which also allows you to override the root volume.

alikhajeh1:
Thanks for reporting this, can you possibly add a TF snippet that can be used to reproduce this?

robh007:
Yeah sorry this probably would have helped.

```hcl
data "aws_ami" "this" {
  owners = ["amazon"]
  most_recent = true
  filter {
    name = "description"
    values = ["Amazon Linux 2 *"]
  }
}

resource "aws_launch_template" "lt1" {
  image_id      = "fake_ami"
  instance_type = "t3.medium"

  block_device_mappings {
    device_name = data.aws_ami.this.root_device_name
    ebs {
      volume_size = 300
    }
  }
}

resource "aws_autoscaling_group" "asg1" {
  launch_template {
    id = aws_launch_template.lt1.id
  }
  desired_capacity = 1
  max_size         = 1
  min_size         = 1
}

```

Infracost reports the following.

```
  aws_autoscaling_group.asg1                                                                                           
  └─ aws_launch_template.lt1                               
     ├─ Instance usage (Linux/UNIX, on-demand, t3.medium)          730  hours       0.0456       0.0456       33.2880  
     ├─ CPU credits                                                  0  vCPU-hours  0.0500       0.0000        0.0000  
     ├─ root_block_device                                  
     │  └─ Storage (general purpose SSD, gp2)                        8  GB-months   0.1100       0.0012        0.8800  
     └─ block_device_mapping[0]                            
        └─ Storage (general purpose SSD, gp2)                      300  GB-months   0.1100       0.0452       33.0000  
  Total                                                                                          0.0920       67.1680  
                                                                                                                       
  OVERALL TOTAL (USD)                                                                            0.0920       67.1680  

```
This code assumes there's a root_block_device, which isn't available within a launch_template.
https://github.com/infracost/infracost/blob/89c9256ee2fe281d9573e1ad20c397800e24e91e/internal/providers/terraform/aws/autoscaling_group.go#L170-L172

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

[...]

PR #779, opened 2021-06-04
bug(aws): root block device calculation - autoscaling launch template.

NeverovDS:
Issue #625


NeverovDS:
@ekifel check plz

ekifel:
@alikhajeh1 LGTM, but can u take a quick look? I'm not 100% sure that we understand issue point correctly

alikhajeh1:
@robh007 if you get the chance, it would be great if you can take a quick look at this to see if it addresses the issue you created.

alikhajeh1:
I think this makes sense as aws_eks_node_group also uses launch_templates. Going to merge this but happy to change it if @robh007 says otherwise!

Issue #49, opened 2020-09-02
Add pricing support for AWS CloudFormation


hassankhosseini:
AWS Cloud Formation costs: https://aws.amazon.com/cloudformation/pricing/

`aws_cloudformation_stack` and `aws_cloudformation_stack_set` are the main resources. For each one, we can use this logic to show the cost components:
1. If `template_body` is set AND its "Resources.types" start with `AWS::*`, `Alexa::*`, or `Custom::*` then return `IsSkipped:    true, NoPrice: true,` as it's a free resource.
2. Else, show both of these:
  - "Handler operations", units "operations". Usage file param `monthly_handler_operations: 10000 # Monthly number of non-free handler operations (resources outside of the AWS::*, Alexa::*, and Custom::* namespaces).`
  - "Durations above 30s", units "seconds". Usage file param `monthly_duration_secs: 0 # Monthly duration of non-free handler operations that go above 30 seconds, in seconds.`

As normal, the cost components should show cost-depends-on-usage if the usage file param is not specified.

Free resources:
aws_cloudformation_stack_set_instance
aws_cloudformation_type

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
@nocquidant, @miyurusankalpa, @guilhermerenew, @danielbayerlein and @RichardDRJ support for costs of CloudFormation used in Terraform projects was released in v0.9.1 today.

If you're interested in seeing Infracost process CloudFormation templates as input, please 👍 [this issue](https://github.com/infracost/infracost/issues/190)

PR #780, opened 2021-06-05
feat(aws): Add pricing support for AWS CloudFormation

NeverovDS:
Issue #49

### DIFF
--- a/internal/providers/terraform/aws/cloudformation_stack.go
+++ b/internal/providers/terraform/aws/cloudformation_stack.go
@@ -17,7 +17,7 @@ func GetCloudFormationStackRegistryItem() *schema.RegistryItem {
 }
 
 func NewCloudFormationStack(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
-	if d.Get("template_body").Type != gjson.Null && checkAWS(d) || checkAlexa(d) || checkCustom(d) {
+	if d.Get("template_body").Type != gjson.Null && (checkAWS(d) || checkAlexa(d) || checkCustom(d)) {
 		return &schema.Resource{
 			NoPrice:   true,
 			IsSkipped: true,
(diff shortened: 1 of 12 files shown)


internal/providers/terraform/aws/cloudformation_stack.go
```diff
@@ -0,0 +1,80 @@
...
+}
+func checkAWS(d *schema.ResourceData) bool {
+	return strings.Contains(strings.ToLower(d.Get("template_body").String()), "aws::*")
+}
+func checkAlexa(d *schema.ResourceData) bool {
+	return strings.Contains(strings.ToLower(d.Get("template_body").String()), "alexa::*")
+}
+func checkCustom(d *schema.ResourceData) bool {
+	return strings.Contains(strings.ToLower(d.Get("template_body").String()), "custom::*")
+}
```
ekifel, line 80: > The template resource types that you have permissions to work with for this create stack action, such as AWS::EC2::Instance , AWS::EC2::* , or Custom::MyCustomInstance . Use the following syntax to describe template resource types: AWS::* (for all AWS resource), Custom::* (for all custom resources), Custom::*logical_ID* `` (for a specific custom resource), ``AWS::*service_name* ::* (for all resources of a particular AWS service), and ``AWS::service_name ::resource_logical_ID `` (for a specific AWS resource).

From this part of AWS documentation you can see that `*` means that something required in the place where `*` set. It is not a part of the final string passing to resources.type variables, so, can you remove it and update tests with these cases too?

internal/providers/terraform/aws/cloudformation_stack.go
```diff
@@ -0,0 +1,80 @@
...
+
+func GetCloudFormationStackRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "aws_cloudformation_stack",
+		RFunc: NewCloudFormationStack,
+	}
+}
+
+func NewCloudFormationStack(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	if d.Get("template_body").Type != gjson.Null && checkAWS(d) || checkAlexa(d) || checkCustom(d) {
```
ekifel, line 20: I think it would be better if you separate `and` logic by brackets.
```suggestion
	if d.Get("template_body").Type != gjson.Null && (checkAWS(d) || checkAlexa(d) || checkCustom(d)) {
```

ekifel (approved):
Thanks @NeverovDS! Just couple of minor changes

NeverovDS:
@ekifel ready

ekifel:
@alikhajeh1, LGTM

alikhajeh1:
It looks like the root `infracost-usage-example.yml` file needs to be updated too. I'll merge after that's done.

NeverovDS:
@alikhajeh1 ready

Issue #525, opened 2021-03-16
add shell-completion to infracost commands 🚀 


guilhermerenew:
## About Issue 👋🏾 
@alikhajeh1 as we talked, follows the issue to add bash-completion to infracost commands! Such a feature can be compatible with the main bash `bashrc/zsh`! Looking at some examples that may help: [completion.go by Helm](https://github.com/helm/helm/blob/7b6dcfae98527c3ff7233fc16cbeac782dd82977/cmd/helm/completion.go)

alikhajeh1:
We use https://github.com/spf13/cobra internally so we should be able to use the `Automatically generated shell autocomplete for your application (bash, zsh, fish, powershell)` feature it has to do the same thing as https://cli.github.com/manual/gh_completion and create a new command:

`infracost completion: Generate shell completion scripts`

ekifel:
@alikhajeh1, I want to work on it!

alikhajeh1:
@PatMyron and @guilhermerenew we released this in v0.9.1 today, please upgrade and run ` infracost completion --help` and add a comment if it works ok on your shells! We should have done this earlier, makes it much faster to run infracost commands 🕺

guilhermerenew:
@alikhajeh1 I performed the autocomplete test following the help steps of the `infracost completion` command and it worked perfectly! This feature will be extremely welcome by the community, good job guys! 🎉

PR #786, opened 2021-06-08
feat(ci): add support for shell-completion

ekifel:
Issue #525 

Initial support

### DIFF
--- a/cmd/infracost/completion.go
+++ b/cmd/infracost/completion.go
@@ -77,7 +77,7 @@ func completionCmd() *cobra.Command {
 	}
 
 	completionCmd.Flags().String("shell", "", "supported shell formats: bash, zsh, fish, powershell")
-	completionCmd.MarkFlagRequired("shell")
+	_ = completionCmd.MarkFlagRequired("shell")
 
 	return completionCmd
 }
--- a/cmd/infracost/completion.go
+++ b/cmd/infracost/completion.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


alikhajeh1:
Thanks!
Would it be possible to use the `--shell` flag? That enables us to have other flags in case we need it and do nicer error handling like https://github.com/cli/cli/blob/trunk/pkg/cmd/completion/completion.go#L56-L61 

Also, can we add https://github.com/spf13/cobra/blob/master/shell_completions.md#descriptions-for-completions and https://github.com/spf13/cobra/blob/master/shell_completions.md#mark-flags-as-required

ekifel:
@alikhajeh1, updated! Command's flags (with descriptions) shows by `ValidArgs: []string{"--", "-"}`, in one time flags worked without that (and don't understand why it worked), but I changed something and couldn't found how to bring it back, so used this `ValidArgs` construction. But if you think that it's weird and not obvious, I can set in `ValidArgs` every flag for every command.

And LMK please if you want me to specify something with flags

cmd/infracost/completion.go
```diff
@@ -0,0 +1,90 @@
...
+
+func completionCmd() *cobra.Command {
+	completionCmd := &cobra.Command{
+		Use:   "completion --shell [bash | zsh | fish | powershell]",
+		Short: "Generate completion script",
+		Long: `To load completions:
+	
+	Bash:
+	
+		$ source <(infracost completion bash)
```
alikhajeh1, line 18:
```suggestion
		$ source <(infracost completion --shell bash)
```

alikhajeh1 (commented):
Thanks @ekifel - I played with it on zsh and it's awesome! On bash, it doesn't show the descriptions, but it looks like that's how `gh` works too. Did you see the same issue?
Did it work ok on Powershell?

@aliscott would be great if you play with this too. Were the help instructions clear for you? Should we merge these (that were auto-generated) with the ones that `gh completion --help` includes?

cmd/infracost/completion.go
```diff
@@ -0,0 +1,90 @@
...
+		$ echo "autoload -U compinit; compinit" >> ~/.zshrc
+	
+		# To load completions for each session, execute once:
+		$ infracost completion --shell zsh > "${fpath[1]}/_infracost"
+	
+		# You will need to start a new shell for this setup to take effect.
+	
+	fish:
+	
+		$ infracost completion fish | source
```
aliscott, line 40:
```suggestion
		$ infracost completion --shell fish | source
```

cmd/infracost/output.go
```diff
@@ -32,15 +32,11 @@ func outputCmd(cfg *config.Config) *cobra.Command {
 
   Merge multiple Infracost JSON files:
 
-      infracost output --format json --path out*.json`,
+			infracost output --format json --path out*.json`,
+		ValidArgs: []string{"--", "-"},
```
aliscott, line 36: What does this do?

aliscott:
Works well for me. Should we make `--path` a required flag for `diff` and `breakdown` so it works with the autocompletion? If so, is there a way to show the current error message if it's missing since it has more context than the default "required flag(s) not set" message.

cmd/infracost/output.go
ekifel: Because without that autocompletion with flags doesn't work. Maybe exist another the way to do it, but I didn't find it

ekifel:
@aliscottб yes, I did so at first, but then I realized that if we make `--path` as a required flag, it will make impossible to use `--config-path`

Issue #784, opened 2021-06-07
azurerm_automation_dsc_nodeconfiguration_test location warnings


alikhajeh1:
`INFRACOST_LOG_LEVEL=info go test -v internal/providers/terraform/azure/automation_dsc_nodeconfiguration_test.go` shows this, which is odd as underneath, that resource calls `NewAzureRMAutomationDscConfiguration`, which does not have this issue. This might be related to how the Terraform plan JSON represents this the node resource. I've included the 2 plan JSON files in this ticket so they can be compared.

```
time="2021-06-07T22:03:01+01:00" level=warning msg="Using eastus for resource azurerm_automation_dsc_nodeconfiguration.fiveNodes as its 'location' property could not be found."
time="2021-06-07T22:03:01+01:00" level=warning msg="Using eastus for resource azurerm_automation_dsc_nodeconfiguration.withoutUsage as its 'location' property could not be found."
time="2021-06-07T22:03:01+01:00" level=warning msg="Using eastus for resource azurerm_automation_dsc_nodeconfiguration.zeroNodes as its 'location' property could not be found."
```

automation_dsc_configuration_test:
```
[...]

PR #776, opened 2021-06-03
feat(azure):Add support for Azure Automation

NeverovDS:
Issue #707

PR #789, opened 2021-06-08
fix(azure): fix problems with lookupRegion and multiple products 

NeverovDS:
Issue #776 and #784 
Update lookupRegion in Azure Automation and some other resources.
Fix multiple products in Azure Automation


NeverovDS:
@alikhajeh1 ready

Issue #600, opened 2021-04-20
Add support for Google BigQuery


alikhajeh1:
The following two resources were requested by a user, would be good to add the other bigquery resources at the same time if it's straightforward:
`google_bigquery_table`
`google_bigquery_dataset`

https://cloud.google.com/bigquery/pricing

tim775:
Terraform resource [`google_bigquery_dataset`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset) should have

- A new `monthly_queries_tb` float usage parameter displayed as `Queries (on-demand)` with unit TB.  The free tier should be ignored since it is per-customer not per-dataset.

[`google_bigquery_table `](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_table) should have
- A `monthly_active_storage_gb` float usage param displayed as `Active storage` unit `GB`.
- A `monthly_long_term_storage_gb` float usage param displayed as `Long-term storage` unit `GB`.
- A `monthly_long_term_storage_gb` float usage param displayed as `Long-term storage` unit `GB`.
- A `monthly_streaming_inserts_mb` float usage param displayed as `Streaming inserts` unit `MB`.
- A `monthly_storage_write_api_gb`float usage param displayed as `Storage Write API` unit `GB`.  Ignore the free tier.
- A `monthly_storage_read_api_tb` float usage param displayed as `Storage Write API` unit `TB`.

The other tf bigquery_* resources should be free.

ekifel:
@alikhajeh1, Hey, I want to work on it

alikhajeh1:
Released in v0.9.1

PR #793, opened 2021-06-09
feat(google): add support for Google BigQuery

ekifel:
Issue #600


internal/providers/terraform/google/testdata/bigquery_dataset_test/bigquery_dataset_test.usage.yml
```diff
@@ -0,0 +1,4 @@
+version: 0.1
+resource_usage: 
+  google_bigquery_dataset.usage:
+    monthly_queries_tb: 100
```
tim775, line 4: It would be nice to make this a decimal so we can see how it handles fractional TB.
```suggestion
    monthly_queries_tb: 100.3
```

tim775 (approved):
One minor suggestion but looks great.  @alikhajeh1 LGTM

Issue #791, opened 2021-06-08
Add support for Azure Redis Cache


bgrainger:
Pricing: https://azure.microsoft.com/en-us/pricing/details/cache/#pricing
Terraform: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/redis_cache

tim775:
Terraform resource azurerm_redis_cache should have
- An hourly charge based on the sku_name and capacity, displayed as `Cache usage (<sku_name>_<family><capacity>, <number of nodes> node)` and unit `hours` where the quantity is 730 * number of nodes.  For sku_name `Basic` `<number of nodes>` is `.  For Standard it is 2.  For Premium number of nodes defaults to 2.  If `replica count` is set number of nodes is 1 + replica count.  If `shard_count` is set it is 2 * shard_count.

It doesn't look like terraform supports Enterprise or Enterprise Flash yet, so I guess we can leave them out for now.

azurerm_redis_firewall_rule and azurerm_redis_linked_server should be free resources.

ekifel:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
@bgrainger thanks for creating this issue and https://github.com/infracost/infracost/issues/790 - these resource were added in v0.9.1 released today - plz add a comment if you notice any issues with them!

PR #795, opened 2021-06-09
feat(azure): add support for Azure Redis Cache

ekifel:
Issue #791 

I added support for usage-file as let users possibility to set it if they'll want to override our node calculations

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -390,6 +390,16 @@ resource_usage:
   #
   # Terraform GCP resources
   #
+  google_bigquery_dataset.my_dataset:
+    monthly_queries_tb: 100 # Monthly number of bytes processed (also referred to as bytes read) in TB.
+
+  google_bigquery_table.usage:
+    monthly_active_storage_gb: 1000    # Monthly number of active storage modifications in GB.
+    monthly_long_term_storage_gb: 1000 # Monthly number of long-term storage modifications in GB.
+    monthly_streaming_inserts_mb: 1000 # Monthly number of streaming data inserts in MB.
+    monthly_storage_write_api_gb: 1000 # Monthly number of storage write api in GB.
+    monthly_storage_read_api_tb: 1000  # Monthly number of storage read api in TB.
+
   google_cloudfunctions_function.my_function:
     request_duration_ms: 300               # Average duration of each request in milliseconds.
     monthly_function_invocations: 10000000 # Monthly number of function invocations.
@@ -702,6 +712,9 @@ resource_usage:
     
   azurerm_private_dns_txt_record.my_txt_record:
     monthly_queries: 1500000000  # Monthly number of DNS queries
+
+  azurerm_search_service.my_service:
+    monthly_images_extracted: 1000000 # Monthly number of extracted images
 
   azurerm_storage_account.my_account:
     data_at_rest_storage_gb: 10000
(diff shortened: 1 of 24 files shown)


tim775:
Nice and clean!  @alikhajeh1 LGTM

ekifel:
I checked cost [estimate](https://azure.com/e/9c62e922ee474186bc4dc9cf3c8a75b0), all is correct except the `Premium P1`, coz in our DB price = $0.555, and Azure pricing calculator using $0.554.

internal/providers/terraform/azure/testdata/redis_cache_test/redis_cache_test.golden
```diff
@@ -0,0 +1,16 @@
+
+ Name                                  Monthly Qty  Unit   Monthly Cost 
+                                                                        
+ azurerm_redis_cache.basic_c2                                           
+ └─ Cache usage (Basic_C4), 4 node           2,920  hours       $613.20
```
alikhajeh1, line 5: `(Basic_C4, 4 nodes)` would be more consistent with other cost component names. Also can we plz add a tiny bit of logic so it shows `1 node` and `2 nodes` etc

infracost-usage-example.yml
```diff
@@ -703,6 +703,9 @@ resource_usage:
   azurerm_private_dns_txt_record.my_txt_record:
     monthly_queries: 1500000000  # Monthly number of DNS queries
 
+  azurerm_redis_cache.my_redis_cache:
+    redis_nodes: 5 # Override the number of nodes.
```
alikhajeh1, line 707: It might be good enough to remove this as the user can only increase node count via terraform and the params that are being used (replicas_per_master and shard_count)? If our calculation is wrong, we can fix it when expert users report that?

internal/providers/terraform/azure/redis_cache.go
```diff
@@ -0,0 +1,73 @@
...
+				HourlyQuantity: decimalPtr(decimal.NewFromInt(nodes)),
+				ProductFilter: &schema.ProductFilter{
+					VendorName:    strPtr("azure"),
+					Region:        strPtr(region),
+					Service:       strPtr("Redis Cache"),
+					ProductFamily: strPtr("Databases"),
+					AttributeFilters: []*schema.AttributeFilter{
+						{Key: "productName", Value: strPtr(productName)},
+						{Key: "skuName", Value: strPtr(sku)},
+						{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/^%s Cache$/", sku))},
```
alikhajeh1, line 63: all 3 have user-input so they probably need case-insensitive matches

alikhajeh1 (commented):
Looks good indeed, I added a few minor comments

internal/providers/terraform/azure/redis_cache.go
```diff
@@ -0,0 +1,80 @@
...
+				HourlyQuantity: decimalPtr(decimal.NewFromInt(nodes)),
+				ProductFilter: &schema.ProductFilter{
+					VendorName:    strPtr("azure"),
+					Region:        strPtr(region),
+					Service:       strPtr("Redis Cache"),
+					ProductFamily: strPtr("Databases"),
+					AttributeFilters: []*schema.AttributeFilter{
+						{Key: "productName", ValueRegex: strPtr(fmt.Sprintf("/^%s$/i", productName))},
+						{Key: "skuName", ValueRegex: strPtr(fmt.Sprintf("/^%s$/i", sku))},
+						{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/^%s Cache$/", sku))},
```
alikhajeh1, line 70:
```suggestion
						{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/^%s Cache$/i", sku))},
```

PR #787, opened 2021-06-08
feat(azure): Add Support for Event hub Namespace

praveenghuge:

PR #797, opened 2021-06-09
chore: make the resource_group_name explicit

alikhajeh1:
This makes it clear where the region is being checked from and in what order.
Otherwise, it’s easy to forget to including resource_group_name in ReferenceAttributes.

@tim775 / @aliscott any thoughts on this? It came up when I was reviewing https://github.com/infracost/infracost/pull/787/files and https://github.com/infracost/infracost/pull/789/files

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -473,7 +473,24 @@ The following notes are general guidelines, please leave a comment in your pull 
 
 > **Note:** Developing Azure resources requires Azure creds. See below for details.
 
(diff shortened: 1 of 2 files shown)


aliscott:
+1 on this, I like making it explicit.

tim775:
OK that makes sense to me.  I guess it would be nice to throw an error when you try to read a reference that isn't defined in the RegistryItem but this is a step in the right direction.

alikhajeh1:
Great - thanks guys! @NeverovDS has kindly updated all of the resources already so I rebased and merged this.

Issue #469, opened 2021-02-25
Add support for Google Load Balancer


sinabakh:
Terraform resources:
// google_compute_forwarding_rule
// google_compute_global_forwarding_rule
// google_compute_target_grpc_proxy
// google_compute_target_http_proxy
// google_compute_target_https_proxy
// google_compute_target_ssl_proxy
// google_compute_target_tcp_proxy
// google_compute_region_target_http_proxy
// google_compute_region_target_https_proxy

Google pricing page:
https://cloud.google.com/vpc/network-pricing#lb

tim775:
For `google_compute_forwarding_rule` and `google_compute_global_forwarding_rule`
- `Forwarding rule` with unit `hours`.  Since we can't really determine how many rules are in a project, this price should ignore the "Minimum Service Charge" for the first 5 rules and use the "Forwarding Rule Additional Service Charge..." price.  Be sure to add a note to the registry items:`Notes: []string{"Price for additional forwarding rule is used"}`.
- A `monthly_ingress_data_gb`usage param display as `Ingress data` unit `GB`

For the `_proxy` resources mentioned above:
- A `monthly_proxy_instances` float usage param should specify the average number of proxy instances.  The value should be used to calculate number of hours and be displayed as `Proxy instance` with unit `hours`
- A `monthly_data_processed_gb` usage param displayed as `Data processed` unit `GB`

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
Released in v0.9.1

PR #800, opened 2021-06-09
feat(google): Add support for Google Load Balancer

NeverovDS:
Issue #469


NeverovDS:
@ekifel ready for review

internal/providers/terraform/google/compute_forwarding_rule.go
```diff
@@ -0,0 +1,84 @@
...
+		CostComponents: costComponents,
+	}
+}
+
+func computeForwardingCostComponent(region string) *schema.CostComponent {
+	return &schema.CostComponent{
+		Name:            "Forwarding rule",
+		Unit:            "hours",
+		UnitMultiplier:  1,
+		MonthlyQuantity: decimalPtr(decimal.NewFromInt(1)),
```
ekifel, line 49:
```suggestion
		HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
```

internal/providers/terraform/google/compute_forwarding_rule.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+}
+
+func GetComputeGlobalForwardingRuleRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "google_compute_global_forwarding_rule",
+		RFunc: NewComputeForwarding,
+		Notes: []string{"Price for additional forwarding rule is used"},
+	}
+}
```
ekifel, line 23: I think It would be better if you separate it (and _proxy resources too) to different files, cos it is more consistent with our codebase and 1 resource = 1 file paradigm

ekifel (approved):
Thanks @NeverovDS! Need to do a little refactoring

internal/providers/terraform/google/compute_forwarding_rule.go
alikhajeh1: I read [a comment](https://news.ycombinator.com/item?id=27468837) on HN today that made me think more about this: "To be a good X you must follow the rules. To be a great X you must have followed the rules so well as to learn why they’re there and when they should be broken." 

@ekifel, FYI @NeverovDS pinged me and asked if it seems pointless to create a lot of extra files so in this case maybe we relax our consistency rule :)

alikhajeh1:
Looks like I merged too soon as there was another comment too - sorry, my bad. @NeverovDS can you plz send another PR?

Issue #730, opened 2021-05-27
Improvements for usage-based resources (feedback wanted!)


alikhajeh1:
I'd like to get feedback from the community on what they'd like from the `--usage-file` and `--sync-usage-file` features next. See the docs for them [here](https://www.infracost.io/docs/usage_based_resources#1-generate-usage-file):
```
--usage-file string  Path to Infracost usage file that specifies values for usage-based resources
--sync-usage-file    Sync usage-file with missing resources, needs usage-file too (experimental)
```

Some ideas for improvements are:
1. Output the comments from the [`infracost-usage-example.yml`](https://github.com/infracost/infracost/blob/master/infracost-usage-example.yml) into the generated files as without these it's sometimes not clear what the options are (e.g. for pre-defined string keys). Only lines being changed by the sync feature should get the comment so if a user has updated a line with their own comment, they won’t see the comment added to it. I think @sinabakh might have started investigating this one.

2. Validate the usage file keys. Currently typos in the YAML syntax are caught by the YAML parser but typos in the keys are ignored so it's easy to miss issues. It would be better if the CLI output an error when it sees a key that doesn't appear in the `infracost-usage-example.yml` file.

3. Fetch usage file values from sources such as AWS CloudWatch (and their GCP/Azure equivalents) or Datadog. For example, would it help if the usage file keys for `aws_s3_bucket`, `aws_lambda_function` and `aws_dynamodb_table` in the `infracost-usage-example.yml` file are populated based on last month's values? If so, what would that help with?

The above ideas might benefit from us moving to a more schema-like usage file internally vs the current `infracost-usage-example.yml`.

sinabakh:
I personally think one of the challenges of moving to more advanced features is the current schema-less structure.
It produces some challenges like parsing issues and limited flexibility. For instance, implementing number 1 is not only hard and limited but hacky in nature.

I believe in the first place, we can create an easy to define schema for usage based data.

aliscott:
For 1. Looks like the [yaml.v3](https://pkg.go.dev/gopkg.in/yaml.v3) introduced some support for parsing and writing comments, so it might not be hard to add if we use that and parse the current file into `yaml.Node` types. If this is quick, I think we should do it.

Overall though I agree that a schema-based approach is better in the long term.

PatMyron:
Could probably quickly start with a schema automatically generated from that example if it's currently the source of truth, especially since schemas can be a bit more verbose to maintain
[...]

Issue #190, opened 2020-10-21
Bare CloudFormation templates support


delissonjunio:
Hi, is there any chance we could work to support bare CloudFormation templates? At my company we don't really use TF but I'd be interested in contributing & adding support to CF if it fits within the project's scope

joetheone:
Yes please! 🙌

PatMyron:
CloudFormation has an API with limited coverage:
https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/217
Definitely reach out if anyone's planning on picking this up. Can at least point towards CloudFormation template recognition logic and unresolved CloudFormation template syntax that'll likely need to be ignored, as well as some other thoughts

PatMyron:
Easiest starting point would be automatically translating free resource types:
https://github.com/infracost/infracost/blob/8e854e6af51bfd0eda3d65dfdb85a59ce6578cd4/internal/providers/terraform/aws/registry.go#L102
Stripping delimiters (`_` or `::`) and comparing case insensitively should automatically map ~400 out of the ~900 of the AWS resource types:
https://github.com/PatMyron/terraformation/blob/main/map/map.py
https://github.com/PatMyron/terraformation/blob/create-pull-request/patch/map/resources.txt

tim775:
Note:  When we add CloudFormation IaC support, we should generate estimates for the `template_body` field of `aws_cloudformation_stack`.

DavidHaasz:
Do we have an update on the status of this. My team currently has a mix of Terraform and Cloudformation IaC and we are looking to do something around comparing estimated cost of infrastructure vs real cost of infrastructure and we would like to leverage Infracost to do that but the current lack of cloudformation support is a blocker

hugorut:
@DavidHaasz we're still focused on making the experience for Terraform the best it can possibly be, e.g. [guardrails](https://www.infracost.io/docs/infracost_cloud/guardrails/), [tagging policies](https://www.infracost.io/docs/infracost_cloud/tagging_policies/) and [finops policies](https://www.infracost.io/docs/infracost_cloud/finops_policies/). Once we're happy with the features for Terraform, we'd like to go broad and provide the same capabilities to other IaC tools. Sorry, we don't have an ETA yet for Cloudformation support.

jswietek:
[...]

PR #803, opened 2021-06-10
feat(aws): Initial sketch of cloudformation support

tim775:


tim775:
I was looking into how to support cloudformation.  The basic idea is to split resource creation into `extract resource info from provider config` and `create Resource/CostComponents from resource info`.  Extracting the info would be provider specific and remain in the `/provider/<cloud>`, but it would call a shared resource helper to generate the the actual Resource/CostComponents.  I refactored dynamodb in this resource as an example.

I think the biggest thing to figure out for multi-provider support is the usage .yml file.  The current support is essentially terraform specific because the usage keys map directly to a terraform resource block.  

One option for multi-provider support would be to keep the current format but create a mapping from the current usage keys to CF resources.   This may not be straightforward because there is not a 1-1 correlation between CF and TF resources.  E.g. CF has separate `AWS::DynamoDB::Table` and `AWS::DynamoDB::GlobalTable` resources, while TF handles both regular and global tables with the `aws_dynamodb_table` resource.

Another option that might feel more natural for CF users would be to let them specify usage in a similar structure as the CF template.  So given a template resource:
```
{
    "AWSTemplateFormatVersion": "2010-09-09",
    "Resources": {
        "AlertsTable": {
            "Type": "AWS::DynamoDB::Table",
            "Properties": {
            ...
```
The usage parameters could be specified as
```
{
    "InfracostUsageVersion": "0.1",
    "Resources": {
        "AlertsTable": {
            "Type": "AWS::DynamoDB::Table",
            "Usage": {
                 "monthly_write_request_units": 3000000 
                 "monthly_read_request_units": 8000000 
                ...
```
If we want to go this route, it probably means adding more structure to the usage parameters so the files can be parsed and generated easily in whatever format is needed.

tim775:
There are additional cloudformation specific issues that need to be figured out:
- How should users specify region?  This information isn't in the template file, users specify which region to run in when executing a template.
- Adding support for stacks. The current implementation simply parses an existing template file.  The next step would be to support downloading and processing the template for an existing CF stack using the aws cli.
- Adding support for changesets/diffs.  Templates don't contain any information about the current state and therefore a single template can't be used to generate a diff.  One option for generating diffs would be to use the AWS CLI to download a template for the existing stack state and compare it to a modified local template.  Another option is to use the AWS CLI to create a new ChangeSet based on the modified template, then download the resulting "plan" information and generate the diff from that.  
- The `goformation` library I'm using only has minimal support for the template lanugae "Ref" function, so we may need to build a more complement implementation.
- We should refactor the tftest goldenfile testing stuff so it can be used with CF (and other future providers).
- How should we extract Tags?  I couldn't figure out a nice generic way to extract tags (if they exist) from any CF resource using the `goformation` template parsing library, so I set them with the resource-specific registry function.  It seems like it would be easy to mistakenly leave this out when creating resources, so it would be good to look at a better way of doing this.

aliscott:
> but it would call a shared resource helper to generate the the actual Resource/CostComponents.

This is cool. I can see how in the future `DynamoDbTableArguments` could be sent to an external service/API along with the usage data, which could then return the resource and cost components. This would mean other clients, not just Infracost, could use that logic as well.

> I think the biggest thing to figure out for multi-provider support is the usage .yml file.

One thing we had in the past was a Terraform provider so users could specify the usage data in HCL code, so it was similar to the TF they were writing. We found that nobody used it, but not sure if that was due to it being HCL or the fact it had to be bundled in their existing TF project. Totally making an assumption here, but I'm wondering if more people write their CF in YAML if they're handwriting it, and maybe JSON is used more if they're generated it.

Would this be how you'd see option 1 looking?

```yml
version: 0.1
resource_usage:

  # TF resource
  aws_dynamodb_table.my_table:
    monthly_write_request_units: 3000000
    monthly_read_request_units: 8000000
  
  # CF resource (AWS::DynamoDB::Table)
  AlertsTable:
    monthly_write_request_units: 3000000
    monthly_read_request_units: 8000000
    
  # CF resource (AWS::DynamoDB::GlobalTable)
  GlobalAlertsTable:
    monthly_write_request_units: 3000000
    monthly_read_request_units: 8000000
```

tim775:
> I can see how in the future DynamoDbTableArguments could be sent to an external service/API 
Right the `/resources` logic could ultimately be an external service that generates and prices the cost components.  I was wondering if there is a use case for an API that processes TF/CF config files (or snippets).

> Would this be how you'd see option 1 looking?

So NO that wasn't what I was thinking, but I like that better 🙂.  I guess that really simplifies things because we're still just using the natural 'address' which for TF is <type>.<name> but for CF is just <name>.  That gets my vote.

PatMyron:
https://github.com/infracost/infracost/issues/190

Not sure if `infracost` is handling any Terraform equivalents already, but this should be probably ignore some unresolved CloudFormation template syntax where values aren't resolved until runtime like [dynamic references](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html) and [intrinsic functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)

internal/resources/aws/dynamodb_table.go
```diff
@@ -0,0 +1,336 @@
...
+type DynamoDbTableArguments struct {
+	Address        string   `json:"address,omitempty"`
+	Region         string   `json:"region,omitempty"`
+	BillingMode    string   `json:"billingMode,omitempty"`
+	WriteCapacity  int64    `json:"writeCapacity,omitempty"`
+	ReadCapacity   int64    `json:"readCapacity,omitempty"`
+	ReplicaRegions []string `json:"replicaRegions,omitempty"`
+}
+
+func NewDynamoDBTable(args *DynamoDbTableArguments, u *schema.UsageData) *schema.Resource {
```
sinabakh, line 19: @tim775 Just leaving my comment here too. I believe the `Resource Abstraction` should not care about what fields come from what file. So all it takes is a bunch of mandatory and optional fields. The caller of this function then can decide how to construct it.

If I want to make an example is let's say we have the resource `X` with fields of `A`, `B` and `C`. In terraform, fields `A` and `B` are present in the state file and field `C` comes from usage data. But, there's no guarantee that all providers are like terraform. Another provider like Pulumi might have support for field `C` in its own config and state and here's where things get super complicated if we take two different parameters.

internal/resources/aws/dynamodb_table.go
```diff
@@ -0,0 +1,366 @@
...
+	ReadCapacity   int64    `json:"readCapacity,omitempty"`
+	ReplicaRegions []string `json:"replicaRegions,omitempty"`
+
+	MonthlyWriteRequestUnits       *int64 `json:"monthlyWriteRequestUnits,omitempty"`
+	MonthlyReadRequestUnits        *int64 `json:"monthlyReadRequestUnits,omitempty"`
+	StorageGB                      *int64 `json:"storageGB,omitempty"`
+	PitrBackupStorageGB            *int64 `json:"pitrBackupStorageGB,omitempty"`
+	OnDemandBackupStorageGB        *int64 `json:"onDemandBackupStorageGB,omitempty"`
+	MonthlyDataRestoredGB          *int64 `json:"monthlyDataRestoredGB,omitempty"`
+	MonthlyStreamsReadRequestUnits *int64 `json:"monthlyStreamsReadRequestUnits,omitempty"`
```
alikhajeh1, line 23: Could we add the fields that @sinabakh mentioned in [his usage-file schema comment](https://github.com/infracost/infracost/issues/730#issuecomment-856993301) here so we have 1 source of truth for them? That source could then be used in the future by `--sync-usage-file` and docs.

tim775: Yes, the usage schema for the resource is defined separately from the args a few lines below:
https://github.com/infracost/infracost/pull/803/files/ded0f66110486cb9aeffb5992fe7223aec699ba5#diff-71ed323ee96294a7b798471f490d0ecde86fa1b63763f1e023d4bf93e40dc21eR38-R46
and it's added to the resource created by NewDynamoDBTable.  (This is a little bit different than adding it to the RegistryItem which is what @sinabakh had proposed).

I made changes to sync-usage-file so it will use this resource.UsageSchema if present, or fall back to the "old way" of using the usage-example.

internal/schema/usage_data.go
```diff
@@ -30,6 +50,15 @@ func (u *UsageData) Get(key string) gjson.Result {
 	return u.Attributes[key]
 }
 
+func (u *UsageData) GetInt(key string) *int64 {
```
aliscott, line 53: I like this, a lot cleaner than checking for gjson.Null everywhere.

aliscott (approved):
This looks great. I like how smoothly it works with the usage data.

I wonder if we should split out the CF-specific stuff for just now into a separate integration branch until we have a bit more to keep things cleaner and the bin size leaner?

tim775:
I extracted the non-cloudformation stuff to https://github.com/infracost/infracost/pull/824

aliscott:
Going with the philosophy of putting stuff out there to get feedback, I think we should merge this since it doesn't affect anything in the core Terraform functionality. @tim775 can you rebase this to pick up the Float64 changes and I will merge?

tim775:
@aliscott I rebased and this is ready.  I added the goformation package.  Is it typical to see all the `go.sum` changes?

aliscott:
> Is it typical to see all the go.sum changes?

@tim775 not usually this many. Not sure why it's added these new lines but when I run `go mod tidy` on this branch it removes a lot of them.

Issue #583, opened 2021-04-15
Add support for AWS Neptune


caphrim007:
Pricing information is found here

https://aws.amazon.com/neptune/pricing/

I think pricing is specific to the instances. Neptune organizes instances in clusters, but afaik there is not pricing that is specific to a cluster. Terraform has modules for instance and cluster.

tim775:
For `neptune_cluster_instance`:
- `Database instance (on-demand, <instance type>)` with unit `hours`.  Iinstance type is determined from the tf `instance_class`.  The `count` field, if present, should be used to multiply the number of monthly hours.
- For t3 instances, a `monthly_cpu_credit_hrs` usage parameter should be displayed as `CPU credits` with unit `vCPU-hours`.

For `aws_neptune_cluster`:
- A `storage_gb` usage parameter displayed as `Storage` with unit `GB`
- An `monthly_io_request` usage parameter displayed as `I/O` with unit `1M request`
- A `backup_storage_gb` usage param displayed as `Backup storage` with unit `GB`.  A similar parameter is used for documentDB and decides to display the price if `backup_retention_period` is present and > 1.  We should be consistent with that implementation.  This captures the footnote from the pricing page: `There is no additional charge for backup storage of up to 100% of your total Neptune database storage for each Neptune database cluster. There is also no additional charge for backup storage if your backup retention period is 1 day and you don’t have any snapshots beyond the retention period.`

For `aws_neptune_cluster_snapshot`:
- A `backup_storage_gb` usage param displayed as `Backup storage` with unit `GB`.

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
@caphrim007 just FYI this was released in v0.9.2 today, plz let us know if it works ok for you.

PR #813, opened 2021-06-14
feat(aws): Add support for AWS Neptune

NeverovDS:
Issue # #583


NeverovDS:
@ekifel ready for review

internal/providers/terraform/aws/neptune_cluster.go
```diff
@@ -0,0 +1,94 @@
...
+	var retentionPeriod *decimal.Decimal
+	if d.Get("backup_retention_period").Exists() || d.References("db_cluster_identifier") != nil {
+		retentionPeriod = decimalPtr(decimal.NewFromInt(d.Get("backup_retention_period").Int()))
+		if retentionPeriod.GreaterThan(decimal.NewFromInt(1)) {
+			var backupStorage *decimal.Decimal
+			if u != nil && u.Get("backup_storage_gb").Type != gjson.Null {
+				backupStorage = decimalPtr(decimal.NewFromInt(u.Get("backup_storage_gb").Int()))
+			}
+			costComponents = append(costComponents, neptuneClusterBackupCostComponent(region, "BackupUsage", backupStorage))
+		}
```
ekifel, line 49: I think it would be better if you set this cost component as free resource if `retentionPeriod` will be lower than 1
```suggestion
		}
	else {
		return &schema.Resource{
 		NoPrice:   true,
 		IsSkipped: true,
 	}
```

internal/providers/terraform/aws/testdata/neptune_cluster_instance_test/neptune_cluster_instance_test.golden
```diff
@@ -0,0 +1,22 @@
...
+ └─ Database instance (on-demand, db.r4.large)                 730  hours                      $297.84 
+                                                                                                       
+ aws_neptune_cluster_instance.dbR4LargeCount2[1]                                                       
+ └─ Database instance (on-demand, db.r4.large)                 730  hours                      $297.84 
+                                                                                                       
+ aws_neptune_cluster_instance.dbT3Medium                                                               
+ └─ CPU credits                                              7,300  vCPU-hours                 $817.60 
+                                                                                                       
+ aws_neptune_cluster_instance.dbT3WithoutUsage                                                         
+ └─ CPU credits                                   Monthly cost depends on usage: $0.11 per vCPU-hours
```
ekifel, line 18: In AWS Neptune's pricing page I found this: 
>  CPU Credits are charged at $0.15 per vCPU-Hour. The CPU Credit pricing is the same for T3 instance across all regions. 

What means that this cost component must be $0.15 per CPU Credit in all regions, but in our DB we have different prices for it. @alikhajeh1 / @tim775, what do you think about it? Should we be consistent with pricing page or it's ok?

ekifel (approved):
Thanks @NeverovDS!

internal/providers/terraform/aws/neptune_cluster_instance.go
```diff
@@ -0,0 +1,86 @@
...
+		monthlyCPUCreditHrs = decimalPtr(decimal.NewFromInt(u.Get("monthly_cpu_credit_hrs").Int()))
+	}
+
+	costComponents := make([]*schema.CostComponent, 0)
+
+	if instanceClass == "db.t3.medium" {
+		costComponents = append(costComponents, neptuneClusterCPUInstanceCostComponent(region, instanceClass, monthlyCPUCreditHrs))
+	} else {
+		costComponents = append(costComponents, neptuneClusterDbInstanceCostComponent(instanceClass, region, instanceClass, hourlyQuantity))
+	}
```
alikhajeh1, line 38: This logic is slightly wrong, for t3, we want to show the instance hourly cost AND the cpu credits (using the logic `(strings.HasPrefix(instanceType, "db.t3.")`)

`db.products.distinct("attributes.usagetype", {"vendorName": "aws", "service":"AmazonNeptune"});` shows that there are two usagetypes with the word "cpu" or "credit":
1. SAE1-CPUCredits:db.t3
2. APE1-CPUCredits:db.t3

This is odd, but then AWS' pricing is sometimes odd :) So I think we can use the usage type regex `CPUCredits:db.t3` to find the price, as they both have the same price (USD 0.15 per vCPU-Hour).

NeverovDS: @alikhajeh1 added changes , check plz

internal/providers/terraform/aws/neptune_cluster_instance.go
```diff
@@ -0,0 +1,85 @@
...
+		monthlyCPUCreditHrs = decimalPtr(decimal.NewFromInt(u.Get("monthly_cpu_credit_hrs").Int()))
+	}
+
+	costComponents := make([]*schema.CostComponent, 0)
+
+	if strings.HasPrefix(instanceClass, "db.t3.") {
+		costComponents = append(costComponents, neptuneClusterCPUInstanceCostComponent(monthlyCPUCreditHrs))
+	} else {
+		costComponents = append(costComponents, neptuneClusterDbInstanceCostComponent(instanceClass, region, instanceClass, hourlyQuantity))
+	}
```
alikhajeh1, line 38:
```suggestion
costComponents = append(costComponents, neptuneClusterDbInstanceCostComponent(instanceClass, region, instanceClass, hourlyQuantity))

	if strings.HasPrefix(instanceClass, "db.t3.") {
		costComponents = append(costComponents, neptuneClusterCPUInstanceCostComponent(monthlyCPUCreditHrs))
	}
```

internal/providers/terraform/aws/neptune_cluster_instance.go
```diff
@@ -0,0 +1,85 @@
...
+		},
+		PriceFilter: &schema.PriceFilter{
+			PurchaseOption: strPtr("on_demand"),
+		},
+	}
+}
+
+func neptuneClusterCPUInstanceCostComponent(quantity *decimal.Decimal) *schema.CostComponent {
+	return &schema.CostComponent{
+
```
alikhajeh1, line 69:
```suggestion
// AWS mentions that CPU Credit pricing is the same for T3 instance across all regions, but they only return prices for Hong Kong and Sao Paulo so we hard-code APE1.
```

internal/providers/terraform/aws/registry.go
```diff
@@ -71,6 +71,9 @@ var ResourceRegistry []*schema.RegistryItem = []*schema.RegistryItem{
 	GetSNSTopicRegistryItem(),
 	GetSNSTopicSubscriptionRegistryItem(),
 	GetSQSQueueRegistryItem(),
+	GetNeptuneClusterRegistryItem(),
+	GetNeptuneClusterInstanceRegistryItem(),
+	GetNeptuneClusterSnapshotRegistryItem(),
 	GetNewEKSNodeGroupItem(),
```
alikhajeh1, line 77: plz remember to search the Terraform resource list and anything that's not included here is probably free for this service

Screen Shot 2021-06-16 at 9 22 51 AM

NeverovDS:
@alikhajeh1 ready

alikhajeh1:
Thanks @NeverovDS! infracost-usage-example.yml needs to be updated with the new resource (that file is alphabetically sorted). Plz also update the aws_docdb_cluster_instance.monthly_io_request too

internal/providers/terraform/aws/neptune_cluster.go
```diff
@@ -0,0 +1,94 @@
...
+		Name:            "Backup storage",
+		Unit:            "GB",
+		UnitMultiplier:  1,
+		MonthlyQuantity: quantity,
+		ProductFilter: &schema.ProductFilter{
+			VendorName: strPtr("aws"),
+			Region:     strPtr(region),
+			Service:    strPtr("AmazonNeptune"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "usagetype", ValueRegex: strPtr(fmt.Sprintf("/%s$/i", usageType))},
```
alikhajeh1, line 87:
```suggestion
				{Key: "usagetype", ValueRegex: strPtr(fmt.Sprintf("/BackupUsage$/i", usageType))},
```
Since the method is used in one place, it might be simpler to hard-code this and remove `usageType` from the method signature?

internal/providers/terraform/aws/neptune_cluster_instance.go
```diff
@@ -0,0 +1,85 @@
...
+	if u != nil && u.Get("monthly_cpu_credit_hrs").Type != gjson.Null {
+		monthlyCPUCreditHrs = decimalPtr(decimal.NewFromInt(u.Get("monthly_cpu_credit_hrs").Int()))
+	}
+
+	costComponents := make([]*schema.CostComponent, 0)
+
+	costComponents = append(costComponents, neptuneClusterDbInstanceCostComponent(instanceClass, region, instanceClass, hourlyQuantity))
+
+	if strings.HasPrefix(instanceClass, "db.t3.") {
```
alikhajeh1, line 36:
```suggestion
	if strings.HasPrefix(strings.ToLower(instanceClass), "db.t3.") {
```

internal/providers/terraform/aws/neptune_cluster_snapshot.go
```diff
@@ -0,0 +1,40 @@
...
+		ReferenceAttributes: []string{
+			"db_cluster_identifier",
+		},
+	}
+}
+
+func NewNeptuneClusterSnapshot(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	var resourceData *schema.ResourceData
+	dbClusterIdentifier := d.References("db_cluster_identifier")
+	resourceData = dbClusterIdentifier[0]
```
alikhajeh1, line 22: it's more robust to check `if len(dbClusterIdentifier) > 0` before this line, since the reference might be in another module or project, which means we can't access it. If `db_cluster_identifier` isn't available then we should show cost-depends-on-usage - plz also add a test case for this (search the code for `account-in-another-module` to see an example of a test case that checks for this, and notice how the `account_name` is a string that doesn't exist in that TF file).

alikhajeh1 (commented):
LGTM! I added a few minor suggestions. @ekifel I didn't check the cost calculations as part of my review, has that already been done, or does @NeverovDS need to add something like [this](https://github.com/infracost/infracost/pull/787#issuecomment-857658100) to verify them?

alikhajeh1:
Thanks @NeverovDS!

PR #822, opened 2021-06-18
Flags support in infracost register cmd

shantanu-parate:
Following flags support in infracost register command

- --name
- --email

`infracost register --name <user_name> --email <user_email>`


alikhajeh1:
As discussed with @shantanu-parate, their use-case is better handled by passing in `-e "INFRACOST_API_KEY=xxxxx"` to the `docker run` command or by hard-coding the `ENV INFRACOST_API_KEY=xxxx` in the Dockerfile.

PR #824, opened 2021-06-18
Refactor/share resources between providers

tim775:


tim775:
@alikhajeh1 This should be ready to merge after a review.  I updated the contributing.md and refactored nat_gateway and lambda_function to use the shared resource pattern since those were refernced in the doc.

alikhajeh1:
LGTM

PR #833, opened 2021-06-25
fix: don't warn about usage file if not syncing

### LINKED ISSUE #832: Don't warn about missing usage file unless usage syncing enabled
If you run the `infracost breakdown` command and set the log-level to warn, a warning is output about the usage file not being specified, even if the `--sync-usage-file` flag hasn't been specified.

Here's an example:

```shell
$ infracost breakdown --path plan.json --format=json --log-level=warn > infracost.json 
time="2021-06-25T16:38:15+01:00" level=warning msg="Can't sync usage file as it's not specified"
```

I'd only expect this if the `--sync-usage-file` flag was specified, for example:

```shell
infracost breakdown --path spacelift.json --format=json --log-level=warn --sync-usage-file
```

This is with Infracost 0.9.1.

adamconnelly:
Don't warn about the usage file not being specified unless the `--sync-usage-file` flag has also been specified.

Fixes #832

One thing I'd like to check is that the output when the `--sync-usage-file` flag is specified is a little verbose:

```shell
$ infracost breakdown --path plan.json --format=json --log-level=warn --sync-usage-file
Warning: Ignoring sync-usage-file as no usage-file is specified.

time="2021-06-25T16:33:25+01:00" level=warning msg="Can't sync usage file as it's not specified"
```

It's because there's an earlier check here: https://github.com/infracost/infracost/blob/87a41b5a32e85b065eb64d3d0564c39105cc2b6b/cmd/infracost/run.go#L279

Maybe the warning could just be removed completely? I just didn't want to do that without checking.

### DIFF
--- a/internal/usage/usage_file.go
+++ b/internal/usage/usage_file.go
@@ -187,18 +187,14 @@ func loadReferenceFile() (map[string]*schema.UsageData, error) {
 	return usageData, nil
 }
 
-func LoadFromFile(usageFilePath string, syncUsageFile bool) (map[string]*schema.UsageData, error) {
+func LoadFromFile(usageFilePath string, createIfNotExisting bool) (map[string]*schema.UsageData, error) {
 	usageData := make(map[string]*schema.UsageData)
 
 	if usageFilePath == "" {
-		if syncUsageFile {
-			log.Warn("Can't sync usage file as it's not specified")
-		}
-
 		return usageData, nil
 	}
 
-	if syncUsageFile {
+	if createIfNotExisting {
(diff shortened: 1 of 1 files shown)


aliscott:
Thanks @adamconnelly!

> Maybe the warning could just be removed completely

Agree, since we already print a warning, I don't think it adds anything to have it here as well.

adamconnelly:
@aliscott I've updated the PR to just remove the warning.

aliscott (approved):
LGTM!

Issue #710, opened 2021-05-24
Add support for Azure Kubernetes Load Balancer and HTTP Application Routing


alikhajeh1:
The [azurerm_kubernetes_cluster](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster), which Infracost supports, has a `load_balancer_sku` param. Once we support the [Azure LB resource](https://github.com/infracost/infracost/issues/666) is merged, we can update the AKS resource to include costs of the LB if that param is used.

aliscott:
I've added some details for the load balancer and DNS pricing for AKS here: https://github.com/infracost/infracost/issues/591#issuecomment-829163605.

The DNS pricing applies if `http_application_routing.enabled == true`. [This page](https://docs.microsoft.com/en-us/azure/aks/http-application-routing) contains a bit more detail, but basically when this option creates a new DNS Zone, so any pricing we have for DNS Zones should apply here.

NeverovDS:
@alikhajeh1, Hey, I want to work on it!

alikhajeh1:
Released in v0.9.2 today.

PR #838, opened 2021-06-29
feat(azure):Add support for Azure Kubernetes Load Balancer and HTTP Application Routing 

NeverovDS:
Issue #710

### DIFF
--- a/internal/providers/terraform/azure/kubernetes_cluster.go
+++ b/internal/providers/terraform/azure/kubernetes_cluster.go
@@ -1,6 +1,8 @@
 package azure
 
 import (
+	"strings"
+
 	"github.com/infracost/infracost/internal/schema"
 	"github.com/shopspring/decimal"
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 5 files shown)


NeverovDS:
@ekifel ready for review

internal/providers/terraform/azure/kubernetes_cluster.go
```diff
@@ -57,6 +58,46 @@ func NewAzureRMKubernetesCluster(d *schema.ResourceData, u *schema.UsageData) *s
...
 
+	if d.Get("network_profile.0.load_balancer_sku").Type != gjson.Null {
+		if strings.ToLower(d.Get("network_profile.0.load_balancer_sku").String()) == "standard" {
+			if strings.Contains(strings.ToLower(region), "usgov") {
+				region = "US Gov"
+			} else if strings.Contains(strings.ToLower(region), "china") {
+				region = "Сhina"
+			} else {
+				region = "Global"
+			}
```
ekifel, line 69: I think it would be nice if you could move that logic to [this](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/azure/util.go) file, as it using for several Azure resources

ekifel (approved):
Thanks @NeverovDS!

internal/providers/terraform/azure/kubernetes_cluster.go
alikhajeh1: I'll do that in a separate PR after this is merged. Is everything else ok with this PR? costs match?

ekifel: LGTM. I checked the costs, they are correct

PR #839, opened 2021-06-29
feat: sketch out adding UsageSchemaItem for all resource calculation …

tim775:
…args

### DIFF
--- a/internal/providers/terraform/aws/dynamodb_table.go
+++ b/internal/providers/terraform/aws/dynamodb_table.go
@@ -17,20 +17,6 @@ func GetDynamoDBTableRegistryItem() *schema.RegistryItem {
 }
 
 func NewDynamoDBTable(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
-	region := d.Get("region").String()
-
-	billingMode := d.Get("billing_mode").String()
-
-	var readCapacity int64
-	if d.Get("read_capacity").Exists() {
-		readCapacity = d.Get("read_capacity").Int()
-	}
-
-	var writeCapacity int64
-	if d.Get("write_capacity").Exists() {
-		writeCapacity = d.Get("write_capacity").Int()
-	}
-
 	replicaRegions := []string{}
 	if d.Get("replica").Exists() {
 		for _, data := range d.Get("replica").Array() {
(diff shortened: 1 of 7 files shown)


internal/providers/terraform/aws/dynamodb_table.go
```diff
@@ -40,13 +26,13 @@ func NewDynamoDBTable(d *schema.ResourceData, u *schema.UsageData) *schema.Resou
...
-		WriteCapacity:  writeCapacity,
-		ReadCapacity:   readCapacity,
+		Region:         d.Get("region").String(),
+		BillingMode:    d.Get("billing_mode").String(),
+		WriteCapacity:  intPtr(d.Get("write_capacity").Int()),
+		ReadCapacity:   intPtr(d.Get("read_capacity").Int()),
 		ReplicaRegions: replicaRegions,
 	}
-	args.PopulateUsage(u)
+	keysToSkipSync := []string{"region", "billing_mode", "write_capacity", "read_capacity", "replica_regions"}
```
tim775, line 35: I want to let terraform say "Don't generate the example usage input for these keys since I can pull the values from HCL".  But this is pretty flakey so I'm open to ideas.

internal/resources/aws/dynamodb_table.go
```diff
@@ -23,29 +26,69 @@ type DynamoDbTableArguments struct {
...
 
-var DynamoDbTableUsageSchema = []*schema.UsageSchemaItem{
-	{Key: "monthly_write_request_units", DefaultValue: 0, ValueType: schema.Int64},
-	{Key: "monthly_read_request_units", DefaultValue: 0, ValueType: schema.Int64},
-	{Key: "storage_gb", DefaultValue: 0, ValueType: schema.Int64},
-	{Key: "pitr_backup_storage_gb", DefaultValue: 0, ValueType: schema.Int64},
-	{Key: "on_demand_backup_storage_gb", DefaultValue: 0, ValueType: schema.Int64},
-	{Key: "monthly_data_restored_gb", DefaultValue: 0, ValueType: schema.Int64},
-	{Key: "monthly_streams_read_request_units", DefaultValue: 0, ValueType: schema.Int64},
+func (args *DynamoDbTableArguments) populateFromUsage(u *schema.UsageData) {
```
tim775, line 48: The idea is to map values from usage to args.  Note that usage take precedence over existings values -- the idea is to give the end user a way to override the IaC parsing.  I guess this is readable but it seems very verbose.  I'm going to look into putting "`infracost:`" tags on DynamoDbTableArguments.

internal/providers/terraform/aws/dynamodb_table.go
aliscott: Could we use the same `infracost:` tags as you suggested below? Something like `infracost:"excludeSync"`?

Issue #778, opened 2021-06-04
Automatically detect AWS EC2 AMI costs


alikhajeh1:
A few users have requested this and our answer has been to use the usage file so far:
https://github.com/infracost/infracost/issues/26
https://github.com/infracost/infracost/issues/272
https://github.com/infracost/infracost/issues/348
https://twitter.com/gortron/status/1348209997583437824

@PatMyron pointed out [an AWS announcement](https://aws.amazon.com/about-aws/whats-new/2020/02/amazon-ec2-adds-ability-to-easily-query-billing-information-of-amazon-machine-images/) that can help us. I did a quick test for a RHEL AMI and if we find the `UsageOperation` from `aws ec2 describe-images --image-ids ami-091967a4c3dbe5eac` and pass it as a filter [here](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/instance.go#L130) (and remove the `operatingSystem` filter) then it returns the correct price. However, we should check if we currently have all [EC2 Operating System Marketplace](https://aws.amazon.com/marketplace/b/c3bc6a75-0c3a-46ce-8fdd-498b6fd88577) prices in the Cloud Pricing API.

The aim of this ticket is update the `aws_instance`, `aws_autoscaling_group` and `aws_eks_node_group` (they all use the same "Instance usage" cost component logic) such that:
1. if usage file's `operating_system` is present, use that as that's the user telling us what they want to see the cost estimate for.
2. else if Infracost is being run with a non-plan-json path (e.g. a Terraform directory), try to make the previously mentioned AWS API call to find the correct price.
3. else default to Linux (as we do just now).

We should also update [this FAQ](https://www.infracost.io/docs/faq#does-infracost-need-cloud-credentials) to mention that we sometimes use the cloud creds to find metadata that impacts price, e.g. AWS AMIs.

ekifel:
@alikhajeh1, I want to work on it!

alikhajeh1:
Released in v0.9.9.

PR #841, opened 2021-06-30
feat(aws): add automatically detect AWS EC2 AMI costs

ekifel:
Issue #778 

Initial support


ekifel:
@alikhajeh1, this is ready for initial review. 
There are strange thing with failed test. It expects correct pricing (±67$, the same as `aws_instance` test) but locally it doesn't fails on that and put 155$

aliscott:
Thanks @ekifel, closing this PR just now since I've not had a chance to dive into this more. We need to do some more investigation into checking the creds, since at the moment this adds significant time for each EC2 resource.  I'll pick this up again at a later date and will reopen the PR then.

PR #851, opened 2021-07-01
Add option for enabling dashboard

aliscott:
Here's a summary of the main changes here:
* There's no global config anymore, it's all passed in to functions using contexts. There's two levels contexts:
    * `RunContext` - config/metadata for the overall run
    * `ProjectContext` - config/metadata for a single project in a run
* The run environment metadata is now set with `ctx.SetContextValue(`. It is write-only, to prevent us from using these variables like globals.
* `summary` has been removed from the top-level JSON and now each project has a `summary` key
* We add a short hash to the end of the project name if we can't detect a VCS repo. The hash is based on the absolute file path. This is to make sure that the project can be uniquely identified.
* `INFRACOST_ENABLE_DASHBOARD` env variables sends run results to (not yet available) Infracost dashboard
* CLI errors and telemetry now get sent to the dashboard API instead of pricing API


alikhajeh1 (approved):

alikhajeh1:
omg

Issue #729, opened 2021-05-27
Review/fix case sensitive Terraform inputs


alikhajeh1:
- [x] Google
- [x] Azure
- [x] AWS

Terraform ignores case sensitivity in most cases for AWS, GCP and Azure (I haven't checked other providers). The aim of this GH issue is to review and update all existing cloud resources to ensure that:

1. case-insensitive string comparison are used, e.g.:
Bad: `input == "large"`
Good: `strings.ToLower(input) == "large"`
We might be able to grep for something like `== "` to review/fix them all.

2. case-insensitive `Value` or `ValueRegex` price filters are used for any user input, e.g.:
Bad: `{Key: "virtualInterfaceType", ValueRegex: strPtr(fmt.Sprintf("/%s/", virtualInterfaceType))},`
Good: `{Key: "virtualInterfaceType", ValueRegex: strPtr(fmt.Sprintf("/%s/i", virtualInterfaceType))},  # notice the `/i` in the regex`
We might be able to grep for something like `, Value` to review/fix them all.

As part of this issue, we should also update the contributing guide.

krishnaindani:
Is this issue open to take? I am new and would like to learn and contribute since it is tagged as the good first issue.

alikhajeh1:
@krishnaindani sure! Thanks for offering to help.

krishnaindani:
> @krishnaindani sure! Thanks for offering to help.

My pleasure and thank you for giving me a chance to work on it. I will go through a bit and may ask some questions before getting started.

krishnaindani:
While going through code I see there is a lot of comparisons in the internal/providers directory where it is something like let's say `input == "LARGE"`.  Do we need to fix them all with strings.ToLower(input) == "large" to convert everything to lower case for comparison

alikhajeh1:
@krishnaindani yep, the `internal/providers/terraform` contains the files that need to be reviewed and fixed. I think the tests should not need to be changed and they should continue to pass, but feel free to post here if you find a test that looks like it needs to be changed.

krishnaindani:
> @krishnaindani yep, the `internal/providers/terraform` contains the files that need to be reviewed and fixed. I think the tests should not need to be changed and they should continue to pass, but feel free to post here if you find a test that looks like it needs to be changed.

Thanks. Yeah from the few changes with the aws I don't see the changes are required for the tests. But will continue to check more on AWS and on GCP and azure as well.

krishnaindani:
[...]

PR #854, opened 2021-07-05
Fix case comparisons for aws

krishnaindani:
- Second PR to fix https://github.com/infracost/infracost/issues/729. This takes care of fixing the case comparison for the aws.


krishnaindani:
@alikhajeh1 Got this for aws, would you review this PR when you have some time?

krishnaindani:
Thank you for getting this merged.

PR #856, opened 2021-07-05
fix(azure): case comparison

krishnaindani:
PR to fix case comparisons for Azure
Issue: https://github.com/infracost/infracost/issues/729

### DIFF
--- a/internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go
+++ b/internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go
@@ -2,6 +2,7 @@ package azure
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/infracost/infracost/internal/schema"
 	"github.com/shopspring/decimal"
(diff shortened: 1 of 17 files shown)


krishnaindani:
@alikhajeh1 I see in the azure tests are not running here https://github.com/infracost/infracost/runs/2992930511?check_suite_focus=true? Does someone need to take any actions there to run? Also see most of the comparisons in Azure are done in a string case insensitive way, so that's a good thing.

krishnaindani:
One more thing, in General guidelines I already see it mentions to use string case insensitive comparisons, so I guess it's already updated. https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md#general-guidelines (on the last bullet point)

alikhajeh1:
@krishnaindani we only run the azure tests on the master branch updates as they need cloud creds (which we can't set in the PR sender's fork). They were fine on my machine so I merged this so I merged this and we'll see if there are any issues in [this](https://github.com/infracost/infracost/runs/2996456349). 

Contributing guide was updated a while ago, that's why the Azure resources were mostly ok :)

krishnaindani:
> this

gotcha, that makes sense. Thanks

krishnaindani:
> @krishnaindani we only run the azure tests on the master branch updates as they need cloud creds (which we can't set in the PR sender's fork). They were fine on my machine so I merged this so I merged this and we'll see if there are any issues in [this](https://github.com/infracost/infracost/runs/2996456349).
> 
> Contributing guide was updated a while ago, that's why the Azure resources were mostly ok :)

gotcha, that makes sense. Thanks

PR #859, opened 2021-07-06
Next version changes

aliscott:


infracost-example.yml
```diff
@@ -7,3 +7,5 @@ version: 0.1
 projects:
   - path: examples/terraform
     usage_file: infracost-usage-example.yml # Define resource usage estimates, see https://infracost.io/usage-file
+  - path: examples/terraform
```
alikhajeh1, line 10: is this needed for testing?

alikhajeh1 (approved):

infracost-example.yml
aliscott: fixed

Issue #861, opened 2021-07-08
Infracost docker image runs with root


camaeel:
Infracost docker image requires to be started with root or to create a child image with dedicated user created. 
If I try to run it with other user:
```
$ docker run -it --rm --entrypoint sh -v ~/src/\[other\]/opa-tf/tmp/tfplan.json:/tfplan.json -u 1000:1000 infracost/infracost
```
 it throws a segmentation fault:
```
/root $ infracost breakdown --format json --path /tfplan.json
Error: mkdir /.config: permission denied
panic: runtime error: invalid memory address or nil pointer dereference
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xd039f7]

goroutine 1 [running]:
github.com/infracost/infracost/internal/apiclient.ReportCLIError(0x0, 0x13a3020, 0xc0001c0cc0, 0x44df2c, 0x20)
	/app/internal/apiclient/errors.go:12 +0x37
main.handleCLIError(0x0, 0x13a3020, 0xc0001c0cc0)
	/app/cmd/infracost/main.go:156 +0x79
main.main.func1(0xc000135ec8, 0x0, 0xc000074120)
	/app/cmd/infracost/main.go:32 +0xbf
panic(0xf359a0, 0x192a8f0)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/infracost/infracost/internal/config.(*RunContext).SetContextValue(...)
	/app/internal/config/run_context.go:55
main.main.func2(0xc000195b80, 0xc0001be6c0, 0x0, 0x4, 0x0, 0x0)
	/app/cmd/infracost/main.go:67 +0x66
github.com/spf13/cobra.(*Command).execute(0xc000195b80, 0xc0001be680, 0x4, 0x4, 0xc000195b80, 0xc0001be680)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:831 +0x582
github.com/spf13/cobra.(*Command).ExecuteC(0xc000195400, 0x351, 0xc000135ee8, 0x7)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
main.main()
	/app/cmd/infracost/main.go:121 +0xb79
[...]

PR #863, opened 2021-07-11
New Structure Proposal 01: Tag Based Approach

sinabakh:


sinabakh:
I tried to create an as minimal as possible POC for the ideas discussed on #861 
Here is a list of changes and ideas:

1. `PopulateArgs` is called so all `Args` can be populated with default values or the provided usage data.
2. Each `Arg` can have a tag for specifying the usage schema with the following format: `key_name,default_value,description,iac_1,iac_2,...`

This way:
A. The `infracost-usagefile` provider will just call the `PopulateArgs` and won't override anything.
B. Other IACs can override whatever fields they want
C. The usagefile can be autogenerated
D. Each IAC can filter few usagefile keys based on the supported fields.

internal/resources/aws/nat_gateway.go
```diff
@@ -6,16 +6,28 @@ import (
...
 
-	MonthlyDataProcessedGB *float64 `json:"monthlyDataProcessedGB,omitempty"`
+	MonthlyDataProcessedGB *float64 `json:"monthlyDataProcessedGB,omitempty" infracost_usage:"monthly_data_processed_gb,0,Monthly data processed by the NAT Gateway in GB,infracost,terraform"`
 }
 
-func (args *NATGatewayArguments) PopulateUsage(u *schema.UsageData) {
+func (args *NATGatewayArguments) PopulateArgs(u *schema.UsageData) {
+	address := strPtr("nat_gateway") // TODO: Better default value?
+	region := strPtr("")             // TODO: FIXME: A default value?
+	monthlyDataProcessedGB := floatPtr(0.0)
```
tim775, line 18: Is there anyway to use the info in the args tags to do this in a generic method?

internal/providers/terraform/aws/nat_gateway.go
```diff
@@ -15,11 +15,15 @@ func GetNATGatewayRegistryItem() *schema.RegistryItem {
...
-	args := &aws.NATGatewayArguments{
-		Address: d.Address,
-		Region:  region,
-	}
-	args.PopulateUsage(u)
+	// We initialize the args and populate the defaults and usage data.
+	args := &aws.NATGatewayArguments{}
+	args.PopulateArgs(u)
+
+	// Then we can override the fields with IAC fields.
```
tim775, line 22: I was thinking about this the other way around: the usage file could override the values pulled from IAC.   Maybe it would just be enough to do what you suggest and log that "Region from the usage file was ignored".

sinabakh: That makes sense, I'm not sure whichever works better. I mean what source of information is more reliable for us, IAC or usage-file.

internal/resources/aws/nat_gateway.go
sinabakh: Yes sure, I think we can do it.

sinabakh: I'll try to change the code to see if it works.

internal/resources/aws/nat_gateway.go
```diff
@@ -6,16 +6,28 @@ import (
 )
 
 type NATGatewayArguments struct {
-	Address string `json:"address,omitempty"`
-	Region  string `json:"region,omitempty"`
+	Address *string `json:"address,omitempty"`
+	Region  *string `json:"region,omitempty" infracost_usage:"region,us-east1,Region where gateway is located,infracost"`
 
-	MonthlyDataProcessedGB *float64 `json:"monthlyDataProcessedGB,omitempty"`
+	MonthlyDataProcessedGB *float64 `json:"monthlyDataProcessedGB,omitempty" infracost_usage:"monthly_data_processed_gb,0,Monthly data processed by the NAT Gateway in GB,infracost,terraform"`
```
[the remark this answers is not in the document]
sinabakh: That's a good point, I can think of few options but I did not have time to test them out. We probably can have a registry-like system for all resources and initialize a dummy instance of all resources to access their tags.

internal/providers/terraform/aws/nat_gateway.go
aliscott: I think there's cases for both.
1. The user might want to specify a default region in the usage file because we can't always get that from the IaC code.
2. The user might want to use the usage file to override the IaC code to get estimates for different instance counts in an autoscaling group.

For the no 1 use case, maybe `default_region` could just be a special case that we have a separate usage file key for.

sinabakh:
Hi! I've been away from development for few weeks and couldn't;t keep up with the changes. I wanted to continue where I left so is there any changes in plans or priorities?

sinabakh:
@aliscott @tim775 Here are all the required codes that I can think of for a POC, there is missing logic but the whole flow is complete and tests pass. I believe we can do some code reviews and discussion on this PR now to see if it's our way or not.

Btw, I will ping you on the parent issue so we can keep track of all possible solutions there and probably have multiple PRs.

internal/providers/terraform/aws/nat_gateway.go
aliscott: @sinabakh what did we decide for this? I think region should be treated the same way as every other field in the usage file.

aliscott:
I'm still hesistant about using tags as opposed to defining the schema in a map like Terraform does: https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/data_source_aws_nat_gateway.go. My main hesitation is any changes in the future, it's easier to add an optional field to a struct than a tag in this format.

One idea I've had if we want to use tags is to structure them more like the gorm tags (https://gorm.io/docs/indexes.html#Index-Tag) so we declare the keys, which means we can add more and the order doesn't matter as much. Something like this:

```
`json:"region,omitempty" infracost_usage:"name:region,default:us-east1,description:Region where gateway is located"`
```

sinabakh:
@aliscott Your suggestion about named tags is awesome! I've never thought of that.

Btw, to be honest I believe that tags are risky as you said. I will try to create a POC based on using structs in 1-2 days so we can see how it turns out.

aliscott:
Trying to also think of things that might come up in the future, that might be worth keeping in mind for just now since this will be a big refactor. Here's what I have so far:
* Custom validations
* Deprecating attributes
* Categorizing the attributes
* Required attributes

sinabakh:
Could you please explain "Categorizing the attributes"?

aliscott:
> Could you please explain "Categorizing the attributes"?

We might want to add metadata (tags/labels or categories) to group them in the usage file or generated docs in the future.

internal/providers/terraform/aws/nat_gateway.go
tim775: My vote is that the usage file should take precedence over IaC.  In other words, an arg calculated from IaC should only be used if the usage file value is empty.

I think the trick will be in making sure the sync_usage stuff only writes defaults for args that don't normally come from IaC.

sinabakh:
I also think the struct way of handling usage schema that @tim775 worked on seems more elegant and we can continue on that path.
The thing that I think we can discuss here is how we can automate the link between schema and args, currently we have sth like this:
```golang
args.MonthlyWriteRequestUnits = u.GetInt("monthly_write_request_units")
args.MonthlyReadRequestUnits = u.GetInt("monthly_read_request_units")
args.StorageGB = u.GetInt("storage_gb")
args.PitrBackupStorageGB = u.GetInt("pitr_backup_storage_gb")
args.OnDemandBackupStorageGB = u.GetInt("on_demand_backup_storage_gb")
args.MonthlyDataRestoredGB = u.GetInt("monthly_data_restored_gb")
args.MonthlyStreamsReadRequestUnits = u.GetInt("monthly_streams_read_request_units")
```

Any ideas?
I personally think we can have a tag in args that specifies the key name for the usage data. I believe that would be a clean use of tags.

internal/providers/terraform/aws/nat_gateway.go
sinabakh: @aliscott Sorry I missed your comment. I agree with @tim775

tim775:
Ok so then we're talking about replacing https://github.com/infracost/infracost/pull/839/files#diff-71ed323ee96294a7b798471f490d0ecde86fa1b63763f1e023d4bf93e40dc21eR48-R86 with something based on tags?  That makes a ton of sense to me since that's the most tedious part.  I think then the minimum we need is the name (since type can be inferred from the struct)?
So maybe we start with:
```go
type NATGatewayArguments struct {
	Address *string `infracost_usage:"address"`
	Region  *string `infracost_usage:"region"`

	MonthlyDataProcessedGB *float64 `infracost_usage:"monthlyDataProcessedGB"`
}
```

sinabakh:
Yes, that's exactly what I was thinking about. I'll try to do some refactoring in this PR as soon as possible and apply this structure.
Also, we probably have to think about how we will handle nested usage data.

tim775:
Closing this since the ideas have been incorporated to other PRs.  Thanks @sinabakh!

Issue #858, opened 2021-07-06
Update `infracost register` to let users know upfront if they already have an API key


alikhajeh1:
I've seen many users with multiple keys soon after they register. I wonder if it's because when they run `infracost register`, it creates the API key then checks/informs them if one already exists. We should change this so the first step in register is to check for an existing key, and tell them that it's saved in `~/.config/infracost/credentials.yml`, then ask if they'd like to create a new key.

krishnaindani:
Hey @alikhajeh1, is this open for work, I would like to try if that's okay?

alikhajeh1:
Yep, go for it, @aliscott will review it - not sure if he has any pointers before the work begins

aliscott:
@alikhajeh1 only question I have for this is: If the user has an API key and still wants to generate a new one, should it always overwrite the one they have saved in `~/.config/infracost/credentials.yml` or should they still get the option to overwrite it or not?

alikhajeh1:
It's probably better to give them a choice as if they overwrite it, there's no going back.

krishnaindani:
I have skimmed the code for the register command yesterday, will go through it so that I have a clear understanding, and ask questions if I get any. And then I can start the work. Thank you so much @alikhajeh1 and @aliscott.

krishnaindani:
> It's probably better to give them a choice as to if they overwrite it, there's no going back.

For this issue, are we addressing it by giving the user the option to provide a different path if they want? I have a bunch of questions if we are going to implement it that way.

alikhajeh1:
@krishnaindani I'd keep it simple and not worry about changing paths. The aim is to ensure users are aware that a key already exists before the existing flow kicks-in.

krishnaindani:
That means we are good with giving the warning here https://github.com/infracost/infracost/blob/master/cmd/infracost/register.go#L60 and overwriting it, right? We just need to change the flow to 

1. check if the key exists
2. If the key exists, ask the user if he still wants to generate. 
3. If yes to 2, overwrite the key that is saved in `config/infracost/credentials.yml`. 
4. If no to 2, will exit. 
 
Does that sound, correct?

alikhajeh1:
Almost...
1. Check if the key exists
2. If it exists, output `You already have an API key saved in %s, would you like to generate a new one?`
[...]

PR #864, opened 2021-07-13
fix: register command

krishnaindani:
Issue: https://github.com/infracost/infracost/issues/858

First Flow:

If the Key is not present:

1. We will prompt for name and email, generate the key and save it. 

Second Flow:

If the Key is present:

1. Ask the user `Would you like to generate a new one?` 
2. If no to 1, we exit. 
3. If yes to 1, we prompt for name, email, and generate the key and show them.
4. Then we prompt for `Would you like to overwrite your existing saved API key?`
5. If no to 4, we exit. 
6. If yes to 4, we save the new API key.

### DIFF
--- a/cmd/infracost/register.go
+++ b/cmd/infracost/register.go
@@ -20,6 +20,25 @@ func registerCmd(ctx *config.RunContext) *cobra.Command {
 		Short: "Register for a free Infracost API key",
 		Long:  "Register for a free Infracost API key",
 		RunE: func(cmd *cobra.Command, args []string) error {
+
+			var msg string
+			var isRegenerate bool
+
+			if _, ok := ctx.Config.Credentials[ctx.Config.PricingAPIEndpoint]; ok {
+
+				isRegenerate = true
+				fmt.Printf("\nYou already have an Infracost API key saved in %s.\n", config.CredentialsFilePath())
+
+				status, err := promptGenerateNewKey()
+				if err != nil {
+					return err
+				}
+
+				if !status {
+					return nil
+				}
+			}
+
 			fmt.Println("Please enter your name and email address to get an API key.")
 			fmt.Println("See our FAQ (https://www.infracost.io/docs/faq) for more details.")
 
@@ -36,6 +55,7 @@ func registerCmd(ctx *config.RunContext) *cobra.Command {
 			}
 
 			d := apiclient.NewDashboardAPIClient(ctx)
+
 			r, err := d.CreateAPIKey(name, email)
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


cmd/infracost/register.go
```diff
@@ -20,6 +20,25 @@ func registerCmd(ctx *config.RunContext) *cobra.Command {
...
 		Long:  "Register for a free Infracost API key",
 		RunE: func(cmd *cobra.Command, args []string) error {
+
+			var msg string
+			var isRegenerate bool
+
+			if _, ok := ctx.Config.Credentials[ctx.Config.PricingAPIEndpoint]; ok {
+
+				isRegenerate = true
+				fmt.Printf("\nYou already have an Infracost API key saved in %s.\n", config.CredentialsFilePath())
```
aliscott, line 30:
```suggestion
				fmt.Printf("You already have an Infracost API key saved in %s.\n", config.CredentialsFilePath())
```

We don't need a new line at the start here, since it's the first message that's printed

cmd/infracost/register.go
```diff
@@ -137,7 +154,7 @@ func promptForEmail() (string, error) {
 
 func promptOverwriteAPIKey() (bool, error) {
 	p := promptui.Prompt{
-		Label:     "Would you like to overwrite your existing saved API key",
+		Label:     "Would you like to overwrite your existing saved API key?",
```
aliscott, line 157:
```suggestion
		Label:     "Would you like to overwrite your existing saved API key",
```

promptui adds a `?` by default.

cmd/infracost/register.go
```diff
@@ -152,3 +169,20 @@ func promptOverwriteAPIKey() (bool, error) {
 
 	return true, nil
 }
+
+func promptGenerateNewKey() (bool, error) {
+	p := promptui.Prompt{
+		Label:     "Would you like to generate a new one?",
```
aliscott, line 175:
```suggestion
		Label:     "Would you like to generate a new API key",
```

promptui adds a `?` by default

cmd/infracost/register.go
```diff
@@ -49,43 +69,40 @@ func registerCmd(ctx *config.RunContext) *cobra.Command {
...
-				fmt.Sprintf("Your API key has been saved to %s", config.CredentialsFilePath()),
-				ui.PrimaryString("infracost breakdown --path=..."),
-			)
-
-			saveAPIKey := true
-
-			if _, ok := ctx.Config.Credentials[ctx.Config.PricingAPIEndpoint]; ok {
-				fmt.Printf("\nYou already have an Infracost API key saved in %s\n", config.CredentialsFilePath())
+			if isRegenerate {
 				confirm, err := promptOverwriteAPIKey()
```
aliscott, line 73:
```suggestion
                                 fmt.Println()
				confirm, err := promptOverwriteAPIKey()
```

We should add a new line before this prompt

We should print a new line here

aliscott (changes requested):
Thanks @krishnaindani, this works great 🚀. A couple of suggestions for the output.

cmd/infracost/register.go
krishnaindani: oh didn't know that it will add. Thanks, will change it.

cmd/infracost/register.go
krishnaindani: will change it

cmd/infracost/register.go
krishnaindani: sure

cmd/infracost/register.go
krishnaindani: sounds good.

krishnaindani:
Thank you for the review @aliscott. Pushed the changes for the one you mentioned.

aliscott:
Thanks @krishnaindani this is great!

krishnaindani:
You're Welcome!! Thank you so much for your review and opportunity to contribute.

PR #873, opened 2021-07-23
fix: Update to use latest filters

tim775:


internal/providers/terraform/aws/testdata/ebs_volume_test/ebs_volume_test.golden
```diff
@@ -6,7 +6,7 @@
                                                                                                   
  aws_ebs_volume.gp3                                                                               
  ├─ Storage (general purpose SSD, gp3)                     40  GB                           $3.20 
- ├─ Provisioned throughput                                  5  Mbps                         $0.20 
+ ├─ Provisioned throughput                             0.0048  Gbps                         $0.20
```
aliscott, line 9: Do we still want to show this in Mbps since that's what the user sets in the Terraform?

aliscott (approved):

internal/providers/terraform/aws/testdata/ebs_volume_test/ebs_volume_test.golden
tim775: Yes.  Is there an obvious way to do that?  I was thinking I'll convert UnitMultiplier to a decimal and then I can set it to 1/1024.

aliscott: Ah, I was assuming this would be easy with UnitMultiplier yeah, but didn't realise we'd have to update it to decimal. Happy to ship this as is just now and fix this later.

tim775: I did the conversion (seems like the time to do it) but it was a lot of files so we can back it out if you want.

aliscott: Amazing!

PR #886, opened 2021-07-29
added multiple free resources

Theragus:
Azure DNS, Backup, Managed Applications, Azure Sentinel, Azure SQL, Azure Database, Azure VM Scale Sets, Availability Sets

### DIFF
--- a/internal/providers/terraform/azure/registry.go
+++ b/internal/providers/terraform/azure/registry.go
@@ -258,6 +258,7 @@ var FreeResources []string = []string{
 	"azurerm_subnet_network_security_group_association",
 	"azurerm_virtual_network",
 	"azurerm_virtual_network_gateway_connection",
+	"azurerm_local_network_gateway",
 
 	// Azure Notification Hub
 	"azurerm_notification_hub",
--- a/internal/providers/terraform/azure/registry.go
+++ b/internal/providers/terraform/azure/registry.go
@@ -146,11 +146,19 @@ var FreeResources []string = []string{
 	"azurerm_automation_variable_int",
 	"azurerm_automation_variable_string",
 
+	// Azure Backup & Recovery Services Vault
+	"azurerm_backup_policy_vm",
+	"azurerm_backup_policy_file_share",
+	"azurerm_site_recovery_network_mapping",
+	"azurerm_site_recovery_replication_policy",
+
 	// Azure Base
 	"azurerm_resource_group",
 	"azurerm_resource_provider_registration",
 	"azurerm_subscription",
 	"azurerm_role_assignment",
+	"azurerm_role_definition",
+	"azurerm_user_assigned_identity",
 
 	// Azure Blueprints
 	"azurerm_blueprint_assignment",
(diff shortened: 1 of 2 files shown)


internal/providers/terraform/azure/registry.go
```diff
@@ -166,8 +174,35 @@ var FreeResources []string = []string{
...
+	"azurerm_dns_ptr_record",
+	"azurerm_dns_srv_record",
+	"azurerm_dns_txt_record",
+	"azurerm_private_dns_a_record",
+	"azurerm_private_dns_a_record",
+	"azurerm_private_dns_cname_record",
+	"azurerm_private_dns_mx_record",
+	"azurerm_private_dns_ptr_record",
+	"azurerm_private_dns_srv_record",
+	"azurerm_private_dns_txtrecord",
```
alikhajeh1, line 192:
```suggestion

```

We support these as paid resources by showing the cost for the queries made to them, e.g.:
```
 azurerm_dns_a_record.withoutUsage                                                       
 └─ DNS queries (first 1B)          Monthly cost depends on usage: $0.40 per 1M queries  
```

internal/providers/terraform/azure/registry.go
```diff
@@ -236,11 +281,43 @@ var FreeResources []string = []string{
 	"azurerm_container_registry_token",
 	"azurerm_container_registry_webhook",
 
+	// Azure Sentinel
```
alikhajeh1, line 284: Out of interest, any ideas on which TF resource we can/should show [these costs](https://azure.microsoft.com/en-us/pricing/details/azure-sentinel/ ) in?

Maybe we don't mark `azurerm_sentinel_data_connector_*` as free and instead show the `$2 per GB-ingested` price for each of them?

internal/providers/terraform/azure/registry.go
```diff
@@ -236,11 +281,43 @@ var FreeResources []string = []string{
...
+	"azurerm_sql_firewall_rule",
+	"azurerm_sql_virtual_network_rule",
+	"azurerm_mssql_firewall_rule",
+
+	// Azure Storage
+	"azurerm_storage_table_entity",
+	"azurerm_storage_management_policy",
+	"azurerm_storage_blob_inventory_policy",
+
+	// Azure Virtual Desktop
```
alikhajeh1, line 309: Looking at https://azure.microsoft.com/en-us/pricing/details/virtual-desktop/, I assume the user has to provision the VMs so we'll show costs for those, and there's no additional fee for "configuring" the VMs with desktop apps? Sorry if that's a dump question, I haven't used these resources before...

alikhajeh1 (changes requested):
Much appreciated! This is great :) A couple of minor comments

internal/providers/terraform/azure/registry.go
Theragus: Hi,
thanks for the feedback. I would've thought that it is only included in the DNS Zone itself since MS is only charging per zones and per x million requests, which is independent from the number of records that the DNS zone has.

alikhajeh1: good point! we debated about it at the time and went with showing that cost component in the DNS records as the zone might be defined in another project or done outside of infra-as-code -- however, we could also show it in the DNS zone. That way users can decide where to [model the usage](https://www.infracost.io/docs/usage_based_resources/#infracost-usage-file), in the DNS zone or record, or both (so they're summed).

internal/providers/terraform/azure/registry.go
Theragus: To be in line with the DNS records, that would probably be a better idea to mark all the connectors with this pricing too. Good Suggestion.

Regarding the Azure Sentinel TF resource, the closest one would be `azurerm_log_analytics_workspace`.
The pricing on Sentinel is pretty weird because it is built on top of Log Analytics. 
If you enable Sentinel on a Log Analytics Workspace additional charges apply based on the amount of GB ingested. That means the standard rates for Log Analytics apply, and on top of that, you have to pay additionally the Azure Sentinel Charges per GB.

https://azure.com/e/8a92a412178147b081d19c771f43fca8

internal/providers/terraform/azure/registry.go
Theragus: Usually, the Host Pool is where the charges will apply because that's where the "work" is being done. 
These are the infrastructure costs that are not influenced by the amount of Desktop Apps or Remote Apps that are configured.
The other cost part is related to M365/O365 Licensing, which is not showing up in any Azure Bill.

internal/providers/terraform/azure/registry.go
Theragus: Looking at it from a larger scale it might make sense to only show this in the DNS Zone, especially if companies have a large number of DNS Records (>100), this could clutter up the output. 
Again I'm not 100% if the output in infracost would only show the resource type once or repeatedly for every record separately?

alikhajeh1: it would repeat for every resource, so as you suggested, it's probably best if we change that to roll it up under the DNS Zone.

internal/providers/terraform/azure/registry.go
alikhajeh1: Cool, I removed them from the free list in the master branch (I forgot to do it before merging this). I also added https://github.com/infracost/infracost/issues/889 for Azure Sentinel

internal/providers/terraform/azure/registry.go
alikhajeh1: I added https://github.com/infracost/infracost/issues/888 for Azure Virtual Desktop

PR #896, opened 2021-08-04
feat: Add configure command

aliscott:
This command allow you to configure Infracost CLI globally with an API key and pricing API endpoint (if self-hosting this).

Currently I've added it as the following commands:

```
infracost configure get <key>
infracost configure set <key> <value>
```

So you'd use it like:

```
infracost configure set pricing_api_endpoint http://self-hosted-pricing-api.internal
infracost configure set api_key MY_API_KEY
```

I was wondering if anyone has any thoughts/suggestions for what this command should be called? I considered calling it `infracost config` but was worried that would cause confusion with the `--config-file` flag which allows you to pass run configuration in.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,0 @@
-INFRACOST_PRICING_API_ENDPOINT=https://pricing.api.infracost.io
--- a/cmd/infracost/configure.go
+++ b/cmd/infracost/configure.go
@@ -15,8 +15,13 @@ var validConfigureKeys = []string{"api_key", "pricing_api_endpoint"}
 func configureCmd(ctx *config.RunContext) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "configure",
-		Short: "Configure Infracost",
-		Long:  "Configure Infracost CLI",
+		Short: "Display or change global configuration",
+		Long: `Display or change global configuration.
+
+Supported settings:
+  - api_key: Infracost API key
+  - pricing_api_endpoint: endpoint of the Cloud Pricing API
+`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Show the help
 			return cmd.Help()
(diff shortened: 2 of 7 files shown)


alikhajeh1 (commented):
LGTM! My suggestions:

1. In the help text, should we reword "Pricing API Endpoint" to "Cloud Pricing API endpoint" or is that too long? (`pricing_api_endpoint` is ok for the flag so maybe we keep the name as is so it matches this).

2. For the help text, the `gh` and `docker` CLIs use `--restart string` for flags, I wonder if this looks better instead of API_KEY_VALUE and https://cloud-pricing-api.internal (we want it to be clear that external URLs are ok too):
```
Display or change configuration settings for infracost.

Current respected settings:
- api_key: Infracost API key
- pricing_api_endpoint: endpoint of the Cloud Pricing API (defaults to https://pricing.api.infracost.io)

USAGE
  ...

EXAMPLES
      infracost configure set api_key string

  Set your Infracost Pricing API Endpoint:

      infracost configure set pricing_api_endpoint http://endpoint

  Set your Infracost API key for a specific Pricing API Endpoint:

      infracost configure set api_key string --pricing-api-endpoint http://endpoint
```

3. For `infracost configure get`, if there are multiple entries and there is no default (i.e. first time users of this command), should we output the first one? Currently it outputs this (also, notice the 2 spaces after Endpoint):
```
 ./build/infracost configure get api_key
Warning: Could not find an API key for Pricing API Endpoint  in your saved config
```

4. `./build/infracost configure set pricing_api_endpoint x1` with an empty `~/.config/infracost/credentials.yml` file causes a panic. Could we get it to output "No API key is set, run `infracost configure set api_key string` to set one or `infracost register` to register for a new one."

aliscott:
I've added new changes to simplify the command. This removes the multiple profiles, which is only used by us as developers to simplify it. It adds a version to the credentials file and migrates to the new format. It also removes the `--pricing-api-endpoint` flag to simplify the interface.

Issue #887, opened 2021-07-29
Add Support for Azure VPN Gateway


Theragus:
Terraform Resource:

`azurerm_virtual_network_gateway`

Pricing Information:

https://azure.microsoft.com/en-us/pricing/details/vpn-gateway/

VPN Gateway uses different SKU Sizes for per hour rates

praveenghuge:
@alikhajeh1 I can work on this

alikhajeh1:
@praveenghuge thank you! feel free to add a comment on what cost components/units you think the resource needs if you like feedback on that before the code (or we can do that if you prefer).

Also, plz note that the contributing docs have been updated based on the new Cloud Pricing API release, any feedback/issues on that is also very welcome :)

PR #901, opened 2021-08-08
feat(azure): add support for VPN Gateway

praveenghuge:
https://github.com/infracost/infracost/issues/887


alikhajeh1:
@praveenghuge thank you! I'll review this week.

infracost-usage-example.yml
```diff
@@ -874,3 +874,8 @@ resource_usage:
 
   azurerm_notification_hub_namespace.my_namespace:
     monthly_pushes: 1000000 # Monthly total number number of additional pushes.
+
+  azurerm_virtual_network_gateway.Basic:
+    s2s_tunnel: 12 # Total number of s2s tunnels.
```
tim775, line 879: I'm not sure if s2s_tunnel should be a usage field.  It looks like each tunnel is represented in terraform by the `azurerm_virtual_network_gateway_connection` resource.  I don't recall if there is a way to gather up all the subresources so we can tell if there are more than 10 (they are priced at $0.15/hr for each tunnel over 10), but if not maybe we just treat each one as an independent resource with the $0.15/hr charge).  @alikhajeh1 do you have an opinion?

internal/providers/terraform/azure/vpn_gateway.go
```diff
@@ -0,0 +1,171 @@
...
+	if sku == "Basic" {
+		sku = "Basic Gateway"
+	}
+
+	costComponents = append(costComponents, vpnGateway(region, sku))
+
+	if u != nil && u.Get("s2s_tunnel").Type != gjson.Null {
+		tunnel = decimalPtr(decimal.NewFromInt(u.Get("s2s_tunnel").Int()))
+		if tunnel != nil {
+			tunnelLimits := []int{10, 30}
```
tim775, line 41: I know there is an upper limit of 30 on the pricing page but I think it would be better to ignore it here.  As written if a user set s2s_tunnel: 45, they would get prices for 30 connections.  I think it would be better for them to get the price for 35, even though that supposedly isn't possible according to the pricing page.
```suggestion
			tunnelLimits := []int{10}
```

internal/providers/terraform/azure/vpn_gateway.go
```diff
@@ -0,0 +1,171 @@
...
+		Unit:           "hours",
+		UnitMultiplier: decimal.NewFromInt(1),
+		HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
+		ProductFilter: &schema.ProductFilter{
+			VendorName: strPtr("azure"),
+			Region:     strPtr(region),
+			Service:    strPtr("VPN Gateway"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/%s/i", sku))},
+				{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/%s/i", sku))},
```
tim775, line 99: Duplicate line.

internal/providers/terraform/azure/vpn_gateway.go
```diff
@@ -0,0 +1,171 @@
...
+		Name:           fmt.Sprintf("VPN Gateway (%s)", sku),
+		Unit:           "hours",
+		UnitMultiplier: decimal.NewFromInt(1),
+		HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
+		ProductFilter: &schema.ProductFilter{
+			VendorName: strPtr("azure"),
+			Region:     strPtr(region),
+			Service:    strPtr("VPN Gateway"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/%s/i", sku))},
```
tim775, line 98: Do we need to add start/end (`/^%s$/i`anchors to these sku regexes? Some of the skus 'overlap' so looking up `/VpnGw3/` will match VpnGw3 and VpnGw3AZ and I believe we would get multiple prices.

tim775 (changes requested):

internal/providers/terraform/azure/vpn_gateway.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+		}
+	} else {
+		costComponents = append(costComponents, vpnGatewayS2S(region, sku, tunnel))
+	}
+
+	if u != nil && u.Get("p2p_connection").Type != gjson.Null {
+		connection = decimalPtr(decimal.NewFromInt(u.Get("p2p_connection").Int()))
+		if connection != nil {
+			connectionLimits := []int{128, 250, 500, 1000, 5000, 10000}
```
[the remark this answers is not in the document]
praveenghuge: added testcase for VpnGw5 with 6000 connections

infracost-usage-example.yml
praveenghuge: removed s2s resource will create vpn_gateway_connection.go for vpn gateway connection resource

internal/providers/terraform/azure/vpn_gateway.go
```diff
@@ -0,0 +1,129 @@
...
+		},
+		PriceFilter: &schema.PriceFilter{
+			PurchaseOption: strPtr("Consumption"),
+		},
+	}
+}
+
+func vpnGatewayP2S(region, sku string, connection *decimal.Decimal) *schema.CostComponent {
+	return &schema.CostComponent{
+		Name:           "VPN gateway P2S tunnels",
```
tim775, line 90:
```suggestion
		Name:           "VPN gateway P2S tunnels (over 128)",
```

tim775 (changes requested):
Looks good, thanks for the quick turn around!  One quick change and then I think it's done.

I'll create a new issue for adding costs for `azurerm_virtual_network_gateway_connection` so we don't lose track of it.

tim775 (approved):

PR #912, opened 2021-08-19
WIP: feat/azurerm synapse

dgcaron:
I still need to add some cost components and am awaiting some feedback but I wanted to have this opened in advance.


dgcaron:
i am adding the dedicated sql pool costs and when I look at the calculator I see this geo redundant backup

image

appearantly it is turned on by default when you deploy using terraform but there is no field for it in the provider. you can disable it from the portal. should we control it using the usage file (so on by default and off if the usage file says so)?

alikhajeh1:
@dgcaron that makes sense, we've done something similar for [other toggles](https://github.com/infracost/infracost/blob/master/infracost-usage-example.yml#L253-L254) that TF doesn't support. 

thanks for digging deep into this!

dgcaron:
@alikhajeh1  the PR is still in draft as it isn't building yet (due to a linter error) and I need to do some cleanup. I'll try to wrap it all up before the end of the week. and then publish the PR ok?

alikhajeh1:
@dgcaron sure - thanks!

alikhajeh1:
@dgcaron anything we can help with on this PR?

dgcaron:
Hi ,

No I am so sorry. I simply didn't had the proper time to check all of the
moving parts here. I'll give it another go this week!

dgcaron:
so this was quite a delivery as the pricing for a synapse environment isn't as easy as it looks. i believe this implementation does display the correct behavior and gives the users the same options  as the calculator.

if you look at https://docs.microsoft.com/en-us/azure/synapse-analytics/plan-manage-costs#how-youre-charged-for-azure-synapse-analytics and the pricing calculator I believe all costs are included now. do you want me to squash the commits before you guys merge it?

alikhajeh1:
@dgcaron I pushed a fix for the linting, I think it just needed `make fmt`

PR #916, opened 2021-08-19
feat(cli): Add 'all' option for fields argument

### LINKED ISSUE #876: Proposal: Adding an option which cause the output to contain all the fields
Having an option such as `--fields=all` which caused the output to contain all the supported fileds: price, monthlyQuantity, unit, hourlyCost, monthlyCost.

kevinkengne:
Fixes #876 

Add`all` for the `--fields` argument, which causes the output to contain all the supported fileds: price, monthlyQuantity, unit, hourlyCost, monthlyCost.

**CLI command executed**

```sh
make run ARGS="breakdown --path examples/terraform --usage-file=examples/terraform/infracost-usage.yml --fields=all: 
```

**Output returned**
```

 Name                                                           Price  Monthly Qty  Unit         Hourly Cost  Monthly Cost 
                                                                                                                           
 aws_instance.web_app                                                                                                      
 ├─ Instance usage (Linux/UNIX, on-demand, m5.4xlarge)          $0.77          730  hours              $0.77       $560.64 
 ├─ root_block_device                                                                                                      
 │  └─ Storage (general purpose SSD, gp2)                       $0.10           50  GB                 $0.01         $5.00 
 └─ ebs_block_device[0]                                                                                                    
    ├─ Storage (provisioned IOPS SSD, io1)                      $0.13        1,000  GB                 $0.17       $125.00 
    └─ Provisioned IOPS                                        $0.065          800  IOPS               $0.07        $52.00 
                                                                                                                           
 aws_lambda_function.hello_world                                                                                           
 ├─ Requests                                                    $0.20          100  1M requests        $0.03        $20.00 
 └─ Duration                                            $0.0000166667   25,000,000  GB-seconds         $0.57       $416.67 
                                                                                                                           
 OVERALL TOTAL                                                                                                   $1,179.31
```

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -258,20 +258,23 @@ func loadRunFlags(cfg *config.Config, cmd *cobra.Command) error {
 	cfg.ShowSkipped, _ = cmd.Flags().GetBool("show-skipped")
 	cfg.SyncUsageFile, _ = cmd.Flags().GetBool("sync-usage-file")
 
+	includeAllFields := "all"
 	validFields := []string{"price", "monthlyQuantity", "unit", "hourlyCost", "monthlyCost"}
 	validFieldsFormats := []string{"table", "html"}
 
 	if cmd.Flags().Changed("fields") {
-		if c, _ := cmd.Flags().GetStringSlice("fields"); len(c) == 0 {
+		fields, _ := cmd.Flags().GetStringSlice("fields")
+		if len(fields) == 0 {
 			ui.PrintWarningf("fields is empty, using defaults: %s", cmd.Flag("fields").DefValue)
 		} else if cfg.Fields != nil && !contains(validFieldsFormats, cfg.Format) {
 			ui.PrintWarning("fields is only supported for table and html output formats")
+		} else if len(fields) == 1 && fields[0] == includeAllFields {
+			cfg.Fields = validFields
 		} else {
-			fields, _ := cmd.Flags().GetStringSlice("fields")
 			vf := []string{}
 			for _, f := range fields {
 				if !contains(validFields, f) {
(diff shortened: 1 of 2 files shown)


cmd/infracost/run.go
```diff
@@ -258,20 +258,23 @@ func loadRunFlags(cfg *config.Config, cmd *cobra.Command) error {
...
+	includeAllFields := "all"
 	validFields := []string{"price", "monthlyQuantity", "unit", "hourlyCost", "monthlyCost"}
 	validFieldsFormats := []string{"table", "html"}
 
 	if cmd.Flags().Changed("fields") {
-		if c, _ := cmd.Flags().GetStringSlice("fields"); len(c) == 0 {
+		fields, _ := cmd.Flags().GetStringSlice("fields")
+		if len(fields) == 0 {
 			ui.PrintWarningf("fields is empty, using defaults: %s", cmd.Flag("fields").DefValue)
 		} else if cfg.Fields != nil && !contains(validFieldsFormats, cfg.Format) {
```
tim775, line 269: Did you happen to look into how this else if works (and if it makes sense with the 'all' option)?  I'm not 100% familiar with this so I can check it out if you haven't.  I'm guessing that cfg.Fields can get populated from a config file so we might need to support 'all' there too.

aliscott: We don't support fields in the config file yet. The config file fields are defined [here](https://github.com/infracost/infracost/blob/master/internal/config/config_file.go#L16-L19) and [here](https://github.com/infracost/infracost/blob/master/internal/config/config.go#L15-L24).

Side note: not sure why we have [YAML defs for other config values](https://github.com/infracost/infracost/blob/master/internal/config/config.go#L26-L45) - we should remove these if not used so it's not confusing.

tim775 (approved):
Nice job, thank you!

PR #922, opened 2021-08-26
New Structure Proposal 02: Struct Based Approach

sinabakh:


sinabakh:
@tim775 Yeah I think that's a good idea. When it's done we can discuss the final structure. I think it needs some tweaks especially in naming the functions.

tim775:
Closing this since the ideas have been incorporated to other PRs.  Thanks @sinabakh!

PR #926, opened 2021-08-30
test: Capture logs in golden file tests

tim775:
test(aws): Capture logs for aws golden file tests

### DIFF
--- a/internal/providers/terraform/aws/autoscaling_group_test.go
+++ b/internal/providers/terraform/aws/autoscaling_group_test.go
@@ -7,10 +7,11 @@ import (
 )
 
 func TestAutoscalingGroup(t *testing.T) {
-	t.Parallel()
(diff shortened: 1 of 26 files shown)


tim775:
Hey @sinabakh, any chance you have a moment to review?  I believe you suggested doing this a while ago 😉

sinabakh:
@tim775 Nice! LGTM. However, I was thinking about the fact that if we add support for a missing resource then some tests will fail. Am I right?

tim775:
>  However, I was thinking about the fact that if we add support for a missing resource then some tests will fail. Am I right?
Yeah that's true you would need to `-update` to fix any tests that break.  That doesn't seem like to big of a hassle and I guess it's expected since you really are changing what output is generated.

sinabakh:
That makes sense. Thanks

PR #945, opened 2021-09-07
feat: framework for estimating usage via cloud API query

xeger:
This is the skeleton with no resources implemented. To see how it's applied, see [this differential PR](https://github.com/xeger/infracost/pull/2) in my fork.

## Remaining Work Before Merge

- [x] get consensus on how to handle limitations (below) & uncomment or delete as needed
- [x] cache terraform output to avoid performance hit from 2x LoadResources
- [x] collect data on usages that came from the cloud; print status at end of run
- [x] framework for prompting & remediating measurability issues at end of run
- [x] fix gocritic issues
- [x] fix broken instant-output with new usage during sync
- [x] fix bug with `projects` vs `providerProjects`
- [ ] add tests (if anyone can think of useful ones?)

### Ignorant of IaC credential handling

Does nothing to mimic Terraform's AWS credential-selection code, so it relies on matching `AWS_` env vars to be exported in the shell, and only works for cases where all projects use identical AWS credentials.

IMHO this is acceptable for now since the user can work around it by running IC in several individual dirs after exporting the right creds.

### Does not incorporate fresh usage into current run

Resource costs are computed (from usage + price) at load time, but we can't sync usage until after the resources are loaded because we need the names/IDs of existing resources. It follows that we can't use fresh usage from the cloud to show cost during the same run.

A naive solution would be to load the project's resources a second time after syncing usage; this doubles the tool's run time when `--sync-usage` is applied, since loading Terraform resources takes awhile. This solution is shown, commented out, with a `TODO` to remind us that it needs to get better.

Some options:

1. Ship what we have & deal with the performance hit
2. Separate the cost-computation code from the resource-loading code in order to recalculate costs after syncing usage
3. Move cloud querying back to resource-load time (i.e. where we pass a useless `context.Context` to PopulateUsage) and introduce an `UpdateUsage` method of Resource that copies load-time usage params back into a usage map. Would also need to figure out how to suppress cloud calls unless `--sync-usage` is passed (i.e. something in the context) - since it's rude to make cloud API calls (which could easily fail) when someone isn't syncing usage.


cmd/infracost/run.go
```diff
@@ -103,6 +104,13 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 			if err != nil {
 				return err
 			}
+			// TODO: reload resources to use freshly-synced usage estimates from cloud
+			//   - how to do this without drastically increasing runtime?
+			//   - we just need to recompute costs...maybe separate that from loading?
+			// err = provider.LoadResources(project, u)
```
tim775, line 110: I'm ok with reloading the resources.  I'm expecting --sync-usage to be run fairly infrequently.  Also I'm curious how much time hitting the aws APIs will add.  It might be neat to add a 'runTime' metric to our mixpanel data.

aliscott: Cmurrently `LoadResources` does the following
* 1 - Runs Terraform init, plan, show and loads the resources into a `map[string]*schema.ResourceData`.
* 2 - Loops through the `map[string]*schema.ResourceData`, finds the matching usage data and constructs an array of `[]*schema.Resource`

My concern is that `#1` is often slow, since `terraform plan` hits the cloud vendor APIs. On larger projects I've seen this step take >3 minutes, so having to repeat this part might be an issue. I wonder if we could split this into those two stages and sync the usage data in between `#1` and `#2`, so we can pass the usage data when the resource is constructed?

Some ideas:
1. Refactor `syncResourcesUsage` to take `[]*schema.ResourceData` instead of `[]*schema.Resource`. Not looked into this too much yet, so might be difficult.
2. Somehow cache the terraform plan JSON when loading resources, so the second time it just loads it from the plan JSON.

xeger: I solved this by caching plan + state JSON at the provider level, per @aliscott's suggestion.

cmd/infracost/run.go
```diff
@@ -91,15 +91,26 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 			return err
 		}
 
-		projects = append(projects, providerProjects...)
+		usageProjects := make([]*schema.Project, len(projects))
```
xeger, line 94: @aliscott wanted to call out this slight hack due to the terragrunt stuff combined with reloading resources after syncing usage.

Do we mean to pass _all_ projects to every estimation run, and not just this local provider's? If I had a config file with 3 project roots, wouldn't the usage file for each one contain all resources or summat?

aliscott: You're right, this looks like a bug. We should only pass the provider's projects to sync usage each time.

aliscott (approved):
LGTM

tim775 (approved):

PR #948, opened 2021-09-07
Fix infracost outputting to stdout

### LINKED ISSUE #947: Infracost not writing to json file
The latest master branch has an issue as the JSON output isn't being saved to files, repro steps:
1. on master branch, run `make build`
2. `./build/infracost breakdown --format json --path examples/terraform > a.json`
3. Notice how the output is sent to the screen and a.json is empty

Infracost v0.9.6 does not have this issue.

Should we update [this line](https://github.com/infracost/infracost/blob/master/.github/workflows/go.yml#L36) and line 44 to do something like `[ -s infracost-output.json ] || echo "file is empty" && exit 1` so we catch this issue in CI in the future?

aliscott:
Fixes #947


alikhajeh1:
LGTM - thanks!

Issue #943, opened 2021-09-06
Can't run with Terraform Cloud with prefix mode


alex-3sr:
Hi All,

I'm trying since this morning to start working infracost in Azure DevOps PR by using Terraform Cloud.
I encounter 2 difficulties during this configuration.

First, even by set **INFRACOST_TERRAFORM_CLOUD_TOKEN** with the correct token, it always need a **.terraformrc** file
```
Running infracost breakdown using:
  $ infracost breakdown --no-color --format json --path /home/vsts/work/1/s/Terraform
time="2021-09-06T16:27:34Z" level=info msg="Detected Terraform directory at /home/vsts/work/1/s/Terraform"
time="2021-09-06T16:27:34Z" level=debug msg="Checking if /home/vsts/work/1/s/Terraform is a git repo"
time="2021-09-06T16:27:34Z" level=debug msg="Creating temporary config file for Terraform credentials"
time="2021-09-06T16:27:34Z" level=debug msg="TF_CLI_CONFIG_FILE is set, copying existing config from .terraformrc to config to temporary config file /tmp/824491119"
Error: open /home/vsts/work/1/s/Terraform/.terraformrc: no such file or directory
##[error]Bash exited with code '1'.
```

The workaround is to set a task before for write a .terraformrc file like for example 
```yaml
- bash: |
    cat << EOF >> .terraformrc
    credentials "app.terraform.io" {
      token = "$(tf_token)"
    }
    EOF
  displayName: Write Terraform.rc file
  workingDirectory: $(System.DefaultWorkingDirectory)/$(tf_path)
```

I think that using **INFRACOST_TERRAFORM_CLOUD_TOKEN**, it could be enough and able to write .terraformrc if needed really, no  ?

BTW, now that I can go on and have a workaround, it seem that it doesn't like my backend configuration and got this error ->
```
Running infracost breakdown using:
  $ infracost breakdown --no-color --format json --path /home/vsts/work/1/s/Terraform
time="2021-09-06T16:39:21Z" level=info msg="Detected Terraform directory at /home/vsts/work/1/s/Terraform"
time="2021-09-06T16:39:21Z" level=debug msg="Checking if /home/vsts/work/1/s/Terraform is a git repo"
time="2021-09-06T16:39:21Z" level=debug msg="Creating temporary config file for Terraform credentials"
[...]

PR #949, opened 2021-09-07
Warn on invalid TF_CLI_CONFIG_FILE instead of error

aliscott:
As discussed in #943.

Instead of hard failing if the user has a non-existent `TF_CLI_CONFIG_FILE` when trying to copy it to a temporary file, this logs a warning and will just create a new blank temporary file. @alikhajeh1 do you think there's any issues with this? I'm hesitant in case it hides an error the user would want to know about.


alikhajeh1:
I think this is fine as the logs show what's happening. Did you mean to commit the `.terraformrc ` file in this PR?

aliscott:
> Did you mean to commit the .terraformrc file in this PR?

Cheers, fixed

Issue #953, opened 2021-09-08
How do I override the infracost command when running on GitLab (CI/CD) ?


jimsmith:
Scenario using GitLab and not always wanting merge requests and we want to see the output in the pipeline itself

Need the ability to override the infracost command that gets ran when in the pipeline and also need the ability to save the format as html and upload as an artifact.

This is what happens when there's **no** merge requests but also happens when there **is a merge request** -> notice the message `404 Project Not Found`

Also when attempting to get infracost_breakdown.json there is no such file

```
Uploading artifacts...
WARNING: /builds/REDACTED/REDACTED/terraform/infracost_breakdown.json: no matching files 
ERROR: No files to upload                          
Uploading artifacts...
WARNING: /builds/REDACTED/REDACTED/terraform/infracost_breakdown.json: no matching files 
ERROR: No files to upload                          
Cleaning up file based variables
```

```
$ /scripts/ci/diff.sh
Running infracost breakdown using:
  $ infracost breakdown --no-color --format json --path /builds/REDACTED/REDACTED/terraform/ --terraform-plan-flags "-var-file=${CI_PROJECT_DIR}/terraform/environments/aws/REDACTED/${REGION}/REDACTED.tfvars"
Running infracost output using:
  $ infracost output --no-color --format diff --path infracost_breakdown.json
Posting comment as post_condition is set to always
Posting comment to GitLab commit ebcca2edf6bea999b796e0f3c08317a66dfe9b2b
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2001  100    35  100  1966    204  11462 --:--:-- --:--:-- --:--:-- 11701
{"message":"404 Project Not Found"}$
```

Oh and not able to use your GitLab template as it does not have support for GitLab's Terraform State Management (HTTP backend state stored inside GitLab)

```
.infracost:
  stage: infracost
  extends:
  image:
    name: infracost/infracost:latest
    entrypoint: [""]
  variables:
    AWS_DEFAULT_REGION: eu-west-2
    AWS_ACCESS_KEY_ID: MASKED
    AWS_SECRET_ACCESS_KEY: MASKED
    format: html > infracost_project.html   # Does this even give me a format in HTML ?
    INFRACOST_LOG_LEVEL: "warn"
    post_condition: '{"always": true}'
    path: "${CI_PROJECT_DIR}/terraform/"
[...]

PR #954, opened 2021-09-09
Generate HTML output and keep JSON and HTML files so they c…

aliscott:
…an be used as assets.

This leaves the following files at the end of the diff.sh run, so they can be store as assets in the CI system.
- `infracost_breakdown.json`
- `infracost_breakdown.html`

I'm not sure if we want this by default or to have flag called something link `keep_assets`/`save_assets`/`create_assets`? @alikhajeh1 what do you think?

See https://github.com/infracost/infracost/issues/953 for a use case.

### DIFF
--- a/scripts/ci/diff.sh
+++ b/scripts/ci/diff.sh
@@ -17,6 +17,7 @@ process_args () {
   percentage_threshold=${6:-$percentage_threshold}
   post_condition=${7:-$post_condition}
   show_skipped=${8:-$show_skipped}
+  save_assets=${9:-$save_assets}
 
   # Validate post_condition
   if ! echo "$post_condition" | jq empty; then
@@ -351,6 +352,9 @@ load_azure_devops_env () {
 
 cleanup () {
   rm -f infracost_breakdown_cmd infracost_output_diff_cmd infracost_output_html_cmd
+  if [ -z "$save_assets" ]; then
+    rm -f infracost_breakdown.json infracost_output_html.json
+  fi
 }
 
 # MAIN
@@ -383,10 +387,12 @@ echo "Running infracost output using:"
 echo "  $ $(cat infracost_output_diff_cmd)"
 diff_output=$(cat infracost_output_diff_cmd | sh)
 
-infracost_output_html_cmd=$(build_output_cmd "infracost_breakdown.json" "html")
-echo "$infracost_output_html_cmd" > infracost_output_html_cmd
-html_output=$(cat infracost_output_html_cmd | sh)
-echo "$html_output" > infracost_breakdown.html
+if [ -n "$save_assets" ]; then
+  infracost_output_html_cmd=$(build_output_cmd "infracost_breakdown.json" "html")
+  echo "$infracost_output_html_cmd" > infracost_output_html_cmd
+  html_output=$(cat infracost_output_html_cmd | sh)
+  echo "$html_output" > infracost_breakdown.html
+fi
 
 echo "Running infracost output using:"
 echo "  $ $(cat infracost_output_html_cmd)"
--- a/scripts/ci/diff.sh
+++ b/scripts/ci/diff.sh
@@ -81,7 +81,7 @@ build_breakdown_cmd () {
 }
 
 build_output_cmd () {
-  output_cmd="${INFRACOST_BINARY} output --no-color --format diff --path $1"
+  output_cmd="${INFRACOST_BINARY} output --no-color --format $2 --path $1"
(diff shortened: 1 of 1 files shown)


alikhajeh1:
Thanks @aliscott! I've generally tried to keep the behavior between CI/CD systems the same but in this case I can see two options:
1. We add this logic to all CI diff scripts, use `save_assets=false` by default and users can set it to true if they want to. It feels like it should be false by default so Jenkins/Atlantis boxes don't fill-up.
2. Only add this to the diff.sh file, in this case it's probably ok to have it true by default as the supported CI/CDs get fresh state, but then we might add other CI systems to the existing diff.sh so maybe it's best to keep the current behavior (don't save by default).

I think I prefer #1.

alikhajeh1:
@aliscott what do you think about shipping #1? We need to re-test our CI/CD integrations for https://github.com/infracost/infracost/pull/980 so we can test this at the same time. Or are you still waiting for @jimsmith to see if this even helps their use-case?

jimsmith:
> @aliscott what do you think about shipping #1? We need to re-test our CI/CD integrations for #980 so we can test this at the same time. Or are you still waiting for @jimsmith to see if this even helps their use-case?

hey sorry I didn't reply back.

Within a day of me reporting the issue, I managed to get to redo our approach of alot of things.

I solved the whole 'gitlab-terraform' version of terraform by doing gitlab jobs that use native terraform so that solved alot of fundamental issues (gitlab shouldn't really be doing a bespoke name for terraform - it's a quickstart use nothing useful for integrating other terraform tools like this one)

as for infracost I created a gitab yml pipeline which is this:

```
# Show cloud cost estimate changes for Terraform.
.infracost:
  variables:
    INFRACOST_LOG_LEVEL: "${INFRACOST_CONTAINER_LOG_LEVEL}"
    post_condition: '{"always": true}'
    HTML_REPORT_FILE: "infracost_project.html"
  script:
    - cd ${TF_ROOT}
    - terraform init -reconfigure -backend-config=lock_address="${TF_HTTP_ADDRESS}/lock" -backend-config=unlock_address="${TF_HTTP_ADDRESS}/lock" # Don't change - this is required for HTTP backend Terraform State
    - 'infracost breakdown --path ${CI_PROJECT_DIR}/terraform/ --format html > ${HTML_REPORT_FILE} --terraform-plan-flags="${REDACTED}"'
    - 'infracost breakdown --path ${CI_PROJECT_DIR}/terraform/ --format table --terraform-plan-flags="${REDACTED}"'
  artifacts:
    paths:
      - ${TF_ROOT}/${HTML_REPORT_FILE}
```

aliscott:
Thanks @jimsmith, that's interesting about the `gitlab-terraform` image. I wan't aware of it before but it's good to know about in case anyone else runs into this. Just want to clarify that this PR is no longer needed to unblock your use case?

Either way, I think it will be a good feature to add here so will look at finishing this at some stage, but want to double check if we should make a push for it for the next release?

jimsmith:
> Thanks @jimsmith, that's interesting about the `gitlab-terraform` image. I wan't aware of it before but it's good to know about in case anyone else runs into this. Just want to clarify that this PR is no longer needed to unblock your use case?
> 
> Either way, I think it will be a good feature to add here so will look at finishing this at some stage, but want to double check if we should make a push for it for the next release?

Hey @aliscott  - yeah this PR is no longer needed for my use case, agree it will be a good feature though for other folk that stumble of this.

alikhajeh1:
Thanks @jimsmith! Two quick follow-ups:
1. What do you think about the idea of the CLI generating a tokenized link to the HTML report from each run? e.g. see [this example](https://dashboard.infracost.io/results/285db4b2-1467-41c0-a162-382ae7f87e89?token=932c8505d6cd3dd7c3cba4d45188eeec9988976ae0d366ccc5ae813ed1b4fc395dfba94d14d54babef943c23042787f66076f04a29a37dced8d0ae963e5cefd5). This would only work for users that use our hosted Cloud Pricing API and who have set an env variable such as `INFRACOST_GENERATE_REPORT=true`. The aim is to make it easy for users to share output from CI and CLI runs with the rest of the team or managers.
  
1. I'm curious why you use `post_condition: '{"always": true}'`? Is it to force Infracost to always run so the assets are created for each commit? It would be great if we can get your thoughts on https://github.com/infracost/infracost/discussions/1016

aliscott:
Closing this since we are moving to a more composable CI structure that should allow this functionality. Currently the GitHub action functionality can be seen here: https://github.com/infracost/setup-infracost.

PR #957, opened 2021-09-17
Proposal: use struct methods instead of functions for resources

aliscott:
I'm wondering if we should use struct methods for the generic resources by default, to avoid potential name clashes as we move over more resources to this format. Here's an example for the DynamoDB resource.

What do you think @tim775?


tim775:
Works for me.  In those cases where we have logic shared between resources I guess they would still be the 'old style' package functions which seems ok too.

aliscott:
@xeger FYI if you have any thoughts on this.

xeger:
What @tim775 said; seems reasonable & doesn't preclude us from sharing logic.

tim775 (approved):

PR #963, opened 2021-09-20
Move ASG, EKS node group, EC2 instance and EBS volume resources to new structure

aliscott:
This moves the following resources to the new format:
* Autoscaling group
* EKS node group
* EC2 instance
* EBS volume


internal/providers/terraform/aws/autoscaling_group.go
```diff
@@ -48,227 +44,152 @@ func NewAutoscalingGroup(d *schema.ResourceData, u *schema.UsageData) *schema.Re
...
-	} else if strings.ToLower(d.Get("placement_tenancy").String()) == "dedicated" {
-		tenancy = "Dedicated"
+		a.LaunchTemplate = newLaunchTemplate(data, u, a.Region, instanceCount, int64(0), onDemandPercentageAboveBaseCount)
+	} else if len(mixedInstanceLaunchTemplateRef) > 0 {
+		data := mixedInstanceLaunchTemplateRef[0]
+		a.LaunchTemplate = newMixedInstancesLaunchTemplate(data, u, a.Region, instanceCount, d.Get("mixed_instances_policy.0"))
 	}
 
-	instanceType := d.Get("instance_type").String()
+	a.PopulateUsage(u)
```
tim775, line 63: I think `u` is usageData for the autoscaling group right?  Is that ok or should we be passing in the usageData associated with the referenced LaunchConfiguration?

aliscott: Looking at the existing tests we set the usage on the ASG, e.g.:
```yml
  aws_autoscaling_group.asg_lc_reserved:
    reserved_instance_type: standard
    reserved_instance_term: 1_year
    reserved_instance_payment_option: no_upfront
```
So this works, but it won't generate the empty usage file correctly. I think I should add these as usage parameters on the ASG and pass them down to the launch template/configuration.

aliscott: Pushed a fix that will generate the empty usage file correctly.

internal/providers/terraform/aws/autoscaling_group.go
```diff
@@ -48,227 +44,152 @@ func NewAutoscalingGroup(d *schema.ResourceData, u *schema.UsageData) *schema.Re
...
-		lc := newLaunchConfiguration(launchConfigurationRef[0].Address, launchConfigurationRef[0], u, region)
-
-		// AutoscalingGroup should show as not supported LaunchConfiguration is not supported
-		if lc == nil {
-			return nil
-		}
-		schema.MultiplyQuantities(lc, desiredCapacity)
-		subResources = append(subResources, lc)
+		data := launchConfigurationRef[0]
+		a.LaunchConfiguration = newLaunchConfiguration(data, u, a.Region, instanceCount)
```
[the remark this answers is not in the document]
aliscott: 🙌  Added some comments explaining the structure.

internal/providers/terraform/aws/autoscaling_group.go
tim775: Ok that makes sense.  I guess we should think about adding golden files for --sync-usage-file output.

tim775 (approved):

PR #967, opened 2021-09-22
Query instance operating system

xeger:


internal/usage/aws/ec2.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	result, err := client.DescribeImages(ctx, &ec2.DescribeImagesInput{
+		ImageIds: []string{ami},
+	})
+	if err != nil {
+		return "", err
+	} else if len(result.Images) == 0 {
+		return "", nil
+	}
+	switch result.Images[0].Platform {
```
tim775, line 31: Nice start!  Just a heads up that I think we can get a better idea of the billing os by looking at the `UsageOperation` field instead of Platform.  There are some details sketched out in #778 and an example of the RunInstance mappings are https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/billing-info-fields.html

xeger: Wow; that's deep. I'll skim the relevant API docs before I enshrine anything with tests.

alikhajeh1:
Depending on how this works, we might want to tweak the "Elastic Compute Cloud (EC2)" row in https://www.infracost.io/docs/supported_resources/

internal/usage/aws/ec2.go
xeger: Good call; UsageOperation lets us infer RHEL and SUSE.

There are also "BYOL" variants of some usage classes, presumably cheaper since you pay the OS vendor directly vs. paying AWS. There is a `licenseModel` attribute query in CPAPI which I suspect is related, but I haven't delved into it at all yet. I'll leave it alone for now when building the cost component, but if you have any guidance LMK.

aliscott:
@xeger can you please reopen this against master? I didn't realise this would be closed when I merged and deleted the `infracost:feat/asg` branch.

xeger:
Done @aliscott; see #984

Issue #913, opened 2021-08-19
Diff command is showing differences even when "terraform show" does not.


azamora-zsc:
I find the tool really useful, but this situation is really weird and brings a lot of doubts about it. The matter is that I'm saving some terraform plans from several terraform modules and analyzing them with infracost. 

When I execute terraform show I get the following:
```
$ terraform show  plan.bin 

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:

Terraform will perform the following actions:

Plan: 0 to add, 0 to change, 0 to destroy.
```

Then I save the plan as json by executing this:
`$ terraform show --json plan.bin > plan.json`

and finally this to get the following result:
```
$ infracost diff --path plan.json                                                
Detected Terraform plan JSON file at plan.json

✔ Calculating monthly cost estimate

Project: plan.json

- aws_autoscaling_group.this
  -$12.83

    - aws_launch_configuration.this

        - Instance usage (Linux/UNIX, on-demand, t2.micro)
          -$9.78

        - EC2 detailed monitoring
          -$2.10

        - root_block_device

            - Storage (general purpose SSD, gp2)
              -$0.95

- aws_db_instance.this
  -$130

    - Database instance
      -$124

    - Database storage
      -$5.96

    - Database storage IOPS
      $0.00

- aws_route53_record.cert_validation
  Monthly cost depends on usage

    - Standard queries (first 1B)
      Monthly cost depends on usage
        -$0.40 per 1M queries

    - Latency based routing queries (first 1B)
      Monthly cost depends on usage
        -$0.60 per 1M queries

    - Geo DNS queries (first 1B)
      Monthly cost depends on usage
        -$0.70 per 1M queries

+ aws_route53_record.cert_validation["xx.yy.com"]
  Monthly cost depends on usage

    + Standard queries (first 1B)
      Monthly cost depends on usage
        +$0.40 per 1M queries

    + Latency based routing queries (first 1B)
      Monthly cost depends on usage
        +$0.60 per 1M queries

    + Geo DNS queries (first 1B)
      Monthly cost depends on usage
        +$0.70 per 1M queries

- aws_route53_record.guac_cdn_cert_validation
  Monthly cost depends on usage

    - Standard queries (first 1B)
      Monthly cost depends on usage
        -$0.40 per 1M queries

    - Latency based routing queries (first 1B)
      Monthly cost depends on usage
        -$0.60 per 1M queries

    - Geo DNS queries (first 1B)
[...]

PR #971, opened 2021-09-23
Fix parsing of module and indexed resources for Terraform v0.12

aliscott:


aliscott:
Fixes https://github.com/infracost/infracost/issues/913

tim775 (approved):

PR #973, opened 2021-09-24
added nat gateway associations as free resources

Theragus:
NAT Gateway and Public IPs are already charged separately


aliscott:
Thanks @Theragus! Tests are failing for unrelated reason.

Issue #884, opened 2021-07-28
Should using terraform workspace in config-file in addition to env variable error?


alikhajeh1:
Running `infracost breakdown --config-file infracost.yml --terraform-workspace=ali-test` correctly shows:
```
Error: --config-file flag cannot be used with the following flags: --path, --terraform-*, --usage-file

Our help text mentions the same thing:
      --config-file string            Path to Infracost config file. Cannot be used with path, terraform* or usage-file flags
```

However, running `INFRACOST_TERRAFORM_WORKSPACE=ali-test infracost breakdown --config-file infracost.yml` does not show the error. Is the environment variable overriding the workspace for all projects in the config file? That doesn't make much sense as why would you want that? So it feels like we should error with the same message...

If the current behavior is intended: our docs mentions the [precedence](https://www.infracost.io/docs/multi_project/config_file#precedence) but it's not clear what users should do from https://www.infracost.io/docs/iac_tools/terraform_cloud_enterprise#terraform-workspaces.

https://github.com/infracost/infracost/issues/884 is the issue that caused me to look into this.

alikhajeh1:
@ajoulie thanks for contributing! your PR is live in release v0.9.9

PR #976, opened 2021-09-24
884 error on terraform env var with config file

ajoulie:
This is an intent to fix https://github.com/infracost/infracost/issues/884. 

The CLI errors if any of the project options (terraform-*, path, and usage-file) are used in conjunction with the --config-file flag independent of the use of environment variables or flags. 

This PR fixes another small bug:  Don't override terraform-workspace flag by ENV variable
 - Old behavior: INFRACOST_TERRAFORM_WORKSPACE environment variable overrides --terraform-workspace flag
 - New behavior: INFRACOST_TERRAFORM_WORKSPACE environment variable does not overrides --terraform-workspace flag
 as mentioned in https://www.infracost.io/docs/multi_project/config_file#precedence.

Any feedback welcome.

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -284,21 +284,13 @@ func loadRunFlags(cfg *config.Config, cmd *cobra.Command) error {
 		cmd.Flags().Changed("terraform-workspace") ||
 		cmd.Flags().Changed("terraform-use-state"))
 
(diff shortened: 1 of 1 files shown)


aliscott:
Thanks @ajoulie 🙌 . This inspired me to add some initial golden tests for these cases in #981 and this PR passes them perfectly!

PR #977, opened 2021-09-24
build: bump to use go1.17

chenrui333:
- bump to use go1.17
- `go mod tidy` to list indirect deps in `go.mod`
- update `golang.org/x/sys` for go1.17 runtime


aliscott:
Thanks @chenrui333!

PR #980, opened 2021-09-24
feat: Post a pull request comment if possible

tim775:


scripts/ci/diff.sh
```diff
@@ -143,19 +144,78 @@ build_msg () {
...
-    GITHUB_SHA=$(cat $GITHUB_EVENT_PATH | jq -r .pull_request.head.sha)
-  fi
-
   if [ -z "$GITHUB_TOKEN" ]; then
     echo "Error: GITHUB_TOKEN is required to post comment to GitHub"
   else
-    echo "Posting comment to GitHub commit $GITHUB_SHA"
-    msg="$(build_msg true)"
+    if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then
+      GITHUB_PULL_REQUEST_NUMBER=$(echo $github_event | jq -r .pull_request.number)
```
alikhajeh1, line 151: I don't see `$github_event` in https://docs.github.com/en/actions/learn-github-actions/environment-variables, how's that set?

tim775: The `load_github_env method` does `github_event=$(cat $GITHUB_EVENT_PATH)`

scripts/ci/diff.sh
```diff
@@ -348,31 +428,31 @@ elif [ "$(echo "$post_condition" | jq '.has_diff')" = "true" ] && [ -n "$diff_re
...
   echo "Not posting comment as percentage diff ($absolute_percent%) is less than or equal to percentage threshold ($percentage_threshold%)."
   cleanup
   exit 0
 fi
 
-if [ ! -z "$GITHUB_ACTIONS" ]; then
+if [ -n "$GITHUB_ACTIONS" ]; then
   echo "::set-output name=past_total_monthly_cost::$past_total_monthly_cost"
   echo "::set-output name=total_monthly_cost::$total_monthly_cost"
   load_github_env
```
alikhajeh1, line 442: Is there a reason why we call this twice? it's already called on line 362 (maybe @aliscott knows as he added that functionality). Same issue with `load_gitlab_env` below

aliscott: Don't think there's a reason for this.

aliscott (approved):
LGTM

scripts/ci/diff.sh
alikhajeh1: @tim775 it seems like we can remove the lines `load_github_env` and `load_gitlab_env` below

PR #984, opened 2021-09-28
feat: detect instance operating system

xeger:
It took me awhile to wrap my head around ASG, LaunchTemplate and LaunchConfiguration but after some fiddling with TF plans and some grepping of AWS docs, I've got a somewhat elegant solution for OS estimation.


aliscott (approved):
LGTM!

PR #988, opened 2021-09-29
feat: Estimate usage of aws_autoscaling_group + aws_eks_node_group

xeger:
Work in progress. Note the TODOs.

I have prototyped _a posteriori_ estimation of an autoscaling group that has already been launched. Needs smoke testing & unit tests but it will probably work.

Note that in Infracost, EKSNodeGroup does not relate to AutoscalingGroup in any way, so the _a posteriori_ estimation will currently only work for ASG resources that are declared as such, not ASGs resulting from the creation of a node group.

Q1: Do we want the EKS node group to take related ASG names from the Terraform state and use them to perform estimation?

When someone runs Infracost, the EKS and ASG resources don't yet exist, necessarily - but the Terraform code may have hints for estimation, i.e. ASG and EKS can both specify min, max and desired node counts. Using these hints would let us estimate usage _a priori_ which may be useful -- maybe more useful than querying usage in real time, given the way people tend to use Infracost.

Q2: Do we want to use the capacity hints in Terraform code to populate usage? If so, do we want that to be _higher_ priority than the _a posteriori_ estimates, or would we rather fall back to the hints if real usage data doesn't exist yet in the cloud?


aliscott:
**Q0**

Currently the [`instanceCount`](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/autoscaling_group.go#L34) is passed to these functions which creates the subresources.

https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/autoscaling_group.go#L50
https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/autoscaling_group.go#L59
https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/autoscaling_group.go#L62

Note: for Mixed Instance Launch Templates, there is a [little more calculation here](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/autoscaling_group.go#L175) since it takes into account the `weighted_capacity` attribute. This logic may be IaC-agnostic so we might want to move it out of the Terraform provider at some point.

**Q1**

Agree, we should treat the cloud data with higher priority and fallback to the Terraform.

> Regardless how we count instances, I have a catch-22 due to the way we structured ASGs, EKS and their subresources

I think the best way of doing this would be to replace the panic with `u["instances"] = count`.
The subresources values are populated in the [`PopulateUsage` method of the ASG](https://github.com/infracost/infracost/blob/master/internal/resources/aws/autoscaling_group.go#L28-L34) so this should create a usage file like below which should populate the `InstanceCount` of the subresources:

```yml
resource_usage:
  aws_autoscaling_group.my_asg:
    instances: 2
```

Another option would be to do something like this:
```go
u[subResources[0].Name] = map[string]interface{}{
   "instances": count,
}
```

Which should create a usage file like:
```yml
resource_usage:
  aws_autoscaling_group.my_asg:
    aws_launch_template.my_lt:
      instances: 2
```

I'm not sure if this answers Q2 as well.

Q4: Should we even have subresources at all? Does a user care that an ASG uses a LT or a LC or do they just want to see the cost components added directly to the ASG?

tim775:
>  I'm wondering if it's better if our remediation can somehow update the IaC code itself. For ASGs for example we just need to add `enabled_metrics = ["GroupTotalInstances"]`. We can then show the cost to the user of enabling these using Infracost itself.

This makes a ton of sense to me, maybe the remediation step should be a separate CLI command.  The best part about this is that we don't need to warn about potential additional costs -- they can run infracost!

 I'm sure you've seen [yor](https://github.com/bridgecrewio/yor), but it's a tool for automatically adding tags to IaC.  If we want to process CURs, part of "remediation" could be to use yor to set up cost allocation tags.

xeger:
@tim775 and @alikhajeh1 -- re remediation, this is the way! We should remove `Remediable` & `remediater` and plan to deal with this via logging or code rewriting instead. (Actually, `Remediable` might have some use w/r/t the logging, but its interface needs to be rethought...)

Issue #992, opened 2021-09-30
Infracost CLI uses Terraform instead of Terragrunt in a Terragrunt project


sfenman:
Using terragrunt in a monorepo with many modules, I am trying to run infracost diff --path=/my/path/to/terragrunt/ but I get
Detected Terraform directory at .
```
  ✖ Running terraform plan 

  Terraform command failed with:
    Error: No configuration files
    Plan requires configuration to be present. Planning without a configuration
    would mark everything for destruction, which is normally not what is desired.
    If you would like to destroy everything, run plan with the -destroy option.
    Otherwise, create a Terraform configuration file (.tf file) and try again.

Error: Error running terraform plan: exit status 1
```

My project's structure:
```
root-dir
├── app-1
│   ├── aws-data
│   │   ├── data.tf
│   │   └── terragrunt.hcl
│   ├── ec2-app
│   │   └── terragrunt.hcl
│   ├── iam-app-user
│   │   └── terragrunt.hcl
│   ├── iam-instance-profile
│   │   └── terragrunt.hcl
│   ├── iam-instance-profile-policy
│   │   └── terragrunt.hcl
│   ├── postgres-rds
│   │   └── terragrunt.hcl
├── ec2-prometheus
│   └── terragrunt.hcl
├── app-3
│   ├── iam-user
│   │   └── terragrunt.hcl
│   └── s3
│       └── terragrunt.hcl
├── region.hcl
└── account.hcl

```
Infracost CLI recognizes my workspace as a Terraform project because of the `region.hcl` and `account.hcl` but these 2 are needed for my Terragrunt configuration. A quick solution is to add an empty terragrunt.hcl (Thanks to @tim775)

Perhaps it would be better to parse a flag or an environment variable to explicitly state if we want to run Terraform or Terragrunt?

PR #996, opened 2021-09-30
fix: don't detect Terraform dir by looking for .hcl files.

aliscott:
This messes with Terragrunt detection and Terraform needs .tf files anyway.

### DIFF
--- a/internal/providers/terraform/dir_provider.go
+++ b/internal/providers/terraform/dir_provider.go
@@ -406,7 +406,7 @@ func (p *DirProvider) runShow(opts *CmdOptions, spinner *ui.Spinner, planFile st
 }
 
 func IsTerraformDir(path string) bool {
-	for _, ext := range []string{"tf", "hcl", "hcl.json", "tf.json"} {
+	for _, ext := range []string{"tf", "tf.json"} {
 		matches, err := filepath.Glob(filepath.Join(path, fmt.Sprintf("*.%s", ext)))
 		if matches != nil && err == nil {
 			return true


tim775 (approved):
Looks good.  Just for a sanity check: Are we sure we're not breaking someone's setup?

tim775:
This fixes #992

aliscott:
> Just for a sanity check: Are we sure we're not breaking someone's setup?

When I run `terraform plan` with no .tf files, but with .hcl files I get the following error:

```
Error: No configuration files

Plan requires configuration to be present. Planning without a configuration would mark everything for destruction, which is normally not what is desired. If you would like to destroy everything, run plan with the -destroy option. Otherwise, create a Terraform configuration file (.tf file) and try again.
```

It looks like Terraform [uses this for detecting a Terraform dir](https://github.com/hashicorp/terraform/blob/7803f69d42c7ccedc962da308a4d1e033cacb0dd/internal/configs/parser_config_dir.go#L62) which [calls this](https://github.com/hashicorp/terraform/blob/7803f69d42c7ccedc962da308a4d1e033cacb0dd/internal/configs/parser_config_dir.go#L127).

From what I can tell it's safe, what do you think?

tim775:
Yes that seems safe.

alikhajeh1:
I thought I recognized the HCL json but it was from this PR https://github.com/infracost/infracost/pull/802/files so this PR doesn't break that :)

alikhajeh1:
This closes https://github.com/infracost/infracost/pull/996, correct?

Issue #849, opened 2021-07-01
Allow resources to use attributes from other resources that reference them


aliscott:
We currently allow resources to use attributes from resources they reference through the `.References` call but this is just a one way lookup. We want to allow resources to also use attributes from resources that reference them as well.

This issue is to keep track of the resources that need this and to start adding details about how we can support this.

Resources that need this:
* [AWS Application Auto Scaling](https://github.com/infracost/infracost/issues/440)

hugorut:
++ AWS Direct Connect Gateway (aws_dx_gateway) see #488

hugorut:
+ google_secret_manager_secret see reasoning on #1023

alikhajeh1:
It might be useful to see how tools like checkov parse the TF code into a graph-based framework

tim775:
I've added support for reverse references as part of #1401.  I'm leaving this issue open until we implement the resources already mentioned.

alikhajeh1:
Are aws_eip and google_compute_address done? If so, I'll close this issue as https://github.com/infracost/infracost/issues/440 is a separate issue anyway.

PR #1009, opened 2021-10-02
feat(aws): support cross region AWS Direct connect data transfer costs

### LINKED ISSUE #488: Support cross region AWS Direct connect data transfer costs.
Data transfer charges for AWS direct connect are currently calculated from the region where the connection is created. If you use DX gateway you can then incur data transfer costs from a specific region to your DX location. We should look to support both data transfer within the DX Gateway resource & / Or adding support into data transfer usage for AWS region to DX location.

hugorut:
Closes #488

Adds support for costing of data transfer from different regions to a direct connection location. Due to the limitations outlined in #488 and without #849 support, we cannot provide this support through "Resource" fetching, instead we need to rely on usage cost file. 

This PR removes prior dx location usage config property of `monthly_outbound_region_to_dx_location_gb` and adds `monthly_outbound_from_region_to_dx_connection_location` which allows users to specify an expanding map of regions that transfer data to the direction connection:

```yaml
    monthly_outbound_from_region_to_dx_connection_location:
        ap_east_1: 3000 # Monthly outbound data transferred to the DX location from ap-east-1 in GB
        eu_west_1: 1000 # Monthly outbound data transferred to the DX location from eu-west-1 in GB
```

This is then used in `DXConnectionRegistryItem` to build a dynamic list of cost components. The output of the above usage example is such:

```
 aws_dx_connection.my_dx_connection_usage                                               
  ...
 ├─ Outbound data transfer (from ap-east-1, to EqDC2)        3,000  GB          $270.00 
 └─ Outbound data transfer (from eu-west-1, to EqDC2)        1,000  GB           $28.20 
```

In order to support map structures this PR adds `UsageData.GetMap`. This supports retrieving a single level map from the usage file through regex matching.

The `GetMap` solution works for the current implementation and can likely handle other scenarios. However, in the future it might be best to remove **flattening** of attributes when parsing the usage file, see here: https://github.com/infracost/infracost/blob/285d9a072a4bec6ccd2cd6a1904dff2900c8f7b8/internal/schema/usage_data.go#L137-L145

---

Edit: Now no longer using `GetMap` instead favouring `Get` method with built nested keys.


internal/schema/usage_data.go
```diff
@@ -51,6 +52,45 @@ func (u *UsageData) Get(key string) gjson.Result {
...
+// 		GetMap with the "prop_1" as key will return a map:
+//
+//			map[string]gjson.Result{
+//				"sub_prop_1": ...,
+//				"sub_prop_2": ...,
+//		   }
+//
+// GetMap only support 1 level of nesting. Further levels must be handled another way.
+// If key cannot be found in UsageData GetMap returns nil.
+func (u *UsageData) GetMap(key string) map[string]gjson.Result {
```
tim775, line 73: Can this be done with `u.Get(key).Map()`?

internal/providers/terraform/aws/testdata/dx_connection_test/dx_connection_test.golden
```diff
@@ -1,14 +1,15 @@
...
+                                                                                        
+ aws_dx_connection.my_dx_connection_usage                                               
+ ├─ DX connection                                              730  hours       $219.00 
+ ├─ Outbound data transfer (from ap-east-1, to EqDC2)        3,000  GB          $270.00 
+ └─ Outbound data transfer (from eu-west-1, to EqDC2)        1,000  GB           $28.20 
+                                                                                        
+ OVERALL TOTAL                                                                  $736.20 
+Logs:
+
+level=warning msg="Skipping resource aws_dx_connection.my_dx_connection_usage usage cost: Outbound data transfer. Could not find mapping for region does_not_exist"
```
tim775, line 15: Nice touch!

infracost-usage-example.yml
```diff
@@ -123,7 +123,9 @@ resource_usage:
     backup_storage_gb: 10000      # Amount of backup storage that is in excess of 100% of the storage size for the cluster in GB.
 
   aws_dx_connection.my_dx_connection:
-    monthly_outbound_region_to_dx_location_gb: 100 # Monthly outbound data transferred from AWS region to DX location in GB.
```
tim775, line 126: I wonder if we should leave the old usage param in place as well?  The two reasons I can think of are 1) backward compatability since someone may already be using it, and 2) it might be a little nicer to be able to set the "normal" non-dx gateway traffic without having to explicity mention the region.  @alikhajeh1 any thoughts?

internal/schema/usage_data.go
hugorut: Sadly not as at the point of fetching this data (in the Resource) we have flattened this data. See these lines here: https://github.com/infracost/infracost/blob/285d9a072a4bec6ccd2cd6a1904dff2900c8f7b8/internal/schema/usage_data.go#L137-L168 .

This means that instead the expected result from gjson parsing of:

```go
UsageData{
    Attributes: map[string]gjson.Result{
        "monthly_outbound_from_region_to_dx_connection_location": gjson.Result{
            Type: "JSON",
            Raw: `{"ap_east_1": 3000}`
        }
    }
}
```

which support `Result.Map` - see here: https://github.com/tidwall/gjson/blob/590010fdac311cc8990ef5c97448d4fec8f29944/gjson.go#L289-L295

We get the following:

```go
UsageData{
    Attributes: map[string]gjson.Result{
        "monthly_outbound_from_region_to_dx_connection_location.ap_east_1": gjson.Result{
            Type: "JSON",
            Raw: `3000`
            Num: 3000
        }
    }
}
```

So `GetMap` supports unflattening from 1 level of nesting. As mentioned in the PR description I think it would be best in the future to try and remove flattening from the start. However I didn't want to do it in this PR as it basically touches every resource. Maybe we should put it into an issue to track?

internal/providers/terraform/aws/testdata/dx_connection_test/dx_connection_test.golden
hugorut: 🥳

infracost-usage-example.yml
hugorut: I think we should keep it to solve backwards compatibility. I have two questions though:

1. do we want to mark this as **deprecated** in the yaml comment in favour of the new approach
2. do we want the new approach to override the old one if the user specifies both? Or do we want to keep both working in tandem?

alikhajeh1: Great idea with `# DEPRECATED`! I think it's fine for the new approach to override the old one so we encourage/nudge people to move towards the new way.

internal/schema/usage_data.go
tim775: Would it work if we invert things and look for the specific keys, so in `NewDXConnection` do something like:
```go
if u != nil { 
    // sort the region keys so that we get a consistent output in the cli
    keys := make([]string, 0, len(regions))
    ...
    
    for _, key := range keys {
         if u.GetMap("monthly_outbound_from_region_to_dx_connection_location." + key ) != nil {
             // create the cost component
         }
    }
```
 I'm a little wary of adding GetMap right now since it seems like there's a lot of stuff inflight wrt usage params.  Also at some point we're going to port this to the 'provder agnostic' style which has a defined "UsageSchema" and I think it may be more straightforward to do that if we treat each value individually instead of as a map (take a look at `LambdaFunctionUsageSchema` in `internal/resources/aws/lambda_function.go` as an example)

hugorut: @tim775 sure thing can scrap the `GetMap` function and just use base `Get` with the key lookup as suggested. I'll change now. Just checking you meant to use: `u.Get("monthly_outbound_from_region_to_dx_connection_location." + key )` in the above example instead of `GetMap`

tim775: Oops, yes I meant `u.Get("monthly_outbound_from_region_to_dx_connection_location." + key )` (I edited my comment)

hugorut: Perf - I'll use the defined `regionMapping` to define a list of acceptable `keys` and then attempt to pull each using `Get` 👍

hugorut:
@tim775 fixed all PR comments:

* Remove `GetMap` now use simply `Get` whilst looping through `regionMap` keys
* Add legacy/backwards compatibility for `monthly_outbound_region_to_dx_location_gb` which is *only* added to costs if `monthly_outbound_from_region_to_dx_connection_location` is not specified.

tim775 (approved):
Looks great, thanks!

PR #1011, opened 2021-10-02
feat: usage keys validation

sinabakh:


sinabakh:
I don't feel that I know exactly what should be where regarding producing an output. So I just added a placeholder message generator in `run.go` that I guess is probably the wrong place. I was not sure to put it here or in the `table.go`.

alikhajeh1:
This is awesome @sinabakh! I think logging a warning is fine as we can still run so no need to error (e.g. for the [Plan JSON API](https://www.infracost.io/docs/integrations/infracost_api), which runs the CLI, the run can succeed with warnings). 

Can we use the yellow warning style we use for other warnings? e.g. see the following screenshot for when `fields` and `--format=json` are used together. @aliscott I guess we're ok with multiple warnings per run for now? Looks like we'll have a max of 2, we can think about collecting them into 1 warning later on.

Screen Shot 2021-10-07 at 3 11 31 PM

Can we also change the warning text to the following so it's 1 comma separated line? We use the word "parameter" in [docs](https://www.infracost.io/docs/usage_based_resources#supported-parameters).
```
Warning: the following usage file parameters are invalid and will be ignored: request_duratio, mnthly_requests, typo1 
```


Longer term we can think about validating the values and maybe even the resource paths, but for now this is a great start.

aliscott:
Thanks @sinabakh! I'm closing this since I've merged your commits in https://github.com/infracost/infracost/pull/1011. I've updated this to work with the new way we're structuring hierarchical usage data, and have made the UI changes @alikhajeh1 suggested.

PR #1014, opened 2021-10-03
Fix sync usage nil pointers

aliscott:
Another option might be to set `runCtx.Config.SyncUsageFile` to false if the `--usage-file` flag isn't passed in.


xeger (approved):
oops! lack of integration coverage strikes again. TY!

PR #1036, opened 2021-10-09
Add usage file YAML comments

aliscott:
This does 3 main things:
1. Switches the usage file to us the yamlv3 library so we can write line comments. We can add to this in the future with header comments for the resources, etc.
2. Updates both the usage file and the resource usage schema to use the same `UsageItem` type. This makes it easier so we don't need to duplicate the hierarchy when we're dealing with complex objects. We still use the simpler `UsageData` type when populating the usage in the resource.
3. Comments out any lines in the usage file when synced and haven't been overwritten. I initially tried doing this with the yamlv3 comment functionality, but it was getting complicated with the edge cases when you have different depths, want to comment out a whole block, etc. Instead I've put in a (sort-of) hack that I think is a lot easier to understand. We mark a YAML node as being commented by adding a prefix to the key and then run post-processing before we write out the YAML to replace those prefixes with a `#`.

### DIFF
--- a/internal/schema/usage_item.go
+++ b/internal/schema/usage_item.go
@@ -10,14 +10,10 @@ const (
 	SubResourceUsage
 )
 
-// type UsageDataValidatorFuncType = func(value interface{}) error
-
 type UsageItem struct {
 	Key          string
 	DefaultValue interface{}
 	Value        interface{}
 	ValueType    UsageVariableType
 	Description  string
-	// These aren't used yet and I'm not entirely sure how they fit in, but they were part of the discussion about usage schema.
-	// ValidatorFunc UsageDataValidatorFuncType
 }


aliscott:
@sinabakh If you get a chance to look over this it would be great to get your feedback and thoughts on this. I'm still in process of adding more usage file golden tests since this is a pretty big refactor.

I merged your changes from https://github.com/infracost/infracost/pull/1011 into here but I haven't made the changes to ignore duplicate misspellings and sort the keys before outputting yet, if you're wanting to that.

alikhajeh1:
This is awesome! I have 3 usability-related notes but none are blockers for this PR though:

1. Could we put the resources that have non-commented lines on top on the first run? seems like it’s sorted alphabetically on the first run and if I re-run, my issue gets fixed (edited) 
Repro steps:
```
rm test1.yml

cd infra/prod

infracost breakdown --path . --sync-usage-file --usage-file test1.yml

cat test1.yml # notice   xxx.aws_dynamodb_table.xxx is mid-file

infracost breakdown --path . --sync-usage-file --usage-file test1.yml

cat test1.yml # notice xxx.aws_dynamodb_table.xxx is at the top
```

2. Can we put a line `##### The following usage values are all commented-out, you can uncomment resources and customize as needed.` so it’s clear all the important stuff is at the top?

3. can we add this to the top of the file:
```
# You can use this file to define resource usage estimates for Infracost to use when calculating
# the cost of usage-based resource, such as AWS S3 or Lambda.
# `infracost breakdown --usage-file infracost-usage.yml [other flags]`
# See https://infracost.io/usage-file/ for docs
```

aliscott:
Merging this so I can rebase other WIP PRs and branches, but if there's any more review comments here I'll address them in another PR. @alikhajeh1 I'll also address your usability suggestions in another PR.

PR #1050, opened 2021-10-14
Add EKS Node Group estimation for instance count

aliscott:


aliscott:
Once #1045 is merged I'll update this to be against master.

PR #1064, opened 2021-10-20
feat: add aws directory service support

### LINKED ISSUE #441: Add support AWS Directory Service
Terraform resources:
- aws_directory_service_conditional_forwarder
- aws_directory_service_directory
- aws_directory_service_log_subscription

AWS pricing page: https://aws.amazon.com/directoryservice/pricing/

Fallowing resources are free of charge:
- aws_directory_service_conditional_forwarder
- aws_directory_service_log_subscription

Currently directory sharing is not possible via terraform so this might be handled by usage (by providing number of accounts).

hugorut:
## Objective:

Add support for aws directory service: Fixes #441

## Pricing details:

Directory service cost output based on:

1. Directory type and size/edition
2. Monthly data transfer out to an external region from the directory
3. Shared directory usage (only applicable to Microsoft AD)

```
 aws_directory_service_directory.microsoft_ad_standard                                         
 └─ Directory service (Microsoft AD, Standard)                     730  hours           $43.80 
                                                                                               
 aws_directory_service_directory.microsoft_ad_with_usage                                       
 ├─ Directory service (Microsoft AD, Standard)                     730  hours           $43.80 
 ├─ Outbound data transfer (from us-east-1, to eu-west-2)        2,000  GB              $40.00 
 └─ Directory sharing                                                8  accounts         $0.14                                                                                      
```

## Status:

- [x] Added to resource_registry.go
- [x] Added internal/resources file
- [x] Added internal/provider/terraform/.../resources file
- [x] Compared test case output to cloud cost calculator.
- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/blob/master/docs/supported_resources.md))

## Issues:

I was unable to find a outbound data transfer pricing scoped to the `AWSDirectoryService` service. So I have used `AWSDataTransfer` service which seems to align with the pricing shown on the pricing page. Let me know if you think this is ok or you manage to find a more specific costing in the pricing db.

--- 

Regarding feedback on the new add resource instructions: I found them useful to get going, but then just dove into code after the fact. I watched the first bit of the video but then dropped off as I had seen previous video with Ali. I think this is a good addition though.  In general I think the docs are pretty good, especially around specific gotchas like updating tests. Things that I think the docs could improve on:

1. I think it could be easier to show in the docs the actual placement of the files with a simple directory tree output. The naming's of the files are so similar I slipped up on first placement.
2. It was a little unclear at first what the mandatory fields on the example resource meant. I initially thought that these were mandatory for all resources but then realised when I got stuck in that these were just fields that were parsed from a provider and thus changed per resource.
3. I think better docs in the code could a good way to help new contributors. I'm going to write an issue for this as I think there's a bit too much to dig into with this PR.

### DIFF
--- a/internal/resources/aws/directory_service_directory.go
+++ b/internal/resources/aws/directory_service_directory.go
@@ -19,12 +19,12 @@ var (
 		{
 			Key:          "additional_domain_controllers",
 			DefaultValue: 0,
-			ValueType:    schema.Int64,
+			ValueType:    schema.Float64,
 		},
 		{
 			Key:          "shared_accounts",
 			DefaultValue: 0,
-			ValueType:    schema.Int64,
+			ValueType:    schema.Float64,
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


alikhajeh1:
> Don't even show show data transfer costs for this resource. It's the same data transfer prices as normal, and for other resources we exclude them since the user can just add them to the aws_data_transfer usage resource. So maybe this is something we need to fix in a more generic way across resources to ensure they are consistent.

+1 on this option, I think data transfer will need more investigation, e.g. can we do lookups from CloudWatch for different resources?

aliscott:
> @aliscott I have resolved all comments, build seems to be failing due to github outage atm: https://www.githubstatus.com/incidents/m78sypncxsdw

👍 thanks! I'll re-run the build when GitHub is back.

PR #1067, opened 2021-10-20
Update contributing docs on adding resources

aliscott:
Updated based on @hugorut's feedback in https://github.com/infracost/infracost/pull/1064.

* Add expected directory tree of file changes
* Update comment for IaC provided attributes


aliscott:
@hugorut would be good if you could take a look at this as well to see if it helps with your feedback.

PR #1088, opened 2021-10-29
Show one line in diff when cost component labels change

aliscott:
I originally did this by separating the cost component name from the labels, but this modified the JSON output. After discussing with @alikhajeh1 we agreed not to change the JSON output format at the moment, so this change parses the labels from the name instead. The one change to the JSON output is we update the name of the diff cost components to 
be show something like `Instance usage (Linux/UNIX, on-demand, t3.small → t3.large)`.

This also updates the output to replace `->` with `→` and `-` with `─` (so the line is continuous).

Before:
```
~ module.base.module.eks.module.node_groups.aws_eks_node_group.workers["default"]
  +$207 ($75.55 -> $283)

    - Instance usage (Linux/UNIX, on-demand, t3.small)
      -$45.55

    + Instance usage (Linux/UNIX, on-demand, t3.large)
      +$243
```

After:
```
~ module.base.module.eks.module.node_groups.aws_eks_node_group.workers["default"]
  +$207 ($75.55 → $283)

    ~ Instance usage (Linux/UNIX, on-demand, t3.small → t3.large)
      +$197 ($45.55 → $243)
```

### DIFF
--- a/internal/schema/diff.go
+++ b/internal/schema/diff.go
@@ -2,7 +2,6 @@ package schema
 
 import (
 	"fmt"
-	"math"
 	"regexp"
 	"strings"
 
@@ -269,7 +268,12 @@ func diffName(current string, past string) string {
 	pastLabels := strings.Split(pastM[2], ", ")
 	currentLabels := strings.Split(currentM[2], ", ")
 
-	labelCount := int(math.Max(float64(len(pastLabels)), float64(len(currentLabels))))
+	// If the names don't have the same label count then return the labels in the format `(old, labels) → (new, labels)`
+	if len(pastLabels) != len(currentLabels) {
+		return fmt.Sprintf("%s (%s) → (%s)", currentM[1], pastM[2], currentM[2])
+	}
+
+	labelCount := len(currentLabels)
 	labels := make([]string, 0, labelCount)
 
 	for i := 0; i < labelCount; i++ {
--- a/internal/schema/diff.go
+++ b/internal/schema/diff.go
@@ -2,10 +2,16 @@ package schema
 
 import (
 	"fmt"
+	"math"
+	"regexp"
+	"strings"
 
 	"github.com/shopspring/decimal"
 	log "github.com/sirupsen/logrus"
 )
+
+// nameBracketReg matches the part of a cost component name before the brackets, and the part in the brackets
+var nameBracketReg = regexp.MustCompile(`(.*?)\s*\((.*?)\)`)
 
 // CalculateDiff calculates the diff of past and current resources
 func calculateDiff(past []*Resource, current []*Resource) []*Resource {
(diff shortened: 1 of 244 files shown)


internal/schema/diff_test.go
```diff
@@ -260,32 +287,115 @@ func TestGetResourcesMap(t *testing.T) {
...
+			current:  "Instance usage (Linux/UNIX, reserved)",
+			past:     "Instance usage (Linux/UNIX, on-demand)",
+			expected: "Instance usage (Linux/UNIX, on-demand → reserved)",
+		},
+		{
+			current:  "Instance usage (Linux/UNIX, reserved, t3.medium)",
+			past:     "Instance usage (Linux/UNIX, on-demand, t3.small)",
+			expected: "Instance usage (Linux/UNIX, on-demand → reserved, t3.small → t3.medium)",
+		},
+	}
```
tim775, line 396: It's probably not actually possible, but it would be good to add a case where the number of labels in the () differ, just to make sure we don't panic if that ever happens.  Maybe something like:
```go
		{
			current:  "Instance usage (Linux/UNIX, reserved, t3.medium)",
			past:     "Instance usage (Linux/UNIX, t3.small)",
			expected: "Instance usage (Linux/UNIX, reserved, t3.medium)",
		},
```

tim775 (approved):

internal/schema/diff.go
```diff
@@ -117,25 +123,39 @@ func diffResourcesByKey(resourceKey string, pastResMap, currentResMap map[string
...
+
+	remainingCostComponents := map[string]*CostComponent{}
+	for _, costComponent := range currentResource.CostComponents {
+		remainingCostComponents[costComponent.Name] = costComponent
+	}
+
+	for _, pastCostComponent := range pastResource.CostComponents {
+		var currentCostComponent *CostComponent
+		if currentResource != nil {
+			currentCostComponent = findMatchingCostComponent(currentResource.CostComponents, pastCostComponent.Name)
```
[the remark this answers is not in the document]
aliscott: 👍 yeah for this case I figured the cost component count is always fairly low so wouldn't make much of a difference.

internal/schema/diff_test.go
aliscott: Good point, added these two test cases:
```go
		{
			current:  "Instance usage (Linux/UNIX, reserved, t3.medium)",
			past:     "Instance usage (Linux/UNIX, on-demand)",
			expected: "Instance usage (Linux/UNIX, on-demand → reserved, t3.medium)",
		},
		{
			current:  "Instance usage (Linux/UNIX, reserved)",
			past:     "Instance usage (Linux/UNIX, on-demand, t3.small)",
			expected: "Instance usage (Linux/UNIX, on-demand → reserved, t3.small)",
		},
```
Although honestly not sure what the expected behaviour should be here though. Maybe we should just return the current value if the label counts don't match?

tim775: I guess the main thing is that it doesn't blow up so I don't think we should go to , but it does seem a little wierd that we're preserving the "t3.small" in the second case.  Maybe if the length changes diff the whole label set like "Instance usage (Linux/UNIX, on-demand, t3.small) ->  (Linux/UNIX, reserved)"?

aliscott: Good idea, have updated to do this.

aliscott:
Merging this and will fix the Azure test in another PR.

PR #1090, opened 2021-10-29
Add support for a post header in ci

fcjr:
Allows setting a header on CI posts -- Useful for when a PR triggers multiple infracost runs.

### DIFF
--- a/internal/config/run_context.go
+++ b/internal/config/run_context.go
@@ -112,6 +112,7 @@ func (c *RunContext) loadInitialContextValues() {
 	c.SetContextValue("os", runtime.GOOS)
 	c.SetContextValue("ciPlatform", ciPlatform())
 	c.SetContextValue("ciScript", ciScript())
+	c.SetContextValue("ciPostHeader", os.Getenv("INFRACOST_CI_POST_HEADER"))
 	c.SetContextValue("ciPostCondition", os.Getenv("INFRACOST_CI_POST_CONDITION"))
 	c.SetContextValue("ciPercentageThreshold", os.Getenv("INFRACOST_CI_PERCENTAGE_THRESHOLD"))
 }
--- a/scripts/ci/atlantis_diff.sh
+++ b/scripts/ci/atlantis_diff.sh
@@ -104,7 +104,11 @@ build_msg () {
     percent_display=" (${change_sym}${percent_display}%%)"
   fi
 
-  msg="##### Infracost estimate #####"
+  msg=""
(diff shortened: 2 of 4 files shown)


alikhajeh1:
@fcjr thank you! Can you plz share some more about the use-case, why do you want that message and what message do you have? I've seen another community user customize the Slack integration with a custom message before the diff text and wondered why.

> Useful for when a PR triggers multiple infracost runs.

Do you have a mono repo setup? @tim775's been thinking about how we can combine runs into 1 PR comment...

fcjr:
> @fcjr thank you! Can you plz share some more about the use-case, why do you want that message and what message do you have? I've seen another community user customize the Slack integration with a custom message before the diff text and wondered why.
> 
> > Useful for when a PR triggers multiple infracost runs.
> 
> Do you have a mono repo setup? @tim775's been thinking about how we can combine runs into 1 PR comment...

Yeah, the use-case is a mono-repo with multiple infras -- sometimes we can combine costs into one config and subsequent ci message (eg the same infra but multiple envs), but some are completely separate.  Occasionally there can be a pr that can affect two different projects (i.e. change to a TF module) this would trigger multiple infracost runs that post to the same PR and you currently have to dig into the dropdown to try and see which is which.  Ideally, there'd be an indication of which job posted which comment.

I'm open to other suggestions to solve this problem if you think something else would fit better into the project, maybe a 'project title' or something similar. =).

alikhajeh1:
@fcjr that makes sense and I think we can use the project path field in the output to automatically do this. Does the following solve your use-case? That could be done by running `jq -r '.projects | map(.name) | join(", ")'` on the Infracost json output but I'm curious if it misses any edge cases in your description?

Screen Shot 2021-10-31 at 11 02 05 PM

aliscott:
@fcjr @alikhajeh1 I'm looking at adding a project breakdown to the PR comment. Would something like this work for this use case?

Screenshot 2021-11-03 at 20 38 03

svix-frank:
> @fcjr @alikhajeh1 I'm looking at adding a project breakdown to the PR comment. Would something like this work for this use case?
> 
> Screenshot 2021-11-03 at 20 38 03

That breakdown is _exactly_ what we are looking for!

aliscott:
@fcjr awesome! We'll ship this in the next version (v0.9.13).

svix-frank:
Sweet thanks @aliscott ill close this out in favor of your new breakdown =)

PR #1105, opened 2021-11-03
feat: add support for module wildcards in usage data

### LINKED ISSUE #874: Enhancement: Add support for wildcard resource usage for module arrays/maps
As of Terraform 0.13 `for_each` and `count` works with modules as well.
Currently we support `count `and `for_each` for modules in price breakdown, however the for usage based resources, some has to express the module resource usages such as:

```
resource_usage:
  module.dev[0].aws_lambda_function.hello_world:
    monthly_requests: 10000
    request_duration_ms: 10
  module.dev[1].aws_lambda_function.hello_world:
    monthly_requests: 20000
    request_duration_ms: 10
```
for resource arrays and

```
resource_usage:
  module.dev["foo"].aws_lambda_function.hello_world:
    monthly_requests: 10000
    request_duration_ms: 10
  module.dev["bar"].aws_lambda_function.hello_world:
    monthly_requests: 10000
    request_duration_ms: 10
```
for resource maps.

Currently it's not possible to define wildcard resource usage for **module** arrays/maps (although it works for **resource** arrays/maps):

```
  module.dev[*].aws_lambda_function.hello_world:
    monthly_requests: 10000
    request_duration_ms: 10
```

sinabakh:
Fix #874

### DIFF
--- a/internal/providers/terraform/parser.go
+++ b/internal/providers/terraform/parser.go
@@ -113,19 +113,67 @@ func (p *Parser) parseJSONResources(parsePrior bool, baseResources []*schema.Res
 
 		if ud := usage[d.Address]; ud != nil {
 			usageData = ud
-		} else if strings.HasSuffix(d.Address, "]") {
-			lastIndexOfOpenBracket := strings.LastIndex(d.Address, "[")
-
-			if arrayUsageData := usage[fmt.Sprintf("%s[*]", d.Address[:lastIndexOfOpenBracket])]; arrayUsageData != nil {
-				usageData = arrayUsageData
+		} else {
+			// Check for array/map indexes.
+			// We should generate all combinations of asterisk and indexes.
+			// So module[0].resource[1] could be any of:
+			//		- module[0].resource[*]
+			//		- module[*].resource[1]
+			//		- module[*].resource[*]
+			stringArrIndexes := findAllCharIndexes(d.Address, '[')
+			combinations := allCombinations(stringArrIndexes)
+			for _, combination := range combinations {
+				newAddress := d.Address
+				for i := len(combination) - 1; i >= 0; i-- {
+					bracketIndex := combination[i]
+					closingBracket := bracketIndex + strings.Index(d.Address[bracketIndex:], "]")
+					newAddress = fmt.Sprintf("%s[*]%s", newAddress[:bracketIndex], newAddress[closingBracket+1:])
+				}
+				if arrayUsageData := usage[newAddress]; arrayUsageData != nil {
+					usageData = arrayUsageData
+					break
+				}
 			}
 		}
+
 		if r := p.createResource(d, usageData); r != nil {
 			resources = append(resources, r)
 		}
 	}
 
 	return resources
+}
+
+func findAllCharIndexes(str string, ch rune) []int {
+	idxs := make([]int, 0)
+	for idx, char := range str {
+		if char == ch {
+			idxs = append(idxs, idx)
+		}
+	}
+	return idxs
+}
+
+func allCombinations(set []int) (subsets [][]int) {
+	length := uint(len(set))
+
+	// Go through all possible combinations of objects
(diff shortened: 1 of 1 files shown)


aliscott:
Thanks @sinabakh! How does this handle multiple matches? I'm thinking we should match in this priority order. What do you think?

1. `module.my_module[0].aws_instance.my_instance[1]`
2. `module.my_module[0].aws_instance.my_instance[*]`
3. `module.my_module[*].aws_instance.my_instance[1]`
4. `module.my_module[*].aws_instance.my_instance[*]`

sinabakh:
@aliscott I agree, I think the code probably works as you mentioned but I will make sure that the order is as we expect it to be.
Currently, I'm puzzled by syncing usage data stuff. I hope I can make it for the next release.

sinabakh:
@aliscott @alikhajeh1 I think this PR is probably ready but I'm not sure about all edge cases and expected behaviors especially when syncing usage files. So I would be more than happy if you would do some tests if you have time to do so.

Regarding @aliscott comment, currently it does the following order:
```

    module.my_module[0].aws_instance.my_instance[1]
    module.my_module[*].aws_instance.my_instance[0]
    module.my_module[0].aws_instance.my_instance[*]
    module.my_module[*].aws_instance.my_instance[*]

```
So 3 is before 2. However, is it valid for a yaml file to contain 2 and 3 at the same time?

aliscott:
Thanks @sinabakh! I'll give this a shot with the sync usage stuff.

> So 3 is before 2. However, is it valid for a yaml file to contain 2 and 3 at the same time?

Yes I think this should be valid, although it's unlikely, and to me it makes sense that `module.my_module[0].aws_instance.my_instance[*]` should have precedence over `module.my_module[0].aws_instance.my_instance[*]`, since then the priority is always `module overrides > submodule overrides > resource overrides`. What do you think?

alikhajeh1:
Can we please review https://www.infracost.io/docs/usage_based_resources#resource-arraysmaps as part of this PR and see if it needs to be updated (or new examples added).

aliscott:
@sinabakh, thanks I've given this a quick shot and here's some initial comments.

1. Generally, I think this could benefit from some golden tests. If it helps, here's an example Terraform repo I've been testing this with: https://gist.github.com/aliscott/6735d70f2da69f33cc5903e5e768e01f

2. For wildcard resources, we still write the existing wildcard rows when we call `--sync-usage-file`, but we don't exclude any new rows from being added. I'm not sure if this functionality will change in the future, but I think we should do a similar thing for modules at the moment. We currently do this for resources here: https://github.com/infracost/infracost/blob/master/internal/usage/sync.go#L140

aliscott (commented):
@sinabakh, thanks I've given this a quick shot and here's some initial comments.

1. Generally, I think this could benefit from some golden tests. If it helps, here's an example Terraform repo I've been testing this with: https://gist.github.com/aliscott/6735d70f2da69f33cc5903e5e768e01f

2. For wildcard resources, we still write the existing wildcard rows when we call `--sync-usage-file`, but we don't exclude any new rows from being added. I'm not sure if this functionality will change in the future, but I think we should do a similar thing for modules at the moment. We currently do this for resources here: https://github.com/infracost/infracost/blob/master/internal/usage/sync.go#L140

alikhajeh1:
What's the next step for this PR? I'd be happy to help with testing and docs

sinabakh:
@alikhajeh1 It needs a few tweaks in syncing the usage file phase. However, I'm currently working on auto-migration stuff whenever I find the time.

aliscott:
@sinabakh @alikhajeh1 now that we support [resource defaults in the usage file](https://www.infracost.io/docs/features/usage_based_resources/#usage-profiles), do you think we still need this functionality?

alikhajeh1:
I vote no just now, we can re-open the PR if we hear users request more granular wildcards...

sinabakh:
I agree that we can de-prioritize this feature.

PR #1106, opened 2021-11-04
Add project list to top of PR comments

aliscott:
This adds a project table to the top of the PR comments for GitHub, GitLab and Azure Devops.

Screenshot 2021-11-04 at 11 34 47

It has the following functionality:
* If there's only one project don't show the "All projects" line
* Doesn't show projects with no changes, instead it lists them below the table
* Truncates the middle of project names to limit them to 64 characters

This also updates the atlantis and jenkins diff scripts to keep them more inline with our standard diff script.
Once this is merged we can update the screenshots in the READMEs and docs.

See https://github.com/infracost/infracost/pull/1090 for previous work.

### DIFF
--- a/scripts/ci/diff.sh
+++ b/scripts/ci/diff.sh
@@ -49,7 +49,7 @@ process_args () {
   if [ -n "$GIT_SSH_KEY" ]; then
     echo "Setting up private Git SSH key so terraform can access your private modules."
     mkdir -p .ssh
-    echo "${GIT_SSH_KEY}" > .ssh/git_ssh_key
+    echo "$GIT_SSH_KEY" > .ssh/git_ssh_key
     chmod 600 .ssh/git_ssh_key
     export GIT_SSH_COMMAND="ssh -i $(pwd)/.ssh/git_ssh_key -o 'StrictHostKeyChecking=no'"
   fi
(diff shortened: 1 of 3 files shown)


scripts/ci/diff.sh
```diff
@@ -86,18 +86,150 @@ build_breakdown_cmd () {
...
+
+  percent_display=$(percent_display "$past_total_monthly_cost" "$total_monthly_cost")
+  change_word=$(change_word "$past_total_monthly_cost" "$total_monthly_cost")
+  change_emoji=$(change_emoji "$past_total_monthly_cost" "$total_monthly_cost")
+
+  msg="$MSG_START **monthly cost will $change_word by $(format_cost "$diff_total_monthly_cost")$percent_display** $change_emoji\n"
+  msg+="\n"
+
+  if [ "$include_html" = true ]; then
+    msg+="<table>\n"
```
tim775, line 115: Just curious: Any reason you did this with html and not markdown tables `| a | b | c |`?

tim775 (approved):

scripts/ci/diff.sh
aliscott: The only reason was @alikhajeh1 suggested not having bold table headers so it wouldn't detract from the bold "monthly cost will increase..." message. Markdown tables always render the first row cells as `<th>` so they are always bold.

PR #1111, opened 2021-11-08
Feat/output markdown

tim775:
This ports the markdown comment generated by `scripts/ci/diff.sh` to the `infracost output` as new formats `markdown` and `markdown-html` (let me know if you have a better suggestions for how to support the "IncludesHTML" option).


aliscott:
> let me know if you have a better suggestions for how to support the "IncludesHTML" option

I think this is good. The other option I can think of is having format values for each CI system:
```
--format github-comment
--format gitlab-comment
--format bitbucket-comment
--format azure-devops-comment
--format slack-message
```

The advantages are:
1. In the future we can utilise any unique features of these, for example: colours are supported in GitLab, but not in GitHub.
2. It might be more self-documenting for users - we wouldn't need to maintain a separate list of what format to use with each CI system

cc'ing @alikhajeh1 as well if he has thoughts on any options.

alikhajeh1:
+1 on VCS-specific formats!

internal/output/combined.go
```diff
@@ -47,13 +51,46 @@ func Combine(currency string, inputs []ReportInput, opts Options) Root {
...
+		}
+
 	}
 
 	combined.Version = outputVersion
 	combined.Currency = currency
 	combined.Projects = projects
 	combined.TotalHourlyCost = totalHourlyCost
 	combined.TotalMonthlyCost = totalMonthlyCost
+	combined.PastTotalHourlyCost = pastTotalHourlyCost
```
aliscott, line 90: Thanks for fixing this!

aliscott (approved):
LGTM

PR #1116, opened 2021-11-09
`--out-file` flag

aliscott:
This adds an `--out-file` flag to `infracost diff`, `infracost breakdown` and `infracost output`.

I've split the golden file test updates into separate commits to help the review. One change here is that we used to print a newline to stderr preceding table and diff outputs to add space between them and the spinners. I've changed this to be a newline on stderr so stdout doesn't start with a blank newline.


aliscott:
@alikhajeh1 can you try this as well to check what you think of the help text and output messaging when we save the file? (docs PR is https://github.com/infracost/docs/pull/115)

alikhajeh1 (commented):
I was expecting things to only output to file when that's used, as otherwise I would use the CLI without that option. Doing that should also avoid this issue where the "Output saved to out.json" is shown after the start of the JSON.

Screen Shot 2021-11-09 at 3 09 16 PM
n

tim775 (approved):

PR #1120, opened 2021-11-09
Fix/output GitHub comment

tim775:
The first commit updates the golenfile to capture test cases we're interested in.  Subsequent commits should be a single change + the updated golden files, so it's probably easiest to review each commit in sequence.


aliscott (approved):
LGTM, the commits made it nice to review. I've tested this with the following cases:
1. two projects, both have diffs
2. two projects, 1 has diff
3. two projects, none have diffs
4. one project, has diff
5. one project, does not have diff

PR #1028, opened 2021-10-07
Feat/add aws mwaa environment

tim775:
## Objective:

Add support for aws_mwaa_environment.  Fixes #895

## Pricing details:

The aws_mwaa_environment has 4 price components described in https://aws.amazon.com/managed-workflows-for-apache-airflow/pricing/.  Additional workers/additional schedulers seem to be an autoscaling feature so those are handled as a usage param.  

## Status:

- [x] Added to resource_registry.go
- [x] Added internal/resources file
- [x] Added internal/provider/terraform/.../resources file
- [x] Added usage parameters to infracost-usage-example.yml
- [x] Added test cases without usage-file
- [x] Added test cases with usage-file
- [x] Compared test case output to cloud cost calculator.
- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/blob/master/docs/supported_resources.md))

## Issues:

None

PR #1123, opened 2021-11-10
refactor: hardcode resource dependencies in testdata

PatMyron:
https://github.com/infracost/infracost/pull/1028#discussion_r746980579

Just a proof of concept for a few resource types
There are still more resource dependencies throughout `testdata` that could be hardcoded as well if this idea seems good


tim775:
This is great, it makes the tf much more readable.  I love all the red.   I was just cargo-culting so I think the more we can do this the better!

PR #1124, opened 2021-11-11
fix(azure): handle lower-case standard throughput

alikhajeh1:


internal/providers/terraform/azure/event_hubs_namespace.go
```diff
@@ -113,7 +113,7 @@ func eventHubsThroughPutCostComponent(region, sku, meterName string, capacity de
 			Service:    strPtr("Event Hubs"),
 			AttributeFilters: []*schema.AttributeFilter{
 				{Key: "skuName", ValueRegex: strPtr(fmt.Sprintf("/%s/i", sku))},
-				{Key: "meterName", Value: strPtr(meterName)},
+				{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/%s/i", meterName))},
```
vdmgolub, line 116: Should we enforce the start and the end of the string? Or is it a substring?
```suggestion
				{Key: "meterName", ValueRegex: strPtr(fmt.Sprintf("/^%s$/i", meterName))},
```

alikhajeh1: Good point! Whilst I prefer strict checks, so far we've been using substrings mostly as that finds what we want and it catches tweaks that the cloud vendor might make when they add variations.

`--log-level warning` shows "multiple prices matches" (if we use a substring) or "no prices matched" (if we use start/end of string) and I think both get caught in our automated tests so we'll know if this needs changing regardless of what we use.

vdmgolub (approved):

PR #1131, opened 2021-11-15
feat(azure): add resources for Azure Front Door service

### LINKED ISSUE #909: Add Support for Azure Frontdoor
Terraform Resource: 

`azurerm_frontdoor`

Pricing Information:

https://azure.microsoft.com/en-us/pricing/details/frontdoor/

vdmgolub:
## Objective:

Add support for `azurerm_frontdoor` and `azurerm_frontdoor_firewall_policy`.  Fixes #909

## Pricing details:

- Azure has two Front Door services: "Standard" and newer "Standard/Premium (Preview)". This PR introduces the "Standard" one.
- Pricing for outbound data transfer depends on the region, in the Pricing API the regions are represented as zones (Zone 1 - Zone 5, US Gov Zone 1).
- Pricing API has 2 prices for Routing rules quantity. The item with `startUsage: 0` is free (for first 5 rules), however Azure's pricing page bills all the rules. Using the price with `startUsage: 5` instead.

## Status:

- [x] Added to resource_registry.go
- [x] Added internal/resources file
- [x] Added internal/provider/terraform/.../resources file
- [x] Compared test case output to cloud cost calculator.
- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/pull/119)

## Issues:

None

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -739,6 +739,20 @@ resource_usage:
     throughput_or_capacity_units: 10  # Number of Throughput Units (for Basic and Standard) and Capacity units (for Dedicated) namespaces.
     capture_enabled: false            # Defines if capture is enabled for the Event Hub Standard namespaces, can be: true, false.
 
+  azurerm_frontdoor.my_frontdoor:
+    monthly_outbound_data_transfer_gb: # Monthly outbound data transfer from the following, in GB:
+      us_gov: 190000                       # US Gov
+      north_america_europe_africa: 200000  # North America, Europe and Africa
+      asia_pacific: 220000                 # Asia Pacific (including Japan)
+      south_america: 10000                 # South America
+      australia: 50000                     # Australia
+      india: 387000                        # India
+    monthly_inbound_data_transfer_gb: 1000 # Monthly inbound data transfer in GB
+
+  azurerm_frontdoor_firewall_policy.my_frontdoor_firewall_policy:
+    monthly_custom_rule_requests: 11000     # Monthly number of custom rule requests
+    monthly_managed_ruleset_requests: 10000 # Monthly number of managed ruleset requests
+
   azurerm_kubernetes_cluster.my_cluster:
     load_balancer:
       monthly_data_processed_gb: 100 # Monthly inbound and outbound data processed in GB.
--- a/internal/providers/terraform/azure/frontdoor.go
+++ b/internal/providers/terraform/azure/frontdoor.go
@@ -1,0 +1,49 @@
+package azure
+
+import (
+	"strings"
+
+	"github.com/infracost/infracost/internal/resources/azure"
+	"github.com/infracost/infracost/internal/schema"
+	"github.com/tidwall/gjson"
+)
+
+// getAzureRMFrontdoorRegistryItem returns a registry item for the resource
+func getAzureRMFrontdoorRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_frontdoor",
+		RFunc: newFrontdoor,
+		ReferenceAttributes: []string{
+			"resource_group_name",
+		},
+	}
+}
+
+// newFrontdoor parses Terraform's data and uses it to build a new resource
+func newFrontdoor(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := lookupRegion(d, []string{"resource_group_name"})
+
+	if strings.HasPrefix(strings.ToLower(region), "usgov") {
+		region = "US Gov Zone 1"
+	} else {
+		region = regionToZone(region)
+	}
+
+	rulesCounter := 0
+	rules := d.Get("routing_rule").Array()
+	for _, rule := range rules {
+		if rule.Get("enabled").Type == gjson.True {
+			rulesCounter++
+		}
+	}
+
+	r := &azure.Frontdoor{
+		Address:       d.Address,
+		Region:        region,
+		FrontendHosts: len(d.Get("frontend_endpoint").Array()),
+		RoutingRules:  rulesCounter,
+	}
+	r.PopulateUsage(u)
+
+	return r.BuildResource()
+}
--- a/internal/providers/terraform/azure/frontdoor_firewall_policy.go
+++ b/internal/providers/terraform/azure/frontdoor_firewall_policy.go
@@ -1,0 +1,52 @@
+package azure
+
+import (
+	"strings"
+
+	"github.com/infracost/infracost/internal/resources/azure"
+	"github.com/infracost/infracost/internal/schema"
+)
+
+// getAzureRMFrontdoorFirewallPolicyRegistryItem returns a registry item for the
+// resource
+func getAzureRMFrontdoorFirewallPolicyRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_frontdoor_firewall_policy",
+		RFunc: newFrontdoorFirewallPolicy,
+		ReferenceAttributes: []string{
+			"resource_group_name",
+		},
+	}
+}
+
+// newFrontdoorFirewallPolicy parses Terraform's data and uses it to build
+// a new resource
+func newFrontdoorFirewallPolicy(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := lookupRegion(d, []string{"resource_group_name"})
+
+	if strings.HasPrefix(strings.ToLower(region), "usgov") {
+		region = "US Gov Zone 1"
+	} else {
+		region = regionToZone(region)
+	}
+
+	customRules := 0
+	if rules := d.Get("custom_rule"); rules.Exists() {
+		customRules = len(rules.Array())
+	}
+
+	managedRulesets := 0
+	if rules := d.Get("managed_rule"); rules.Exists() {
+		managedRulesets = len(rules.Array())
+	}
+
+	r := &azure.FrontdoorFirewallPolicy{
+		Address:         d.Address,
+		Region:          region,
+		CustomRules:     customRules,
+		ManagedRulesets: managedRulesets,
+	}
+	r.PopulateUsage(u)
+
+	return r.BuildResource()
+}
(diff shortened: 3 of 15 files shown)


infracost-usage-example.yml
```diff
@@ -739,6 +739,19 @@ resource_usage:
     throughput_or_capacity_units: 10  # Number of Throughput Units (for Basic and Standard) and Capacity units (for Dedicated) namespaces.
     capture_enabled: false            # Defines if capture is enabled for the Event Hub Standard namespaces, can be: true, false.
 
+  azurerm_frontdoor.my_frontdoor:
+    monthly_outbound_data_transfer_gb: # Monthly outbound data transfer from the following, in GB:
+      north_america_europe: 200000         # North America, Europe and Africa
```
vdmgolub, line 744: The Pricing Calculator calls the Zone1 "North America, Europe and Africa". Using `north_america_europe` for a shorte name.

internal/providers/terraform/azure/frontdoor_firewall_policy.go
```diff
@@ -0,0 +1,45 @@
...
+			"resource_group_name",
+		},
+	}
+}
+
+// newFrontdoorFirewallPolicy parses Terraform's data and uses it to build
+// a new resource
+func newFrontdoorFirewallPolicy(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := lookupRegion(d, []string{"resource_group_name"})
+	region = regionToZone(region)
```
vdmgolub, line 24: Should we include check for "US Gov 1" region (exists in Pricing API), if TF doesn't list it as an available region?

internal/providers/terraform/azure/testdata/frontdoor_test/frontdoor_test.golden
```diff
@@ -0,0 +1,25 @@
...
+                                                                                                  
+ azurerm_frontdoor.frontdoor_with_usage                                                           
+ ├─ Routing rules (first 5 rules)                             3,650  hours                $109.50 
+ ├─ Routing rules (per additional rule)                         730  hours                  $8.76 
+ ├─ Frontend hosts (over 100)                                    10  hosts                 $50.00 
+ ├─ Inbound data transfer                                     1,000  GB                    $10.00 
+ └─ Outbound data transfer                                                                        
+    ├─ North America, Europe and Africa (first 10TB)         10,000  GB                 $1,700.00 
+    ├─ North America, Europe and Africa (next 40TB)          40,000  GB                 $6,000.00 
+    └─ North America, Europe and Africa (next 100TB)        150,000  GB                $19,500.00
```
vdmgolub, line 12: FYI The issue lists `Outbound data transfer` as a first component, but subresources are pushed to the bottom.
Also, Pricing API includes pricing for "over 150TB" tier which the same as "next 100TB". Should we include an additional tier here?

internal/providers/terraform/azure/frontdoor_firewall_policy.go
alikhajeh1: For Azure, I think TF lists the regions that are supported by the account (meaning if the user's creds supports gov cloud regions, they'll see them). Search the codebase for `US Gov` in case some of those example can be used here, e.g. `dnsQueriesCostComponent`

vdmgolub: Oh, good to know. I'll add the check then. Thank you!

internal/providers/terraform/azure/testdata/frontdoor_firewall_policy_test/frontdoor_firewall_policy_test.tf
```diff
@@ -0,0 +1,95 @@
...
+    match_condition {
+      match_variable = "RemoteAddr"
+      operator       = "IPMatch"
+      match_values   = ["192.168.1.0/24", "10.0.0.0/24"]
+    }
+  }
+
+  custom_rule {
+    name     = "Rule2"
+    enabled  = false
```
vdmgolub, line 74: I can't find the information if disabled custom rules are billed or not. As example we don't count the disabled tests for [azurerm_application_insights_web_test](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/application_insights_web_test), [infracost resource](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/azure/application_insights_web_t.go#L23).

infracost-usage-example.yml
aliscott: I would go with `north_america_europe_africa` or `zone_1`. @alikhajeh1 any opinions?

internal/providers/terraform/azure/testdata/frontdoor_firewall_policy_test/frontdoor_firewall_policy_test.tf
aliscott: This took a while to hunt down: https://docs.microsoft.com/en-us/azure/frontdoor/front-door-faq#will-i-be-billed-for-the-azure-front-door-resources-that-are-disabled-

> Azure Front Door resources, like Front Door profiles, routing rules are not billed in disabled. WAF policies and rules are billed even if disabled.

So the `azurerm_frontdoor_firewall_policy` rules should still have a cost if disabled, but the `azurerm_frontdoor` rules shouldn't.

internal/providers/terraform/azure/testdata/frontdoor_test/frontdoor_test.golden
aliscott: The pricing calculator uses the same price for >150 as well, but the pricing page says "Contact Us". I think we have two options here.
1. Have the last tier as `over 50TB` using the $0.13 price.
2. Have the `next 100TB` tier and a `over 150TB` tier that just use the same price.

Any thoughts @alikhajeh1?

internal/resources/azure/frontdoor.go
```diff
@@ -0,0 +1,324 @@
...
+
+// Frontdoor struct represents Azure's Front Door network service.
+//
+// More resource information here: https://docs.microsoft.com/en-us/azure/frontdoor/front-door-overview
+// Pricing information here: https://azure.microsoft.com/en-us/pricing/details/frontdoor/ (Azure Front Door tab)
+type Frontdoor struct {
+	Address string
+	Region  string
+
+	FrontendHostsQuantity int
```
aliscott, line 21: We don't add `Quantity` suffixes elsewhere. I'm open to changing this, just thought I'd flag up this inconsistency.

internal/providers/terraform/azure/frontdoor.go
```diff
@@ -0,0 +1,40 @@
...
+		region = "US Gov Zone 1"
+	} else {
+		region = regionToZone(region)
+	}
+
+	r := &azure.Frontdoor{
+		Address:               d.Address,
+		Region:                region,
+		FrontendHostsQuantity: len(d.Get("frontend_endpoint").Array()),
+		RoutingRulesQuantity:  len(d.Get("routing_rule").Array()),
```
aliscott, line 35: As per previous comment, I think we should only count enabled `routing_rule` blocks for this resource?

internal/resources/azure/util.go
```diff
@@ -0,0 +1,20 @@
+package azure
+
+import (
+	"github.com/shopspring/decimal"
+)
+
+func strPtr(s string) *string {
```
aliscott, line 7: Off topic for this PR, but since we use these everywhere what are your thoughts about moving these to their own package longer term? I've been hesitant because of the general golang advice of avoiding `util` packages.

internal/providers/terraform/azure/testdata/frontdoor_firewall_policy_test/frontdoor_firewall_policy_test.tf
vdmgolub: Thank you! Your documentation kung-fu is strong :D

infracost-usage-example.yml
vdmgolub: I guess for the usage file `north_america_europe_africa` is more user-friendly. Although it doesn't include ALL regions in these names. Maybe we could also add a comment with a link to calculator, so if not sure users will have a quick way to find it?

internal/providers/terraform/azure/testdata/frontdoor_test/frontdoor_test.golden
alikhajeh1: I vote for Option 1 as the golden file tests will catch if/when prices change for the last tier. So for now we can keep it simple and adjust later if needed.

infracost-usage-example.yml
alikhajeh1: +1 on `north_america_europe_africa` as it's self-documenting.

internal/resources/azure/frontdoor.go
vdmgolub: For context: I wasn't sure if it's better to pass the whole array and calculate the length here. But decided that this struct should be aware of information related to pricing only which is quantity in this case. So went with this name to distinguish array and int values. I've also seen some examples using `numberOfX`. But with Go's types it should be simple to recognize it I guess. Will change to `FrontendHosts`/`RoutingRules` style. 👍

internal/resources/azure/util.go
vdmgolub: I think it okay sometimes to bend a rule/good practice in favor of simplicity/usability when it makes sense. But I'm not very familiar with Go way to make such suggestion yet :D

vdmgolub:
@aliscott Thank you! All comments are addressed. I've put each change into a separate fixup commit for easier review. Please let me know when I can rebase. I'd like to keep the [test fix](https://github.com/infracost/infracost/pull/1131/commits/f4d767aed624cc2d9af2f3e8b8ce8d3a16e02659) commit separate from new resources, so we can have 2 commits in the end.

internal/providers/terraform/azure/registry.go
```diff
@@ -214,6 +216,9 @@ var FreeResources = []string{
 	"azurerm_firewall_policy",
 	"azurerm_firewall_policy_rule_collection_group",
 
+	// Azure Front Door
+	"azurerm_frontdoor_custom_https_configuration",
```
aliscott, line 220: Looks like Hashicorp have added support for a new resource `azurerm_frontdoor_rules_engine` since we broke this one down. I can't find any charges for it and https://www.neowin.net/news/azure-front-door-and-cdn-rules-engine-now-generally-available/ seems to suggest it's free, so I think we should add it here and to the docs.

aliscott (approved):
Changes look great. One small note about a new resource Hashicorp has added.

aliscott (approved):

PR #1133, opened 2021-11-16
Fix AWS/Azure tests to unblock CI

vdmgolub:


internal/providers/terraform/aws/testdata/data_transfer_test/data_transfer_test.golden
```diff
@@ -197,4 +197,16 @@
  OVERALL TOTAL                                                           $325,124.38 
 ──────────────────────────────────
 24 cloud resources were detected:
-∙ 24 were estimated, 24 include usage-based costs, see https://infracost.io/usage-file
\ No newline at end of file
+∙ 24 were estimated, 24 include usage-based costs, see https://infracost.io/usage-file
+Logs:
+
+level=warning msg="Multiple products found for aws_data_transfer.ap-southeast-1 Outbound data transfer to Internet (first 10TB), using the first product"
```
alikhajeh1, line 203: Seems odd that there are multiple matches, we don't need to update price lookup filters?

vdmgolub: Sorry, I've the explanation of the fix into the commit, but forgot to add it to the PR description. Hope this clarifies it:

```The issue is caused by ap-southeast-1 region:
- 2 products for Outbound data transfer to Internet
- First product has APS1 in `usagetype`. It has correct pricing
- Second product is APS4 and pricing differs
- The test picks the correct pricing. Modifying this test to unblock CI.
- Will report to AWS to investigate the root cause```

The solution is to introduce an additional filter checking `APS`/`APS4`, but it seems odd because APS4 doesn't exist.
We discussed it with Alistair and decided to add temporary fix. we'll decide what to do next when we hear back from AWS.

alikhajeh1: Makes sense, looking at https://docs.aws.amazon.com/AmazonS3/latest/userguide/aws-usage-report-understand.html I wonder if APS4 is a new region that'll be released soon? I [tweeted](https://twitter.com/AliKhajeh/status/1460672144652877825) to see if anyone knows.

PR #1135, opened 2021-11-16
fix(azure): param can include an empty string in TF plan JSON

alikhajeh1:
The empty string wasn’t being checked before thus the resource could be estimated as a dedicated cluster, which is a lot more expensive. 

Big thanks to @fernandomatsuosantos for helping us debug this. It was hard to track down as two identical projects were showing different cost estimates.

---

There's no easy way to write a test for this resource combo as the same TF version was producing slightly different plan JSONs (maybe the azure provider version was different). So I'm wondering if we should create an issue to review other places we use `d.Get("...").Type != gjson.Null`. Maybe we should have a helper function that can be used as there are 3 cases we need to handle when processing params in TF plan JSONs: param is not present, param is present but it's set to `null`, param is present and set to `""`. We switched from `.Exists()` to `d.Get("value_name").Type != gjson.Null` but that's still not enough.

### DIFF
--- a/internal/providers/terraform/azure/event_hubs_namespace.go
+++ b/internal/providers/terraform/azure/event_hubs_namespace.go
@@ -34,7 +34,7 @@ func NewAzureRMEventHubs(d *schema.ResourceData, u *schema.UsageData) *schema.Re
 		capacity = decimal.NewFromInt(d.Get("capacity").Int())
 	}
 
-	if d.Get("dedicated_cluster_id").Type != gjson.Null {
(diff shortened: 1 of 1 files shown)


vdmgolub:
`UsageData` [already has](https://github.com/infracost/infracost/blob/master/internal/schema/usage_data.go#L76) `IsEmpty()` function exactly for such check:
```go
// Return true if the key doesn't exist, is null, or is an empty string.
// Needed because gjson.Exists returns true as long as a key exists, even if it's empty or null.
func (u *UsageData) IsEmpty(key string) bool {
	g := u.Get(key)
}
```

We probably can extract it and reuse wherever needed.

vdmgolub (approved):

PR #1141, opened 2021-11-19
chore(aws): fix double products for aws_data_transfer resource

vdmgolub:
The previous fix to the golden file tried to resolve double products
found for `ap-southeast-1` region. First product has `APS1` prefix (correct)
and the second one has `APS4` (incorrect one). The fix worked until
the product results order returned `APS1` records first. But it was not
guaranteed. Mostly like it's an issue on AWS side.
Introducing an additional filter with the prefix should match exactly
the correct product. It might not work for regions that don't have such
prefix in `usagetype`.

### DIFF
--- a/internal/resources/aws/data_transfer.go
+++ b/internal/resources/aws/data_transfer.go
@@ -1,0 +1,270 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/shopspring/decimal"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/infracost/infracost/internal/resources"
+	"github.com/infracost/infracost/internal/schema"
+	"github.com/infracost/infracost/internal/usage"
+)
+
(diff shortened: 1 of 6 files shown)


internal/resources/aws/data_transfer.go
```diff
@@ -0,0 +1,278 @@
...
+		})
+	}
+
+	if regionCode, ok := RegionCodeMapping[r.Region]; ok {
+		fmt.Println()
+		attributeFilters = append(attributeFilters, &schema.AttributeFilter{
+			Key:        "usagetype",
+			ValueRegex: strPtr(fmt.Sprintf("/^%s-/i", regionCode)),
+		})
+	}
```
vdmgolub, line 270: This is the fix. As `RegionCodeMapping` includes only one key for now, it won't affect other regions and it could be reused in future if needed.

aliscott (approved):
LGTM. Thanks for restructuring the resource at the same time!

PR #1146, opened 2021-11-22
Strip output added by setup-terraform wrapper

aliscott:


tim775 (approved):

alikhajeh1:
This is great - thanks!

internal/providers/detect.go
```diff
@@ -64,6 +65,11 @@ func isTerraformPlanJSON(path string) bool {
 		PlannedValues interface{} `json:"planned_values"`
 	}
 
+	b, hasWrapper := terraform.StripSetupTerraformWrapper(b)
+	if hasWrapper {
+		log.Debugf("Stripped setup-terraform wrapper output from %s", path)
```
aliscott, line 70: Unsure if we should debug or warn here.

alikhajeh1: One way to think of it would be: what's the default we use in CI? do we want users to see this log by default? (I lean towards less as it feels like the sort of thing that could be tricky to debug. Should the log say why it's doing this too?

aliscott: We show info level by default, so could log it at that level. The other option would be warn, but I'm not sure if it needs that.

alikhajeh1: +1 on info, it doesn't do any harm and if the user is using Infracost alongside other tools they might be wondering why it works with Infracost but not the other tools

aliscott: How's this for messaging?
```
Stripped wrapper output from plan.json since setup-terraform GitHub Action was used without `terraform_wrapper: false`
```

alikhajeh1: Stripped wrapper output from plan.json (to make it a valid JSON file) since setup-terraform GitHub Action was used without `terraform_wrapper: false`

PR #1154, opened 2021-11-27
Add additional_backup_storage_gb usage based cost component to aws_db_instance resource

### LINKED ISSUE #1063: Add additional backup storage for AWS RDS
Add a new usage-based cost component to the `aws_db_instance` resource:
* Name: Additional backup storage
* Unit: GB
* Usage key: `additional_backup_storage_gb`

AWS Pricing pages:
* https://aws.amazon.com/rds/aurora/pricing/?pg=pr&loc=1
* https://aws.amazon.com/rds/postgresql/pricing/?pg=pr&loc=3
* https://aws.amazon.com/rds/mysql/pricing/?pg=pr&loc=2
* https://aws.amazon.com/rds/mariadb/pricing/?pg=pr&loc=4
* https://aws.amazon.com/rds/oracle/pricing/?pg=pr&loc=5
* https://aws.amazon.com/rds/sqlserver/pricing/?pg=pr&loc=6

It seems to be a set $0.095 per GB for all DB types except for Aurora DBs where there are different charges per region.

joshniec:
Closes #1063 

This PR implements the `additional_backup_storage_gb` usage based cost component for `aws_db_instance`. 

Some oddities I came across when I was querying a local cloud-pricing-api docker container I spun up was that there were a few "Custom" type AWS RDS instances, for example:

```
cloud_pricing=# cloud_pricing=# select attributes from products where "vendorName" = 'aws' and "service" = 'AmazonRDS' and "productFamily" = 'Storage Snapshot' and "region" <> '' and "attributes"->>'usagetype' like '%BackupUsage%' and "region" = 'us-east-1' and "attributes"->>'databaseEngine' = 'Any';
--------------------
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDS:ChargedBackupUsage", "engineCode": "0", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "storageMedia": "AmazonS3", "databaseEngine": "Any"}
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDSCustom:ChargedBackupUsage", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "databaseEngine": "Any", "databaseEdition": "Any", "deploymentModel": "Custom"}
```

This appears to be this product: https://aws.amazon.com/rds/custom/. There is no native support for it in terraform, however, in the attributes column, the main differentiator between a custom instance and a non-custom instance was the fact that the `engineCode` attribute existed on non-custom instances. Therefore I opted check that the `engineCode` attribute exists, which would filter out all of the "Custom" RDS type pricing. 

Another was the `databaseEngine` attribute, which there is only three of for this service's pricing: `Any`, `Aurora MySQL`, and `Aurora PostgreSQL`. You can see I handled this by creating a new variable `backupStorageDbEngine` set to `Any` by default, checking if `*databaseEngine` has a prefix of `Aurora`, if so, set `backupStorageDbEngine` equal to `databaseEngine`.


internal/providers/terraform/aws/db_instance.go
```diff
@@ -192,6 +192,34 @@ func NewDBInstance(d *schema.ResourceData, u *schema.UsageData) *schema.Resource
 		})
 	}
 
+	var backupStorageGB *decimal.Decimal
```
joshniec, line 195: Should I wrap this in a `if d.Get("backup_retention_period).Exists() { }` to check that `backup_retention_period` is configured on the `aws_db_instance` resource?

alikhajeh1: Good idea! Maybe we can use the following condition so we also show the cost component if a user has explicitly set the usage param (e.g. to do what-if analysis before even enabling the Terraform param).

```
if d.Get("backup_retention_period).Exists() || (u != nil && u.Get("additional_backup_storage_gb").Exists())
```

---

This isn't required in this PR, but we have an [`IsEmpty` func](https://github.com/infracost/infracost/pull/1135#issuecomment-970499598) that we can extract and re-use instead of `Exists()` (as it catches some rare conditions we've seen in TF plan JSONs). @vdmgolub any idea where we should put this function so it can be re-used for `d.` and `u.` checks?

vdmgolub: @alikhajeh1 @joshniec Both `ResourceData` (d, [link](https://github.com/infracost/infracost/blob/master/internal/schema/resource_data.go#L49-L54)) and `UsageData` (u, [link](https://github.com/infracost/infracost/blob/master/internal/schema/usage_data.go#L74-L79)) structs have this `IsEmpty(key)` function (identical code in both places). I guess that'll do for now.

joshniec: Thanks, I went ahead and refactored my code to make use of the `IsEmpty(key)` function. I also replaced other calls of `d.Get(key).Exists()` and `u.Get(key).Exists()` to the `IsEmpty(key)` function in the same file (which let me remove the `"github.com/tidwall/gjson"` import).

internal/providers/terraform/aws/db_instance.go
```diff
@@ -192,6 +191,36 @@ func NewDBInstance(d *schema.ResourceData, u *schema.UsageData) *schema.Resource
...
 	}
 
+	if !d.IsEmpty("backup_retention_period") || (u != nil && !u.IsEmpty("additional_backup_storage_gb")) {
+		var backupStorageGB *decimal.Decimal
+		if u != nil && !u.IsEmpty("additional_backup_storage_gb") {
+			backupStorageGB = decimalPtr(decimal.NewFromInt(u.Get("additional_backup_storage_gb").Int()))
+		}
+
+		backupStorageDbEngine := "Any"
+		if strings.HasPrefix(*databaseEngine, "Aurora") {
```
aliscott, line 201:
```suggestion
		if databaseEngine != nil && strings.HasPrefix(*databaseEngine, "Aurora") {
```
Should we add a nil ptr check here?

internal/providers/terraform/aws/db_instance.go
```diff
@@ -192,6 +191,36 @@ func NewDBInstance(d *schema.ResourceData, u *schema.UsageData) *schema.Resource
...
+			MonthlyQuantity: backupStorageGB,
+			ProductFilter: &schema.ProductFilter{
+				VendorName:    strPtr("aws"),
+				Region:        strPtr(region),
+				Service:       strPtr("AmazonRDS"),
+				ProductFamily: strPtr("Storage Snapshot"),
+				AttributeFilters: []*schema.AttributeFilter{
+					{Key: "databaseEngine", Value: strPtr(backupStorageDbEngine)},
+					{Key: "usagetype", ValueRegex: strPtr("/BackupUsage/i")},
+					{Key: "engineCode", ValueRegex: strPtr("/[0-9]+/")},
```
aliscott, line 218: Do we need an `engineCode` filter? Looks like it's always a number anyway.

aliscott (commented):
This looks great! Thanks @joshniec 😄

internal/providers/terraform/aws/db_instance.go
joshniec: So this is a weird one, and I am very much so open to suggestions because I don't like having to do it either... but since you asked for it... get ready (wall of text). 

I can instead use the following `usagetype` query after playing around with it more: `"/(RDS:|Aurora:)(Charged)?BackupUsage/i"`, do you think that would be more elegant? For a longer explanation as for why I initially included `engineCode`: 

For every `Any` databaseEngine, there exists a "non-custom" version of the database in the `cloud-pricing-api` database, and a "custom" version of the database (see: https://aws.amazon.com/rds/custom/). I ran the `cloud-pricing-api` database tool locally to confirm my queries / attribute filters would only return _one_ result during any construction of a `aws_db_instance`. 

Scenario: Calculating the costs of a `aws_db_instance` with a `backupStorageDbEngine` value set to `Any` in the region `us-east-1`.
The following psql query will represent this search in the `cloud-pricing-api` **if I exclude the `AttributeFilter` for  `engineCode`** (only selecting the `attributes` and `prices` columns here since that is most relevant):
```psql
SELECT attributes FROM products WHERE "vendorName" = 'aws' AND "service" = 'AmazonRDS' AND "productFamily" = 'Storage Snapshot' AND "attributes"->>'databaseEngine' = 'Any' AND "attributes"->>'usagetype' LIKE '%BackupUsage%' AND "region" = 'us-east-1';
```

However, here are the results:
```
                                                                                                                                                          attributes                                                                       |                             prices
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDSCustom:ChargedBackupUsage", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "databaseEngine": "Any", "databaseEdition": "Any", "deploymentModel": "Custom"} | {"70bc2b1d05db0a22bbad7b1a9b539b70-ee3dd7e4624338037ca6fea0933a662f": [{"USD": "0.0950000000", "unit": "GB-Mo", "priceHash": "70bc2b1d05db0a22bbad7b1a9b539b70-ee3dd7e462433803
7ca6fea0933a662f", "description": "USD 0.095 per GB-Month of snapshot storage", "endUsageAmount": "Inf", "purchaseOption": "on_demand", "startUsageAmount": "0", "effectiveDateStart": "2021-11-01T00:00:00Z"}]}
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDS:ChargedBackupUsage", "engineCode": "0", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "storageMedia": "AmazonS3", "databaseEngine": "Any"}               | {"87eb8774ee46ef6838ab3918adb95698-ee3dd7e4624338037ca6fea0933a662f": [{"USD": "0.0950000000", "unit": "GB-Mo", "priceHash": "87eb8774ee46ef6838ab3918adb95698-ee3dd7e4624338037ca6fea0933a662f", "description": "$0.095 per additional GB-month of backup storage exceeding free allocation", "endUsageAmount": "Inf", "purchaseOption": "on_demand", "startUsageAmount": "0", "effectiveDateStart": "2021-11-01T00:00:00Z"}]}
(2 rows)
```

As you can see there are two results returned, if we look strictly at the attributes column, and the values in that row, we can see that the difference between the two is that one is designated as a `Custom` RDS database solution (https://aws.amazon.com/rds/custom/). [Terraform ](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance)has no support for "Custom" RDS databases as far as I can tell by their documentation. 

When I inspected the two rows (one custom, one non-custom), there were very few differences I could find.
```
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDSCustom:ChargedBackupUsage", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "databaseEngine": "Any", "databaseEdition": "Any", "deploymentModel": "Custom"}
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDS:ChargedBackupUsage", "engineCode": "0", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "storageMedia": "AmazonS3", "databaseEngine": "Any"}
```

Namely: 
- `usagetype` differs (however, `usagetype` also is different for all four database types for a particular region, `Aurora:BackupUsage`, `RDSCustom:ChargedBackupUsage`, `AuroraBackupUsage` and `RDS:ChargedBackupUsage`)
  - I could update the Regex filter here using the following regex filter: `"/(RDS:|Aurora:)(Charged)?BackupUsage/i"` which would obviate the need for the `engineCode` filter. 
- All databases _except_ the "Custom" RDS database included the `engineCode` attribute. By filtering for this attribute, it would exclude the "Custom" RDS database pricing. 

(for reference, here are all the database pricing attributes for backup storage in `us-east-1`):
```
cloud_pricing=# SELECT attributes FROM products WHERE "vendorName" = 'aws' AND "service" = 'AmazonRDS' AND "productFamily" = 'Storage Snapshot' AND "attributes"->>'databaseEngine' IS NOT NULL AND "attributes"->>'usagetype' LIKE '%BackupUsage%' AND"region" = 'us-east-1';
                                                                                                                                                                   attributes
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 {"location": "US East (N. Virginia)", "operation": "CreateDBInstance:0016", "usagetype": "Aurora:BackupUsage", "engineCode": "16", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "storageMedia": "AmazonS3", "databaseEngine": "Aurora MySQL"}
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDSCustom:ChargedBackupUsage", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "databaseEngine": "Any", "databaseEdition": "Any", "deploymentModel": "Custom"}
 {"location": "US East (N. Virginia)", "operation": "CreateDBInstance:0021", "usagetype": "Aurora:BackupUsage", "engineCode": "21", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "storageMedia": "AmazonS3", "databaseEngine": "Aurora PostgreSQL"}
 {"location": "US East (N. Virginia)", "operation": "", "usagetype": "RDS:ChargedBackupUsage", "engineCode": "0", "regionCode": "us-east-1", "servicecode": "AmazonRDS", "servicename": "Amazon Relational Database Service", "locationType": "AWS Region", "storageMedia": "AmazonS3", "databaseEngine": "Any"}
```

aliscott: Thanks @joshniec, that was a great explanation 🙌 . This makes me wonder in the future if we'll need better filtering (!=, not contains, etc). I think the way you have it makes sense just now, we might just want to add a comment to say why we're filtering on `engineCode`.

aliscott (approved):

aliscott:
@joshniec thanks for adding this! Will merge now so it gets into v0.9.14.

PR #1171, opened 2021-12-01
fix: Update pricing for AWS/Azure resources

vdmgolub:
## AWS S3
There are currently 2 products with `Requests-Tier4` usagetype: one for
Glacier and one for Standard. Glacier related product records have values
in `operation` attribute, but it is an empty string for Standard
storage. The `operation` regex with empty string returned both products.

SQL query for pricing:
```sql
select 
attributes->>'usagetype' as usagetype
, attributes->>'operation' as operation
, region
, prices
, *
from products 
where "vendorName" = 'aws' 
and service = 'AmazonS3' 
and region = 'us-east-1' 
and attributes->>'usagetype' = 'Requests-Tier4'
order by 
attributes->>'usagetype'
```

## Azure PostgreSQL Flexible Server

From Azure announcement: https://techcommunity.microsoft.com/t5/azure-database-for-postgresql/azure-database-for-postgresql-flexible-server-is-now-ga/ba-p/2987030

> As of December 1st, 2021, the price of the B1MS Burstable compute tier is being reduced by 50%.

## Azure Storage Account

Pricing has changed for Hot LRS Storage Accounts. The resource doesn't
show the cost components if their price can't be found anymore.

### DIFF
--- a/internal/providers/terraform/azure/testdata/function_app_test/function_app_test.golden
+++ b/internal/providers/terraform/azure/testdata/function_app_test/function_app_test.golden
@@ -39,7 +39,7 @@
                                                                                                                                
  azurerm_storage_account.example                                                                                               
  ├─ Capacity                                                        Monthly cost depends on usage: $0.0208 per GB              
- ├─ List and create container operations                            Monthly cost depends on usage: $0.05 per 10k operations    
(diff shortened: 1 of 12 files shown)


aliscott (approved):
Thanks @vdmgolub, LGTM

PR #1172, opened 2021-12-02
fix: use --terraform-plan-flags during init and info

tim775:
@aliscott This includes the `--terraform-plan-flags` when running terragrunt info-all AND when running terraform init.  I'm not sure about using them during init, maybe we should only do that when it is a terragrunt init?

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -32,6 +32,8 @@ type Project struct {
 	// TerraformCloudToken sets the Team API Token or User API Token so infracost can use it to access the plan.
 	// Only applicable for terraform cloud/enterprise users.
 	TerraformCloudToken string `yaml:"terraform_cloud_token,omitempty" envconfig:"INFRACOST_TERRAFORM_CLOUD_TOKEN"`
+	// TerragruntFlags set additional flags that should be passed to terragrunt.
+	TerragruntFlags string `envconfig:"INFRACOST_TERRAGRUNT_FLAGS"`
 	// UsageFile is the full path to usage file that specifies values for usage-based resources
 	UsageFile string `yaml:"usage_file,omitempty" ignored:"true"`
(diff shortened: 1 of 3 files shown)


aliscott:
> I'm not sure about using them during init, maybe we should only do that when it is a terragrunt init?

I'm hesitant about repurposing the `--terraform-plan-flags` to also work for the init. Are there any issues with flags that are valid for plan, but not init, and vice versa?

We've seen some cases before where a `--terraform-init-flags` option would be useful for providing the `-backend-config` flag, so introducing that is another option.

tim775:
> > I'm not sure about using them during init, maybe we should only do that when it is a terragrunt init?
> 
> I'm hesitant about repurposing the `--terraform-plan-flags` to also work for the init. Are there any issues with flags that are valid for plan, but not init, and vice versa?
> 
> We've seen some cases before where a `--terraform-init-flags` option would be useful for providing the `-backend-config` flag, so introducing that is another option.

Yeah I feel your hestiation. I think `--terraform-init-flags` is probably the way to go if we're ok with adding another flag.

Another option would be `--terragrunt-flags` which would be appied in plan, info, and init.  On the one hand I don't like the idea of adding a tg specific flag, but that would probably be the clearest for users so I don't know.

aliscott:
> Another option would be --terragrunt-flags which would be appied in plan, info, and init. On the one hand I don't like the idea of adding a tg specific flag, but that would probably be the clearest for users so I don't know.

Or maybe we could start with a `INFRACOST_TERRAGRUNT_FLAGS` env variable?

alikhajeh1:
Plz also add any new envs to https://www.infracost.io/docs/integrations/environment_variables/

aliscott (approved):
LGTM

PR #1181, opened 2021-12-04
Resources migration t1

sinabakh:


sinabakh:
Automatic resource migration proved to be harder than what I originally thought HAHA :))

However, it's getting somewhere. I'm pretty sure we can migrate around 60% of resources automatically, and probably we can achieve something near 90% with help of other members.

I'm doing a few tweaks to make all the 60% work and then we can discuss the challenges regarding the remaining 40%

PS. This PR is just a reference and there will be other PRs for the actual migrations.

aliscott:
@sinabakh is this PR up-to-date with the latest migrator script changes now?

sinabakh:
@aliscott Yes it is.

aliscott:
Looks like it just needs some linting fix ups and then we can merge it?

sinabakh:
Do we need to merge it? I just created this PR as a reference.
We can merge it to another branch so we can look at it for other migrations.

sinabakh:
Also, it needs few other tweaks for remaining resources.

aliscott:
> Do we need to merge it?

I don't have a strong opinion, but since it's fairly separate I thought it would be good to have it in the master branch, so we don't accidentally clean up the branch. But I'm happy either way.

sinabakh:
@aliscott Number 1 is fixed, however not so elegant. In the resource file it will always check the existence with: `r.Foo != nil` so we need to replace nil with the right thing manually.

sinabakh:
I'll close this PR since the discussions are all resolved and the process is in its final stages. (The few azure resources are left) So there's no need for this PR. However, I will keep the branch in my local fork.

PR #1184, opened 2021-12-05
docs: Update contributing guides

vdmgolub:


cmd/resourcegen/embed/resources/$cloud_provider$/$filename$.go.tmpl
```diff
@@ -71,66 +79,26 @@ var {{ .ResourceName }}UsageSchema = []*schema.UsageItem{
...
+	resources.PopulateArgsWithUsage(r, u)
 }
 
-// BuildResource builds a schema.Resource from a valid {{ .ResourceName }}.
-// This method is called after the resource is initialised by an iac provider.
+// BuildResource builds a schema.Resource from a valid {{ .ResourceName }} struct.
+// This method is called after the resource is initialised by an IaC provider.
 // See providers folder for more information.
-func ({{ .ResourceLetter }}  *{{.ResourceName}}) BuildResource() *schema.Resource {
+func (r *{{.ResourceName}}) BuildResource() *schema.Resource {
```
hugorut, line 89: Think removing `ResourceLetter` is probably a good idea as it simplifies things. But just want to make sure we're happy that all resources will have a `r` as receiver name. i.e:


```go

func (r AResource) BuildResource() {
}
```

It's kinda convention to have this start with the same letter as the struct (in the example case it would be `a`. However, I don't think it matters too much and cleans up a lot of code generation code.

cmd/resourcegen/main.go
```diff
@@ -39,7 +39,7 @@ func main() {
 	var c config
 	flag.StringVar(&c.CloudProvider, "cloud-provider", "aws", "Cloud provider to create resource for, one of [aws, azure, google]")
 	flag.StringVar(&c.Filename, "resource-name", "", "The resource name to generate, use underscores between names, e.g. autoscaling_group (required)")
-	flag.BoolVar(&c.WithHelp, "with-help", true, "Generate your resources with doc blocks and examples to help you get started. Useful for understanding how to add a resource")
+	flag.BoolVar(&c.WithHelp, "with-help", false, "Generate your resources with doc blocks and examples to help you get started. Useful for understanding how to add a resource")
```
hugorut, line 42: 👍

cmd/resourcegen/main.go
```diff
@@ -84,52 +84,53 @@ type config struct {
...
+	return "new" + c.ResourceName
 }
-`
-	case "google":
-		return `provider "google" {
-  credentials = "{\"type\":\"service_account\"}"
-  region      = "us-central1"
+
+func (c config) RegistryFuncName() string {
+	return "get" + c.ResourceName + "RegistryItem"
```
hugorut, line 101: Not a big thing here but I realise this will be different for azure. As the azure registry seems to have a `getAzureRM`  convention. I think we should probably remove that convention though.

cmd/resourcegen/main.go
```diff
@@ -84,52 +84,53 @@ type config struct {
...
 }
-`
-	case "azure":
-		return `provider "azurerm" {
-  skip_provider_registration = true
-  features {}
+
+func (c config) ProviderDirectory() string {
+	return path.Join("internal/providers/terraform/", c.CloudProvider)
```
hugorut, line 105:
```suggestion
	return "internal/providers/terraform/" + c.CloudProvider
```

no need for path.Join as `c.CloudProvider` won't have a slash in the name

hugorut (commented):
Add new resource guide is very comprehensive - nice. No major flags from me. Looks great.

cmd/resourcegen/embed/resources/$cloud_provider$/$filename$.go.tmpl
vdmgolub: I'm still new to Go naming conventions. But the idea behind is to make the code look similar between resources as much as possible. The goal is to create a scaffold where only mapping logic and cost components differ.
And as we already have `d schema.ResourceData` inconsistency, I thought we could do the same for the struct name.

cmd/resourcegen/main.go
vdmgolub: Yep, I thought to use `AzureRM` here too, but AWS/Google don't do that. And as we make the lowercased function names a default now, they will be internal for `azure` package and `AzureRM` part became kinda redundant.

cmd/resourcegen/main.go
vdmgolub: This is my inner Rubyist tries to avoid string concatenations as much as possible :D If the path string doesn't have last slash, it would look better, I guess :)

vdmgolub:
The Azure test failure is fixed in #1191.

PR #1186, opened 2021-12-06
fix: Fix pricing in golden file tests

vdmgolub:
## AWS

Per [announcement](https://aws.amazon.com/blogs/storage/s3-storage-class-price-reductions/?nc1=b_rp):

- Several storage prices are reduced.
- S3 Glacier bulk retrievals are free now. Thus their cost components
  are removed.
- S3 Glacier is renamed to S3 Glacier Flexible Retrieval.
- Update DynamoDB Table product filter. Latest update brought in a new price for IA storage and the filter was matching two records.

## Azure

Azure Pricing is inconsistent for "Hot LRS" tier: the previous pricing update removed List operations and added Write operations. But the latest one flip-flopped them again. Switching to "Cool LRS" tier seems to be stable.

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -308,10 +308,8 @@ resource_usage:
       monthly_expedited_select_data_scanned_gb: 500000 # Monthly data scanned by S3 Select in GB (for expedited level of S3 Glacier)
       monthly_expedited_select_data_returned_gb: 500000 # Monthly data returned by S3 Select in GB (for expedited level of S3 Glacier)
       monthly_standard_data_retrieval_requests: 500000 # Monthly data Retrieval requests (for standard level of S3 Glacier).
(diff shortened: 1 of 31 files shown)


aliscott (approved):
@alikhajeh1 heads up that Glacier has been [renamed Glacier Flexible Retrieval](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-s3-glacier-storage-class-amazon-s3-glacier-flexible-retrieval/). This will be a breaking change for anyone using `glacier` in the usage file.

alikhajeh1:
We validate the usage-file params so users'll be notified of the issue, correct? We should mention this in the release notes too.

vdmgolub:
@alikhajeh1 @alikhajeh1 The main reason I've renamed it is because Amazon also introduced a new `S3 Glacier Instant Retrieval` storage class. I don't see the Pricing API has prices for it yet, but they might be added with the next update.

PR #1191, opened 2021-12-07
fix(aws): Fix aws_transfer_server default protocol

vdmgolub:
Found a bug in my code while writing a new contributing code. According to TF registry the default protocol should be `SFTP`: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/transfer_server#protocols.

I've also refactored the cost components to be the same as in the guide. We're pushing to have the same style across resources, so I decided it will be a good reference.


aliscott (approved):
LGTM

PR #1193, opened 2021-12-08
Parallelize multi-project runs

aliscott:


alikhajeh1:
I tested it with a 4 repo TFC-backed infra repo on my macbook pro, total run time went from 2mins 40secs to 40secs - love it!

Regarding `--parallelism`, it's not clear if that's used by Cloud Pricing API calls as well as multi-project steps (I assume it's both?). Would it be simpler for users to do a `--no-parallelism` option (similar to `--no-cache`)? Do we even want this as a flag, or just an env variable? Having it as a flag makes me want to always specify it: I want fast tools.

alikhajeh1:
I prefer `parallel` as it's easier to type 😅 I have a knee-jerk reaction when new flags get added to `breakdown` and `diff` - would `infracost configure` be a better place for this? The idea being that in the future it can be more of a global config to limit (or speed-up) infracost across all commands - though I like the idea of being fast-as-possible-by-default.

aliscott:
> would `infracost configure` be a better place for this?

I'm not sure if this is a global option though. We recommend to turn it off to enable debugging, so seems like it should be a per-run option.

alikhajeh1:
Maybe an env to just switch it off makes more sense initially? Flags are quite prominent and it seems like this one is mainly there for debugging cases

aliscott:
So the options are:
Option 1: `--parallelism (int)` flag
Option 2: `--parallel (int)` flag
Option 3: `--no-parallelism` flag
Option 4: `--no-parallel` flag
Option 5: `INFRACOST_PARALLELISM=int` env var
Option 6: `INFRACOST_DISABLE_PARALLELISM=true` env var
Option 7: `INFRACOST_DISABLE_PARALLEL_PROJECT_RUNS=true` env var

Please add any more. My hesitance with Option 3 & 6 is that we're not totally disabling parallelism with this option, only disabling parallel project runs, but option 7 seems very verbose.

@tim775 any thoughts as well?

alikhajeh1:
In that case, +1 on option 5

tim775:
I'm ok with option 5 since we seem to want to limit the number of flags.  A few more options:

Option 8: --parallelism (true/false default true) flag
Option 9: --parallel (true/false default true) flag
Option 10: INFRACOST_PARALLELISM=true/false (default true) env var
Option 11: INFRACOST_MULTI_PROJECT_PARALLELISM=int

PR #1197, opened 2021-12-08
Fix semver comparison of Terragrunt version

Chasbob:
When trying to use the terragrunt integration I was getting the following error:

```bash
$ infracost breakdown --path prod
Detected Terragrunt directory at .
  ✔ Running terragrunt run-all terragrunt-info
Error: Terragrunt 0.35.12 is not supported. Please use Terragrunt version >= v0.28.1.
```

The problem seems to be the `semver` comparision [here](https://github.com/infracost/infracost/blob/1042c84a6294c81ffcba8e6b0bdfd9d729ab0eb2/internal/providers/terraform/dir_provider.go#L451).

Since the output of `terragrunt -version` doesn't include the `v` prefix, this will always evaluate as [less than](https://pkg.go.dev/golang.org/x/mod/semver#Compare).

```bash
$ terraform -version
Terraform v1.0.11
on linux_amd64
$ terragrunt -version
terragrunt version 0.35.12
```

### DIFF
--- a/internal/providers/terraform/dir_provider.go
+++ b/internal/providers/terraform/dir_provider.go
@@ -448,7 +448,7 @@ func checkTerraformVersion(v string, fullV string) error {
 		return errors.Errorf("Terraform %s is not supported. Please use Terraform version >= %s.", v, minTerraformVer)
 	}
 
-	if strings.HasPrefix(fullV, "terragrunt") && semver.Compare(v, minTerragruntVer) < 0 {
+	if strings.HasPrefix(fullV, "terragrunt") && semver.Compare("v"+v, minTerragruntVer) < 0 {
 		return errors.Errorf("Terragrunt %s is not supported. Please use Terragrunt version >= %s.", v, minTerragruntVer)
 	}
 


aliscott:
Thanks @Chasbob, how did you install `terragrunt`? I wonder if some install methods add the `v` prefix and some don't, so we might want the fix to only add the `v` prefix if it doesn't already exist.

When I install `terragrunt` with `brew` I get:

```
terragrunt -version
terragrunt version v0.35.13
```

aliscott:
@Chasbob just checking if you have more info on this, would be good to get this into the next release?

alikhajeh1:
> so we might want the fix to only add the v prefix if it doesn't already exist.

That doesn't hurt and seems like the most robust option

cdf-risq:
> @Chasbob just checking if you have more info on this, would be good to get this into the next release?

Sorry for the delay, I installed from the arch community repository https://archlinux.org/packages/community/x86_64/terragrunt/

> so we might want the fix to only add the v prefix if it doesn't already exist.

That makes sense, I can take a look.

tim775:
@Chasbob thanks for the PR, we've added some code to handle this that will go out in the next release (#1261).

vdmgolub:
@Chasbob this is live in v0.9.16.

Issue #1202, opened 2021-12-15
fatal error: concurrent map writes


stevie-:
When we run infracost in our github action workflow we often see a fatal error.

Versions:
```
Download action repository 'infracost/actions@v1' (SHA:dc9f4f16929775a70b3b5164c007f46695f53f37)
Download action repository 'infracost/actions@comment-tag' (SHA:524f39cf4f16f5ebe042fbf28298f7183d7c9ca1)
###
Notice: Setup Infracost CLI version 0.9.15
```

```
Run infracost breakdown --config-file=/tmp/infracost-config.yml --format=json --out-file=/tmp/infracost.json
  infracost breakdown --config-file=/tmp/infracost-config.yml --format=json --out-file=/tmp/infracost.json
  shell: /bin/bash -e {0}
  env:
    AWS_ACCOUNT: central
    AWS_ROLE_TO_ASSUME: REDACTED
    AWS_DEFAULT_REGION: eu-central-1
    AWS_REGION: eu-central-1
    AWS_ACCESS_KEY_ID: ***
    AWS_SECRET_ACCESS_KEY: ***
    AWS_SESSION_TOKEN: ***
    TERRAFORM_VERSION: 1.0.10
    TERRAGRUNT_VERSION: 0.35.13
    TF_PLUGIN_CACHE_DIR: /home/runner/.terraform.d/plugin-cache
    TERRAGRUNT_DOWNLOAD: /home/runner/.terragrunt-cache//2021-12-15
    infracost_configfile: /tmp/infracost-config.yml
    infracost_comment_tag: Validate Terraform Code with central account
    INFRACOST_GITHUB_ACTION: true
    INFRACOST_SKIP_UPDATE_CHECK: true
    INFRACOST_VCS_REPOSITORY_URL: https://github.com/REDACTED
    INFRACOST_VCS_PULL_REQUEST_URL: https://github.com/REDACTED/pull/392
    INFRACOST_LOG_LEVEL: info
fatal error: concurrent map writes

goroutine 22 [running]:
runtime.throw({0x142f960, 0xc0001048d0})
	/opt/hostedtoolcache/go/1.17.4/x64/src/runtime/panic.go:1198 +0x71 fp=0xc000104878 sp=0xc000104848 pc=0x433631
runtime.mapassign_faststr(0xc00009b3fb, 0x15d9494, {0xc00004c090, 0x4})
[...]

PR #1206, opened 2021-12-15
Fix race conditions

aliscott:
This removes three possible race conditions, including the one in https://github.com/infracost/infracost/issues/1202.

There's still possible race conditions with the logging, but I've been looking at how we can change the logging and spinners anyway which will remove these, and the logging race conditions shouldn't cause any issues at the moment anyway.

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Rhymond/go-money"
@@ -74,14 +75,24 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 		}
 	}
 
+	// Create a mutex for each path, so we can synchronize the runs of any
+	// projects that have the same path. This is necessary because Terraform
+	// can't run multiple operations in parallel on the same path.
+	pathMuxs := map[string]*sync.Mutex{}
+	for _, projectCfg := range runCtx.Config.Projects {
+		pathMuxs[projectCfg.Path] = &sync.Mutex{}
+	}
+
 	for i := 0; i < parallelism; i++ {
 		errGroup.Go(func() error {
 
 			for projectCfg := range jobs {
+				mux := pathMuxs[projectCfg.Path]
+
 				ctx := config.NewProjectContext(runCtx, projectCfg)
 				projectContextChan <- ctx
 
-				configProjects, err := runProjectConfig(cmd, runCtx, ctx, projectCfg)
+				configProjects, err := runProjectConfig(cmd, runCtx, ctx, projectCfg, mux)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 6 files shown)


cmd/infracost/run.go
```diff
@@ -74,14 +75,24 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 		}
 	}
 
+	// Create a mutex for each path, so we can synchronize the runs of any
+	// projects that have the same path. This is necessary because Terraform
+	// can't run multiple operations in parallel on the same path.
+	pathMuxs := map[string]*sync.RWMutex{}
```
hugorut, line 81: Not sure if we need a `RWMutex` here - not a big issue tho. 

```suggestion
	pathMuxs := map[string]*sync.Mutex{}
```

cmd/infracost/run.go
```diff
@@ -285,7 +301,8 @@ func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config
 		EnableLogging: runCtx.Config.IsLogging(),
 		NoColor:       runCtx.Config.NoColor,
 	}
-	spinner = ui.NewSpinner("Calculating monthly cost estimate", spinnerOpts)
+	spinner := ui.NewSpinner("Calculating monthly cost estimate", spinnerOpts)
+	defer spinner.Fail()
```
hugorut, line 305: do we want `spinner.Fail()` here - or `spinner.Stop()`? This question goes for all the `Fail` method calls.

internal/providers/detect.go
```diff
@@ -155,7 +156,14 @@ func isTerraformDir(path string) bool {
 	return terraform.IsTerraformDir(path)
 }
 
+// goformation lib is not threadsafe, so we run this check synchronously
+// See: https://github.com/awslabs/goformation/issues/363
+var cfMux = &sync.RWMutex{}
```
hugorut, line 161: Again not sure if we need a read/write mutex. Would prefer just simple mutex to avoid confusion.

```suggestion
var cfMux = &sync.Mutex{}
```

hugorut (commented):
Looks good 🥇 . Minor clarification needed on the `Fail` functionality. Also would prefer use of base `Mutex` but don't mind if you want to stick to `RWMutex`

cmd/infracost/run.go
aliscott: I did `Fail()` because if there's a panic or unexpected error then we show an ❌ . If we've previously called `.Success()` then it won't do anything. I don't think it really matters in this case if the spinner doesn't show that though, so open to change it to `Stop()` if you think that's less confusing.

hugorut: 👍  makes sense to me - I was just a bit confused by the method name.

cmd/infracost/run.go
```diff
@@ -74,14 +75,24 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 		}
 	}
 
+	// Create a mutex for each path, so we can synchronize the runs of any
+	// projects that have the same path. This is necessary because Terraform
+	// can't run multiple operations in parallel on the same path.
+	pathMuxs := map[string]*sync.Mutex{}
+	for _, projectCfg := range runCtx.Config.Projects {
+		pathMuxs[projectCfg.Path] = &sync.RWMutex{}
```
hugorut, line 83:
```suggestion
		pathMuxs[projectCfg.Path] = &sync.Mutex{}
```

cmd/infracost/run.go
```diff
@@ -181,7 +192,12 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 	return nil
 }
 
-func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config.ProjectContext, projectCfg *config.Project) ([]*schema.Project, error) {
+func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config.ProjectContext, projectCfg *config.Project, mux *sync.RWMutex) ([]*schema.Project, error) {
```
hugorut, line 195:
```suggestion
func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config.ProjectContext, projectCfg *config.Project, mux *sync.Mutex) ([]*schema.Project, error) {
```

hugorut (approved):
LGTM

PR #1207, opened 2021-12-16
docs: add region usage appendix

hugorut:
Adds helper appendix to adding new resource docs to make sure developers know to use the shared `RegionsUsage` struct. This means we only have one place to maintain region keys and it's easier for contributors to get up and running.

### DIFF
--- a/contributing/add_new_resource_guide.md
+++ b/contributing/add_new_resource_guide.md
@@ -33,6 +33,7 @@ The steps below describe adding an AWS resource using Terraform IaC provider, bu
   - [References to other resources](#references-to-other-resources)
   - [Google zone mappings](#google-zone-mappings)
   - [Azure zone mappings](#azure-zone-mappings)
+  - [Region usage](#region-usage)
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


contributing/add_new_resource_guide.md
```diff
@@ -1162,3 +1163,63 @@ Unless the resource has global or zone-based pricing, the first line of the reso
 		...
 	}
   ```
+
+### Region usage
+
+A number of resources have usage costs which vary on a per-region basis. This means that developers often have to define
+a usage file with a complex map key with all the cloud provider regions. E.g. `google_artifact_registry_repository` can have 
+any number of google regions under the `monthly_egress_data_transfer_gb` usage param:
```
alikhajeh1, line 1171:
```suggestion
A number of resources have usage costs which vary on a per-region basis. This means that you often have to define a usage file with a complex map key with all the cloud provider regions. For example, `google_artifact_registry_repository` can have any number of Google regions under the `monthly_egress_data_transfer_gb` usage parameter:
```

contributing/add_new_resource_guide.md
```diff
@@ -1162,3 +1163,63 @@ Unless the resource has global or zone-based pricing, the first line of the reso
...
+    monthly_egress_data_transfer_gb:
+      us_east1: 100
+      us_west1: 100
+      australia_southeast1: 100
+      europe_north1: 100
+      southamerica_east1: 100
+```
+
+If you have a resource like this, rather than defining your own usage field, it is advisable you use one of the shared `RegionsUsage`
+structs that handle this structure for you. These structs can be found in both [the google](../internal/resources/google/util.go) & [aws](../internal/resources/aws/util.go) resource util files.
```
alikhajeh1, line 1185:
```suggestion
If you have a resource like this, rather than defining your own usage field, you should use one of the shared `RegionsUsage` structs that handle this structure for you. These structs can be found in both [the `google`](../internal/resources/google/util.go) & [`aws`](../internal/resources/aws/util.go) resource util files.
```

contributing/add_new_resource_guide.md
```diff
@@ -1162,3 +1163,63 @@ Unless the resource has global or zone-based pricing, the first line of the reso
...
+
+```go
+type MyResource struct {
+    ...
+    MonthlyDataProcessedGB *RegionsUsage `infracost_usage:"monthly_processed_gb"`
+}
+```
+
+And then after `PopulateUsage` is called it can be accessed to retrieve set values. `RegionsUsage` helper struct also comes with 
+a `Values` method that returns the set values as a `slice` with key/value pairs that is helpful to iterate over to create cost components, e.g with a usage like so:
```
alikhajeh1, line 1197:
```suggestion
And then after `PopulateUsage` is called it can be accessed to retrieve set values. `RegionsUsage` helper struct also comes with a `Values` method that returns the set values as a `slice` with key/value pairs that is helpful to iterate over to create cost components, e.g with a usage like so:
```

alikhajeh1 (approved):
Thanks!

I guess when the Appendix becomes too long (it's fine now), we'll have to do another refactor to see where these section should fit into the flow.

(cc @vdmgolub)

hugorut:
@alikhajeh1  ❤️

PR #1210, opened 2021-12-19
Add CI Dockerfile

aliscott:
* Adds CI Dockerfile
* Adds `infracost-comment` wrapper script
* Adds VCS repo and PR URL detection so we don't need to export these env vars in the integrations
* Fixes a bug with Azure DevOps repo URL parsing


Dockerfile.ci
```diff
@@ -0,0 +1,35 @@
...
+# Install the latest compost version
+RUN npm install -g @infracost/compost
+
+WORKDIR /root/
+
+# Scripts are used by CI integrations and other use-cases
+COPY scripts/infracost-comment.sh /usr/bin/infracost-comment
+COPY --from=builder /app/build/infracost /usr/bin/
+
+ENV INFRACOST_SKIP_UPDATE_CHECK='true'
```
alikhajeh1, line 32: do we not want CI logs to show upgrades when they're available? what do other tools like Pulumi do?

alikhajeh1 (commented):
awesome!

Dockerfile.ci
aliscott: In existing CI integrations we set this. If the user is using the `ci-0.9` tag then it will update automatically. So this will only make a difference if they are locking to a patch version (e.g. `ci-0.9.16`). In this case I don't think we should show the message since the user is explicitly setting that version. I don't think it really matters either way since the message will also not be obvious since it will only appear in the CI logs.

alikhajeh1: Good point, I had forgotten about `ci-0.9` :)

scripts/infracost-comment.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# This script posts a comment on a pull request or commit.
+# It first runs `infracost output` to generate the comment and then wraps
+# compost (https://github.com/infracost/compost) to post the comment.
+# 
+# Usage:
+#  ./infracost-comment.sh --platform <PLATFORM> --format <FORMAT> --behavior <BEHAVIOR> --target-type=<TARGET_TYPE> <INFRACOST_JSON_PATHS>
+#
+# Example:
+#   ./infracost-comment.sh --platform gitlab --format gitlab-comment --behavior update --target-type=merge-request <INFRACOST_JSON_PATHS>
```
tim775, line 12:
```suggestion
#   ./infracost-comment.sh --platform gitlab --format gitlab-comment --behavior update --target-type=merge-request my_infracost_breakdown.json
```

tim775 (approved):

PR #1211, opened 2021-12-20
chore(azure): disable Hot LRS example for Azure StorageV2 account resource

vdmgolub:
Azure pricing records for Hot LRS operations are identical apart from
meterName. CPAPI calculates productHashes not including the meterName
attribute thus different operations may be overriden by each other
during update jobs. Disabling the Hot LRS example unblocks CLI's CI
while we can come up with a solution.


hugorut (approved):
LGTM

PR #1216, opened 2021-12-21
feat: add run uuid to cli

hugorut:
* Adds uuid to run context which is populated to ApiClient
* Remove `X-Trace-Id` header which was unused and replace with `X-Infracost-Trace-Id` populated with run context uuid

See https://github.com/infracost/cloud-pricing-api/pull/137 for pricing api implementation
See https://github.com/infracost/dashboard/pull/26 for dashboard implementation

### DIFF
--- a/internal/apiclient/pricing.go
+++ b/internal/apiclient/pricing.go
@@ -30,7 +30,7 @@ type PriceQueryResult struct {
 }
 
 func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
-	currency := cfg.Config.Currency
+	currency := ctx.Config.Currency
 	if currency == "" {
 		currency = "USD"
 	}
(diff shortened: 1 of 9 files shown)


hugorut:
Tested with both `cloud-pricing-api` and `hosted-cloud-pricing-api` and going through to mixpanel fine. Sentry alerts also capturing request header `X-Infracost-Trace-Id` when errors caught.

internal/apiclient/pricing.go
```diff
@@ -29,8 +29,8 @@ type PriceQueryResult struct {
 	Result gjson.Result
 }
 
-func NewPricingAPIClient(cfg *config.Config) *PricingAPIClient {
-	currency := cfg.Currency
+func NewPricingAPIClient(cfg *config.RunContext) *PricingAPIClient {
```
vdmgolub, line 32: No biggy, but it seems like `ctx` would be a more descriptive variable name with the updated type. What do you think?
```suggestion
func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
```

internal/prices/prices.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/tidwall/gjson"
 )
 
-func PopulatePrices(cfg *config.Config, project *schema.Project) error {
+func PopulatePrices(cfg *config.RunContext, project *schema.Project) error {
```
vdmgolub, line 15: Same comment about the name.
```suggestion
func PopulatePrices(ctx *config.RunContext, project *schema.Project) error {
```

vdmgolub (approved):
Looks great 👍 I've added a couple minor comments about arguments naming, please take a look.

internal/apiclient/pricing.go
hugorut: good shout

PR #1217, opened 2021-12-22
feat: Add json schema for output json

tim775:
This adds the a JSON schema file for the output JSON at `examples/jsonschema/outputschema.json` (though I'm open to suggestions on a better place to keep it).


aliscott (approved):
Looks great! Down the line can we use this schema for validating the files passed to `infracost output`?

Suggestion for where to put it: top level `schema` dir, and in a file called `infracost.json`, `infracost_output.json` or `infracost.schema.json`

tim775:
> Down the line can we use this schema for validating the files passed to `infracost output`?

Yeah I imagine that would be pretty straightforward since there are packages that can do that.

PR #1212, opened 2021-12-20
feat: add share links

vdmgolub:

PR #1218, opened 2021-12-22
fix: do not send run results to Dashboard when using self-hosted CPAPI

vdmgolub:
With self-hosted Cloud Pricing API the request to Infracost Dashboard
will not be authenticated.

Related to #1212.


aliscott (approved):
LGTM

Issue #1213, opened 2021-12-21
`aws_elasticache_replication_group.cluster_mode` is present in non clustered aws_elasticache_replication_group state


hugorut:
When applying a `aws_elasticache_replication_group` in non clustered mode terraforms uses [number_cache_clusters](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_replication_group#number_cache_clusters) to specify replication nodes.

However after applying terraform, the state adds the `cluster_mode` block to the resulting state. e.g:

```hcl

resource "aws_elasticache_replication_group" "non-cluster" {
  replication_group_description = "This Replication Group"
  replication_group_id          = "tf-rep-group-2"

  engine = "redis"

  node_type             = "cache.r5.4xlarge"
  number_cache_clusters = 3
}

```

results in something like so:

```hcl
resource "aws_elasticache_replication_group" "non-cluster" {
  replication_group_description = "This Replication Group"
  replication_group_id          = "tf-rep-group-2"

  engine = "redis"

  node_type             = "cache.r5.4xlarge"
  number_cache_clusters = 3
  
  cluster_enabled               = false
  cluster_mode {
      num_node_groups         = 1
      replicas_per_node_group = 1
  }
}

```

note the `cluster_enabled` and `cluster_mode` additions. This causes problems with cmds such as infracost diff as infracost uses the `cluster_mode` block and overrides the cost meaning it doesn't pick up on price changes if users have changed `number_cache_clusters` flag.

alikhajeh1:
@hugorut thanks for creating this! Can you possibly also add any details of new cost components or units - or is it just a quantity that we need to multiply by?

So far I've:
- added the "needs details" label
- removed that label after details are added to the issue description.
- set the project of the issue to Main so show the community that it's something that we'd like help with sooner rather than later (the details make a big difference to the community being able to pick it up).

hugorut:
@alikhajeh1 I was actually a bit quick to jump on this into an enhancement - it actually seems we do handle the `number_cache_clusters` in `aws_elasticache_replication_group` but it seems this feature isn't being picked up by the user. I'm going to rename the title and mark as a bug and ask for further details of their terraform file to replicate.

vdmgolub:
This is live in v0.9.16.

PR #1224, opened 2021-12-23
fix(aws): solve cluster_mode inconsistency with terraform state

hugorut:
Fixes: https://github.com/infracost/infracost/issues/1213

Solves issue where terraform state was causing resources to have no price change in `infracost diff` due to `cluster_mode` block addition.

### DIFF
--- a/internal/providers/terraform/aws/elasticache_replication_group.go
+++ b/internal/providers/terraform/aws/elasticache_replication_group.go
@@ -1,9 +1,10 @@
 package aws
 
 import (
-	"github.com/infracost/infracost/internal/schema"
-
 	"github.com/shopspring/decimal"
+	"github.com/tidwall/gjson"
+
+	"github.com/infracost/infracost/internal/schema"
 )
 
 func GetElastiCacheReplicationGroupItem() *schema.RegistryItem {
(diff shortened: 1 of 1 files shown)


internal/providers/terraform/aws/elasticache_replication_group.go
```diff
@@ -22,7 +23,19 @@ func NewElastiCacheReplicationGroup(d *schema.ResourceData, u *schema.UsageData)
...
+	//   cluster_mode {
+	//    replicas_per_node_group = 1
+	//    num_node_groups         = 1
+	//  }
+	//
+	// even when cluster mode is disabled. This causes an issue with infracost diff as it sees no change in prices
+	// as both the new and the old resource have cluster_mode with 1 1 set. In order to circumvent this problem we
+	// need to explicitly check that cluster_enabled attribute (output attribute) is set to false in the terraform state.
+	// This will only be present in a state/diff run and won't be available in breakdown or output runs.
+	clusterDisabled := d.Get("cluster_enabled").Type != gjson.Null && d.Get("cluster_enabled").Bool()
```
aliscott, line 37:
```suggestion
	clusterDisabled := d.Get("cluster_enabled").Type != gjson.Null && !d.Get("cluster_enabled").Bool()
```

Is there a golden test we could add for this?

hugorut: I guess we'd have to do a `diff` golden file test, I can look into something

aliscott: We've done them before by adding a sanitized plan JSON to the test data (e.g. https://github.com/infracost/infracost/blob/master/cmd/infracost/testdata/terraform_v0.12_plan.json).

But I don't think it's important, so I wouldn't spend time on it just now.

hugorut: good spot on the `!` i had altered this because of a `lint` error and changed the stmt

aliscott (approved):

PR #1225, opened 2021-12-23
fix(azure): add further filters to compute dtu to avoid price flip floppping

hugorut:


internal/resources/azure/sql_database.go
```diff
@@ -119,6 +119,7 @@ func (r *SQLDatabase) dtuPurchaseCostComponents() []*schema.CostComponent {
 			ProductFilter: r.productFilter([]*schema.AttributeFilter{
 				{Key: "productName", ValueRegex: strPtr("/^SQL Database Single/i")},
 				{Key: "skuName", ValueRegex: strPtr(fmt.Sprintf("/^%s$/i", skuName))},
+				{Key: "meterName", ValueRegex: strPtr("/DTUs$/i")},
```
vdmgolub, line 122: A newer fancy way :D
```suggestion
				{Key: "meterName", ValueRegex: regexPtr("DTUs$")},
```

vdmgolub (approved):

internal/resources/azure/sql_database.go
hugorut: nice

PR #1227, opened 2021-12-26
refactor(aws): resources migration batch1

sinabakh:
giphy


sinabakh:
@aliscott I migrated a few resources and fixed and changed all `_gb` keys to float. There seems to be a problem in syncing the usage file that is happening just for `cloudwatch_log_group` that keeps the `.0` for floats. For other resources, it works just fine. I think it might be related to wildcard and module syncing but I couldn't find the problem.

It would be great if you could take a look at the failing test.

aliscott:
@sinabakh I tested out the failing test. I think we just want to update the golden file for this case. The sync usage adds .0 for float values, and I think that makes sense. We already have it for the `monthly_data_processed_gb` in that golden file.

internal/providers/terraform/aws/cloudwatch_log_group.go
```diff
@@ -1,84 +1,18 @@
 package aws
 
 import (
+	"github.com/infracost/infracost/internal/resources/aws"
 	"github.com/infracost/infracost/internal/schema"
-	"github.com/shopspring/decimal"
 )
 
 func GetCloudwatchLogGroupItem() *schema.RegistryItem {
```
aliscott, line 8: Might not be possible, but can we update these to be internal functions (`getCloudwatchLogGroupItem`). We've started using that for new resources.

sinabakh:
> @sinabakh I tested out the failing test. I think we just want to update the golden file for this case. The sync usage adds .0 for float values, and I think that makes sense. We already have it for the `monthly_data_processed_gb` in that golden file.

I think that might not be the case, there are other resources that have float usage data with no `.0` and tests are not failing. `transfer_server` is an example, it has `monthly_data_downloaded_gb` and `monthly_data_uploaded_gb` and they are both floats but there is no problem in thinking the usage file for it.

internal/providers/terraform/aws/cloudwatch_log_group.go
sinabakh: I think it's easy, I'll applyt it.

aliscott:
> > @sinabakh I tested out the failing test. I think we just want to update the golden file for this case. The sync usage adds .0 for float values, and I think that makes sense. We already have it for the `monthly_data_processed_gb` in that golden file.
> 
> I think that might not be the case, there are other resources that have float usage data with no `.0` and tests are not failing. `transfer_server` is an example, it has `monthly_data_downloaded_gb` and `monthly_data_uploaded_gb` and they are both floats but there is no problem in thinking the usage file for it.

I only see usage data for `aws_transfer_server` that is read by tests, not any that is written/synced by tests (https://github.com/sinabakh/infracost/blob/refactor-migration-aws-batch1/internal/providers/terraform/aws/testdata/transfer_server_test/transfer_server_test.usage.yml).

internal/providers/terraform/aws/cloudwatch_log_group.go
```diff
@@ -1,84 +1,18 @@
...
 	return &schema.RegistryItem{
 		Name:  "aws_cloudwatch_log_group",
-		RFunc: NewCloudwatchLogGroup,
+		RFunc: NewCloudwatchLogGroupItem,
```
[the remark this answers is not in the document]
sinabakh: It was getting the name from registry func name, fixed it by using the resource actual name.

sinabakh:
@aliscott Ready for review and merge

aliscott (approved):
Thanks @sinabakh! Will merge this after the v0.9.16 release so it gets into v0.9.17.

PR #1228, opened 2021-12-27
feat: create a share run link for output command

vdmgolub:


cmd/infracost/output.go
```diff
@@ -143,6 +143,33 @@ func outputCmd(ctx *config.RunContext) *cobra.Command {
...
+
+			if ctx.Config.EnableDashboard {
+				if ctx.Config.IsSelfHosted() {
+					ui.PrintWarning(cmd.ErrOrStderr(), "The dashboard is part of Infracost's hosted services. Contact hello@infracost.io for help.")
+				}
+
+				projectContexts := []*config.ProjectContext{}
+				for range combined.Projects {
+					projectContexts = append(projectContexts, config.EmptyProjectContext())
+				}
```
hugorut, line 155: would prefer allocating the length of the slice since we know it's length 
```suggestion
				projectContexts := make([]*config.ProjectContext{}, len(combined.Projects)
				for i, _ := range combined.Projects {
					projectContexts[i] = config.EmptyProjectContext()
				}
```

cmd/infracost/output.go
```diff
@@ -143,6 +143,33 @@ func outputCmd(ctx *config.RunContext) *cobra.Command {
 			if cmd.Flags().Changed("fields") && !contains(validFieldsFormats, format) {
 				ui.PrintWarning(cmd.ErrOrStderr(), "fields is only supported for table and html output formats")
 			}
+
+			if ctx.Config.EnableDashboard {
+				if ctx.Config.IsSelfHosted() {
+					ui.PrintWarning(cmd.ErrOrStderr(), "The dashboard is part of Infracost's hosted services. Contact hello@infracost.io for help.")
+				}
```
hugorut, line 150: do we want to return here - or is it enough to just flag a warning?

hugorut (commented):
minor tweaks

cmd/infracost/output.go
vdmgolub: Warning is enough, the client won't make a request, but we still want the results. We do the same in the breakdown/diff commands.

hugorut (approved):
🚢🚢

PR #1232, opened 2021-12-29
chore: optimize run's share URL generation for output command

vdmgolub:
In case when the output command receives only one file and it already
has its share URL, sending its data to Dashboard would create a
redundant duplicate.

### DIFF
--- a/cmd/infracost/output.go
+++ b/cmd/infracost/output.go
@@ -149,25 +149,7 @@ func outputCmd(ctx *config.RunContext) *cobra.Command {
 					ui.PrintWarning(cmd.ErrOrStderr(), "The dashboard is part of Infracost's hosted services. Contact hello@infracost.io for help.")
 				}
 
-				projectContexts := make([]*config.ProjectContext, len(combined.Projects))
-				for i := range combined.Projects {
-					projectContexts[i] = config.EmptyProjectContext()
-				}
-				combinedRunIds := []string{}
-				for _, input := range inputs {
-					if id := input.Root.RunID; id != "" {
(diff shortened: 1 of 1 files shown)


cmd/infracost/output.go
```diff
@@ -149,25 +149,30 @@ func outputCmd(ctx *config.RunContext) *cobra.Command {
...
-				}
+					dashboardClient := apiclient.NewDashboardAPIClient(ctx)
+					result, err := dashboardClient.AddRun(ctx, projectContexts, combined)
+					if err != nil {
+						log.Errorf("Error reporting run: %s", err)
+					}
 
-				combined.RunID, combined.ShareURL = result.RunID, result.ShareURL
+					combined.RunID, combined.ShareURL = result.RunID, result.ShareURL
+				}
```
hugorut, line 175: could we separate this into a function so that we could return early and/or remove to much nesting. I generally hate `if/else` and try and avoid if possible. e.g. something like: 

```go

combined.RunID, combined.ShareURL = combineRuns(input)

```

with

```go

func combineRuns(inputs []output.ReportInput) (int64, string) {
    if len(inputs) == 1 && inputs[0].Root.RunID != "" {
        result := inputs[0].Root
	return result.RunID, result.ShareURL
    }
    
    ...
}
```

hugorut (commented):
Minor gripe about too much nesting

hugorut (approved):
@vdmgolub as discussed over zoom - happy to have this merged. Just personal preference about `if/else` & nesting. I'll leave the decision up to you.

PR #1233, opened 2021-12-29
feat: add resourcecheck cmd

hugorut:
Adds resourcecheck cmd to test regions are implemented, sends slack notification if we are missing a newly added resource:

Screenshot 2021-12-29 at 12 15 43

Runs on a daily schedule


aliscott (approved):
Nice idea! LGTM

PR #1234, opened 2021-12-29
Update comment script to use environment variables

aliscott:
I haven't added the autodetect for the comment format yet. I think we should look at adding a compost command for that, but I've excluded it from this just now.


scripts/ci/comment.sh
```diff
@@ -0,0 +1,110 @@
...
+  echo "Skipping compost:" >&2
+  echo "  $compost_cmd" >&2
+else
+  if [ "$dry_run" = true ]; then
+    echo "Running compost in dry run mode:" >&2
+  else
+    echo "Running compost:" >&2
+  fi
+  echo "  $compost_cmd" >&2
+  eval "$compost_cmd"
```
vdmgolub, line 103: Do you think it's worth checking that the compost app exists? Although the Dockerfile installs it, it is still possible to run the script from the source code. Something like this maybe?
```bash
if [ -f "compost" ]; then
    eval "$compost_cmd"
else
  echo "Compost is not installed, can't post the comment"
  exit 1
fi
```

UPD: Maybe the current behavior is enough.
```
/scripts/ci/comment.sh: line 103: compost: command not found
```

scripts/ci/comment.sh
```diff
@@ -0,0 +1,110 @@
...
+#
+# For testing:
+#   - COMMENT_DRY_RUN: Run compost in dry run mode so comments aren't posted, updated or deleted.
+#   - COMMENT_SKIP_COMPOST: Skip the call to compost
+# 
+# Usage:
+#   COMMENT_FORMAT=<FORMAT> COMMENT_BEHAVIOR=<BEHAVIOR> COMMENT_TARGET_TYPE=<TARGET-TYPE> ./comment.sh <INFRACOST_JSON_PATHS>
+#
+# Example:
+#   COMMENT_FORMAT=gitlab-comment COMMENT_BEHAVIOR=update COMMENT_TARGET_TYPE=merge-request ./comment.sh <INFRACOST_JSON_PATHS>
```
vdmgolub, line 37: As it is an example, we probably could also provide the files:
```suggestion
#   COMMENT_FORMAT=gitlab-comment COMMENT_BEHAVIOR=update COMMENT_TARGET_TYPE=merge-request ./comment.sh infracost-file-1.json infracost-file-2.json
```

vdmgolub (approved):
Looks great, easy to read and understand! 👍 Left couple thoughts, but these are super minor.

scripts/ci/comment.sh
aliscott: 👍

Issue #1238, opened 2021-12-31
certificate signed by unknown authority issue introduced in v0.9.15


rickardosmith:
I get the following printed to console when running version 0.9.15
```powershell
Detected Terraform directory at .
  ✔ Checking for cached plan... found
✖ Calculating monthly cost estimate


Error: Error sending API request: Post "https://pricing.api.infracost.io/graphql": x509: certificate signed by unknown authority
```

When I downgraded to version 0.9.14, the tool runs as expected
```powershell
Detected Terraform directory at .
  ✔ Checking for cached plan... found

✔ Calculating monthly cost estimate


Update: A new version of Infracost is available: v0.9.14 → v0.9.15
  Go to https://www.infracost.io/docs/update for instructions     

Running in "."

Summary: {
  "totalDetectedResources": 8,
  "totalSupportedResources": 1,
  "totalUnsupportedResources": 0,
  "totalUsageBasedResources": 1,
  "totalNoPriceResources": 7,
  "unsupportedResourceCounts": {},
  "noPriceResourceCounts": {
    "aws_internet_gateway": 1,
    "aws_network_interface": 1,
    "aws_route_table": 1,
    "aws_route_table_association": 1,
    "aws_security_group": 1,
    "aws_subnet": 1,
    "aws_vpc": 1
  }
}

Total Monthly Cost:        8.392 USD
Total Monthly Cost (diff): 8.392 USD
```

**Additional information**:
- I was running with version 0.9.5 at first, then upgraded version 0.9.15, noticed the error, then downgraded to version 0.9.14
- System info:
```
Edition	Windows 10 Pro
Version	21H2
OS build	19044.1415
Experience	Windows Feature Experience Pack 120.2212.3920.0
```

tim775:
Thanks for the issue.  I'm not able to reproduce on my dev machine so I'm spinning up a windows box in the cloud.  

If you have a chance could you see if setting `infracost configure set tls_insecure_skip_verify false` is a work-around?

tim775:
I was able to reproduce on windows, and confirm that `.\infracost configure set tls_insecure_skip_verify true` is a work-around.

tim775:
I guess there is a long-standing issue where the golang doesn't know how to load system certs on windows: https://github.com/golang/go/issues/16736.

rickardosmith:
> 

@tim775 confirming that the workaround does work in my environment.

[...]

PR #1239, opened 2021-12-31
Fix/windows tls certs

tim775:
This is a fix for #1238.  Go doesn't know how to find the root CAs on windows, so this PR causes us to skip the cert verify step if no root CAs are found and verification has not been explicitly enabled.

### DIFF
--- a/internal/apiclient/pricing.go
+++ b/internal/apiclient/pricing.go
@@ -55,8 +55,17 @@ func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
 		}
 	}
 
+	skipVerify := false
+	if ctx.Config.TLSInsecureSkipVerify != nil {
+		skipVerify = *ctx.Config.TLSInsecureSkipVerify
+	} else if len(rootCAs.Subjects()) == 0 && ctx.Config.TLSCACertFile == "" {
+		// Skip verify hasn't been explicitly set, there's no cert file, and there are no root CAs (i.e. this is
+		// windows).  Default to skipping the verify.
+		skipVerify = true
+	}
+
 	tlsConfig := &tls.Config{
-		InsecureSkipVerify: ctx.Config.TLSInsecureSkipVerify, // nolint: gosec
+		InsecureSkipVerify: skipVerify, // nolint: gosec
 		RootCAs:            rootCAs,
 	}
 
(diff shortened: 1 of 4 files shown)


alikhajeh1:
Should we log something to debug so the user has a way to tell if TLS verification is being skipped? What should Windows users do if they want TLS? wait until we upgrade to an upcoming go version that doesn't have this issue?

internal/apiclient/pricing.go
```diff
@@ -55,8 +55,17 @@ func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
 		}
 	}
 
+	skipVerify := false
+	if ctx.Config.TLSInsecureSkipVerify != nil {
+		skipVerify = *ctx.Config.TLSInsecureSkipVerify
+	} else if len(rootCAs.Subjects()) == 0 && ctx.Config.TLSCACertFile == "" {
```
aliscott, line 61: In this case, should we just pass a nil `tlsConfig` to the API client instead of skipping verify? If we do that will it still work with Windows and verify the pricing.api.infracost.io cert?

aliscott:
> What should Windows users do if they want TLS?

From what I can tell this only changes the default case where the user isn't passing in their own cert. So, this will still work for Windows users who want to skip verification of a self-hosted endpoint, or still verify their own self-hosted cert.

aliscott:
Do you think we should add an extra story for running these smoke test steps on different OS runners as well? https://github.com/infracost/infracost/blob/master/.github/workflows/go.yml#L40-L56 Would this have picked up this issue?

internal/apiclient/pricing.go
```diff
@@ -35,12 +35,14 @@ func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
 		currency = "USD"
 	}
 
-	rootCAs, _ := x509.SystemCertPool()
-	if rootCAs == nil {
-		rootCAs = x509.NewCertPool()
-	}
+	tlsConfig := tls.Config{} // nolint: gosec
```
aliscott, line 38: I'm assuming passing an empty `tls.Config` works the same as passing nil yeah?

aliscott (approved):
LGTM

internal/apiclient/pricing.go
tim775: Yeah the pattern in the libraries is
```
	if t1.TLSClientConfig == nil {
		t1.TLSClientConfig = new(tls.Config)
	}
```

PR #1240, opened 2021-12-31
fix: Indent the "Calculating monthly cost estimate" spinner

tim775:
This makes the indenation on the "Calculating monthly cost estimate" spinner consistent with the other spinner lines.  It looks like they have been inconsistent for a while, but recent changes removed the blank line so it is now more apparent.

Old:
```
Detected Terraform directory at ./examples/terraform
  ✔ Checking for cached plan... found
✔ Calculating monthly cost estimate 
```

With this fix:
```
Detected Terraform directory at ./examples/terraform
  ✔ Checking for cached plan... found
  ✔ Calculating monthly cost estimate 
```

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -329,6 +329,7 @@ func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config
 	spinnerOpts := ui.SpinnerOptions{
 		EnableLogging: runCtx.Config.IsLogging(),
 		NoColor:       runCtx.Config.NoColor,
+		Indent:        "  ",
 	}
 	spinner := ui.NewSpinner("Calculating monthly cost estimate", spinnerOpts)
 	defer spinner.Fail()


aliscott:
The reason for this originally is that the indented spinners are for a specific project, but the `Calculating monthly cost estimate` spinner is for all projects. Saying that, if this doesn't look confusing for multiple projects then I think it's good to go.

tim775:
The recent parallelism changes effect how logging for multiple projects is displayed so maybe the original reasoning isn't relevent.  With two projects and parallelism 1 i get 
```
Detected Terraform directory at examples/terraform
  ✔ Checking for cached plan... found
  ✔ Calculating monthly cost estimate 

Detected Terraform directory at examples/terraform
  ✔ Checking for cached plan... found
  ✔ Calculating monthly cost estimate 
```

aliscott:
👍  makes sense, thanks!

PR #1261, opened 2022-01-03
fix: Make sure terraform/grunt version is prefixed with v

tim775:
This is a follow up to #1197


aliscott (approved):
LGTM. I'm unable to reproduce this issue, but this seems safe enough to merge anyway.

PR #1267, opened 2022-01-05
fix(google): fix few logging resources rfunc call

sinabakh:


sinabakh:
I don't know if my changes make sense or not. I faced this issue while migrating google resources.

aliscott (approved):
Thanks!

PR #1287, opened 2022-01-18
chore: update fatal error messaging in cli

hugorut:
* updates cli messaging if there is a fatal panic in the application
* refactors `main.go` to delegate all responsibility to a `Run` function 
* refactor `cmd_test.go` to use `Run` so we can capture application logs and ensure total e2e coverage


cmd/infracost/cmd_test.go
```diff
@@ -32,46 +36,43 @@ func DefaultOptions() *GoldenFileOptions {
...
-		logBuf = testutil.ConfigureTestToCaptureLogs(t, runCtx)
-	} else {
-		testutil.ConfigureTestToFailOnLogs(t, runCtx)
-	}
-
-	var cmdErr error
 	var actual []byte
 
-	cmdErr = rootCmd.Execute()
+	main.Run(func(c *config.RunContext) {
```
aliscott, line 53: 🙌 this is neat

cmd/infracost/main.go
```diff
@@ -128,7 +145,8 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
 	))
 
 	rootCmd.SetVersionTemplate("Infracost {{.Version}}\n")
-	rootCmd.SetOut(os.Stdout)
+	rootCmd.SetOut(ctx.OutWriter)
+	rootCmd.SetErr(ctx.ErrWriter)
```
aliscott, line 149: 🙌

aliscott (commented):
Looks good.

Do we want to hide the stacktrace on panics? It's not what I would expect based on my usage of other tools, and I find it helpful if I want to find a workaround, or submit a PR myself.

If we do want to hide it, then I think we should either dump it to a tmp crash file and output the location of the file, or tell them how they can show stacktraces by enabling debug logging (although re-running might not cause the same panic).

hugorut:
> Looks good.
> 
> Do we want to hide the stacktrace on panics? It's not what I would expect based on my usage of other tools, and I find it helpful if I want to find a workaround, or submit a PR myself.
> 
> If we do want to hide it, then I think we should either dump it to a tmp crash file and output the location of the file, or tell them how they can show stacktraces by enabling debug logging (although re-running might not cause the same panic).

Not sure - that's why I added the two approaches: It's hidden by default but they can see the stack trace if the log level is set to `debug`. I'm in two minds about showing the stack trace, one side of me thinks it's not good as it highlights that maybe we haven't handled this case (even though we have), but the other side thinks it would be good for personal debugging. I'm leaning towards not showing by default to add a more "smooth" experience, but really not too sure. @alikhajeh1 what's your feeling on this?

alikhajeh1:
+1 on only showing it on debug. I'd also consider shortening the output to:

An unexpected error occurred. We've been notified of it and will investigate it soon. If you'd like to follow-up, please re-run with `--log-level debug` and create an issue at https://github.com/infracost/infracost/issues

aliscott:
> I'm leaning towards not showing by default to add a more "smooth" experience

I've been thinking about this more and I think I'm a fairly strong +1 for either showing the panic even on the stderr or dumping it to a `crash.log` file and outputting the location of that file for the following reasons:
1. The error that happened won't necessarily happen again if I re-run with `--log-level=debug`. It could be an intermittent issue with the API that cascades to a panic, it could be a race condition or it could be an issue that's only present with that log level.
2. We won't necessarily receive the error upstream 
3. It's what I expect as a CLI user for an unexpected error as it's consistent with what other tools do
4. It helps me as a user work out what the issue could be and come up with a workaround.

hugorut:
> I've been thinking about this more and I think I'm a fairly strong +1 for either showing the panic on the stderr or dumping it to a `crash.log` file and outputting the location of that file for the following reasons:

Makes sense - I'll revert to keep the stack trace then and just change messaging so that it's more prominent that we're handling this.

aliscott (approved):
LGTM!

PR #1289, opened 2022-01-18
feat: add eval cmd

hugorut:
## Adds eval cmd support

Note: 

> _This work is a POC so that we can see if we like this and would like to adopt it. I've done just enough to get the functionality working, but not tidy up the code. If we do adopt this we'll need to polish up the interface, tidy up the code and add some golden tests. However, this PR is purely for evaluative purposes._

### What

This work adds a new `eval` cmd (working title) that is a slim wrapper over the `breakdown` cmd. The `eval` cmd also takes a `--policy` flag which user can pass a path to their infracost OPA policy. 

`eval` then takes the output of a breakdown and evaluates it against the passed policy. 

#### Examples

```
$ infracost  eval --path examples/terragrunt --usage-file=examples/terraform/infracost-usage.yml  --policy policy.rego
```

with a passing policy:

Screenshot 2022-01-18 at 20 12 32

with a failing policy:

Screenshot 2022-01-18 at 20 13 19

### Why

1. Having first class support for policies as part of infracost simplifies usage. For example the OPA example in the `azure devops pipeline` currently looks like this:

```yml
pr:
  - master

jobs:
  - job: open_policy_agent
    displayName: Open Policy Agent
    pool:
      vmImage: ubuntu-latest

    steps:
      - task: InfracostSetup@0
        displayName: Setup Infracost
        inputs:
          apiKey: $(infracostApiKey)

      - bash: |
          INSTALL_LOCATION="$(Build.SourcesDirectory)/bin"
          mkdir -p ${INSTALL_LOCATION}

          wget -O ${INSTALL_LOCATION}/opa https://openpolicyagent.org/downloads/v${OPA_VERSION}/opa_linux_amd64_static
          chmod +x ${INSTALL_LOCATION}/opa

          echo "##vso[task.setvariable variable=PATH;]$(PATH):${INSTALL_LOCATION}"

        displayName: Setup OPA
        env:
          OPA_VERSION: 0.35.0

      - bash: infracost breakdown --path=examples/opa/code/plan.json --format=json --out-file=/tmp/infracost.json
        displayName: Run Infracost

      - bash: opa eval --input /tmp/infracost.json -d examples/opa/policy/policy.rego --format pretty "data.infracost.deny" | tee /tmp/opa.out
        displayName: Run OPA
      
      - bash: infracost test --path=examples/opa/code/plan.json -d examples/opa/policy/policy.rego

      - bash: |
          denyReasons=$(</tmp/opa.out)
          if [ "$denyReasons" != "[]" ]; then
            echo "##vso[task.logissue type=error]Policy check failed"
            echo "##vso[task.logissue type=error]$denyReasons"
            echo "##vso[task.complete result=Failed;]DONE"
          else
            echo "Policy check passed."
          fi
        displayName: Check Policies
```

with eval it goes to:

```yml
pr:
  - master

jobs:
  - job: open_policy_agent
    displayName: Open Policy Agent
    pool:
      vmImage: ubuntu-latest

    steps:
      - task: InfracostSetup@0
        displayName: Setup Infracost
        inputs:
          apiKey: $(infracostApiKey)

      - bash: infracost eval --path=examples/opa/code/plan.json --policy=policy.rego
        displayName: Check Cost Policies
```

2. With `eval` infracost uses OPA as a library - there is no dependency on the binary. This means better errors and debugging.
3. Adopting policies as a first class citizen
4. Less noisy - something we hear a lot


tim775:
This is cool!  I wonder if we need a separate `eval` command or if we could add policy checking to existing steps.  I.e., if you pass `--policy` to `breakdown`, `diff`, or `output`, it will run the "Evaluating cost policy" step.

aliscott:
> I.e., if you pass --policy to breakdown, diff, or output

For this case we could also add the policy result to the Infracost JSON so we could use it in PR comments, slack messages, etc.

hugorut:
> This is cool! I wonder if we need a separate `eval` command or if we could add policy checking to existing steps. I.e., if you pass `--policy` to `breakdown`, `diff`, or `output`, it will run the "Evaluating cost policy" step.

hmm, I'm not sure about that as it might conflict with the other commands purpose. E.g. I don't want my `breakdown` to fail and not show output if my Cost Policies have failed. The exception to that rule is `infracost comment` which I think could be augmented by `--policy` as we could set rules to post on failure e.t.c. My thinking with separating into a isolated cmd is that users could use this as a `red/green` cmd which is ideally suited to CI.

tim775:
> I don't want my breakdown to fail and not show output if my Cost Policies have failed.

Yeah that's true, but I think I want it to fail and show my breakdown output.  Especially of that output is annotated with information about which resources caused the policy to fail.

I wonder how much "Check the policy" is a separate step vs something that's always happening.

hugorut:
more and more I think about this I agree with @tim775 that we should have this a a flag which is can always be part of `breakdown` -> I'm going to work on the two options in my free time and do a loom which I'll ask for feedback from the OPA/policy community. Putting this PR in draft for the time being.

alikhajeh1:
Thanks @hugorut! I noticed that in our recent cost policy [demo video](https://www.youtube.com/watch?v=1rMIfebfd8M) I wanted to get across two things:
1. the cost estimate diff is put in pull requests so the team becomes more cost-aware over time.
2. the guardrails (cost policies) kick-in if there are major issues.

That gives me the feeling that `infracost comment github ... --path infracost-*.json --policy policy.rego` is what I wanted so the comment contains both the estimate and any policy violations (red/green) - that saves me from having to read the CI logs.

**TL;DR:** +1 on having `--policy-path` in comment. To be consistent I suppose it's needed in breakdown, diff and output too.

aliscott (approved):
LGTM. My concern is what happens when a policy isn't in the format we expect. Currently we get a debug level log which I didn't see and the policy shows as passed. Should we have a neutral "Policy is invalid" message and show the issue in the comment?

hugorut:
> LGTM. My concern is what happens when a policy isn't in the format we expect. Currently we get a debug level log which I didn't see and the policy shows as passed. Should we have a neutral "Policy is invalid" message and show the issue in the comment?

@aliscott good idea - I definitely think we should make it more visible 👍

PR #1297, opened 2022-01-20
Feat/add comment cmd

tim775:


internal/comment/github.go
```diff
@@ -0,0 +1,480 @@
...
+
+// IsHidden returns true if the comment is hidden or minimized.
+func (c *githubComment) IsHidden() bool {
+	return c.isMinimized
+}
+
+// GitHubExtra contains any extra inputs that can be passed to the GitHub comment handlers.
+type GitHubExtra struct {
+	// APIURL is the URL of the GitHub API. This can be set to a custom URL if
+	// using GitHub enterprise. If not set, the default GitHub API URL will be used.
```
vdmgolub, line 63:
```suggestion
	// using GitHub Enterprise. If not set, the default GitHub API URL will be used.
```

internal/comment/github.go
```diff
@@ -0,0 +1,480 @@
...
+		SubjectID:  githubv4.NewString(githubv4.String(comment.(*githubComment).globalID)),
+		Classifier: githubv4.ReportedContentClassifiersOutdated,
+	}
+
+	return h.v4client.Mutate(ctx, &m, input, nil)
+}
+
+// githubCommitHandler is a PlatformHandler for GitHub commits. It
+// implements the PlatformHandler interface and contains the functions
+// for finding, creating, updating, deleting comments on GitHub commits.
```
vdmgolub, line 305: Should we mention this? :)

```suggestion
// for finding, creating, updating, deleting and hiding comments on GitHub commits.
```

internal/comment/gitlab.go
```diff
@@ -0,0 +1,537 @@
...
+// IsHidden always returns false for GitLab since GitLab doesn't have a
+// feature for hiding comments.
+func (c *gitlabComment) IsHidden() bool {
+	return false
+}
+
+// GitLabExtra contains any extra inputs that can be passed to the GitLab comment handlers.
+type GitLabExtra struct {
+	// ServerURL is the URL of the GitLab server. This can be set to a custom URL if
+	// using GitLab enterprise. If not set, the default GitLab server URL will be used.
```
vdmgolub, line 61: I think it's the official name of the product, but `GitLab Enterprise` is probably okay too :)
```suggestion
	// using GitLab for Enterprise. If not set, the default GitLab server URL will be used.
```

internal/comment/gitlab.go
```diff
@@ -0,0 +1,537 @@
...
+
+	return h.graphqlClient.Mutate(ctx, &m, variables)
+}
+
+// CallHideComment calls the GitLab API to minimize the merge request comment.
+func (h *gitlabPRHandler) CallHideComment(ctx context.Context, comment Comment) error {
+	return errors.New("Not implemented")
+}
+
+// githubCommitHandler is a PlatformHandler for GitLab commits. It
```
vdmgolub, line 295:
```suggestion
// gitlabCommitHandler is a PlatformHandler for GitLab commits. It
```

vdmgolub (approved):
This is great! Really cool 😀 That's interesting that we have to use both REST and GraphQL APIs to work with comments. 🙃
All my comments are minor typo fixes. I saw couple opportunities to DRY the commands, but it seems too early, maybe we can do it after adding more platforms.

cmd/infracost/comment_gitlab.go
```diff
@@ -0,0 +1,137 @@
...
+  Post a new comment to a commit:
+
+      infracost comment gitlab --repo my-org/my-gitlab-repo --commit 2ca7182 --path infracost.json --behavior hide-and-new --gitlab-token $GITLAB_TOKEN`,
+		ValidArgs: []string{"--", "-"},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// pricingClient := apiclient.NewPricingAPIClient(ctx)
+			// err = pricingClient.AddEvent("infracost-comment", ctx.EventEnv())
+			// if err != nil {
+			// 	log.Errorf("Error reporting event: %s", err)
+			// }
```
vdmgolub: Seems like a leftover
```suggestion
```

cmd/infracost/comment_github.go
```diff
@@ -0,0 +1,140 @@
...
+					log.Errorf("Error reporting event: %s", err)
+				}
+
+				cmd.Println("Comment posted to GitHub")
+			} else {
+				cmd.Println(string(body))
+				cmd.Println("Comment not posted to GitHub (--dry-run was specified)")
+			}
+
+			if outFile, _ := cmd.Flags().GetString("out-file"); outFile != "" {
```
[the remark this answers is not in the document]
tim775: Hmmm.  I guess if you want only the markdown you can use `infracost output --format github-token`... so maybe there's not a strong use case for out-file and we should just remove it for now.  I was thinking it would be nice for testing, but I made the --dry-run print the comment to stdout and that may be enough for tests.

@aliscott any thoughts?

internal/comment/gitlab.go
tim775: Doesn't seem important for a comment

cmd/infracost/comment_github.go
```diff
@@ -0,0 +1,133 @@
...
+				err = commentHandler.CommentWithBehavior(ctx.Context(), behavior, string(body))
+				if err != nil {
+					return err
+				}
+
+				pricingClient := apiclient.NewPricingAPIClient(ctx)
+				err = pricingClient.AddEvent("infracost-comment", ctx.EventEnv())
+				if err != nil {
+					log.Errorf("Error reporting event: %s", err)
+				}
```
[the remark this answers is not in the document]
tim775: I'm still trying to grok the style of go code, but it seems like DRYing up code is done more cautiously than ruby.  I think the philosophy tends to recognize that adding abstractions adds more cognitive load, so in this case maybe that means the one-liner `events.Send(...` leaves some question unanswered that the 4 liner answers immediately: where does it get sent?  what gets sent? what happens if it fails?  I'm not saying I'm opposed to the refactor, just that I think that may be part of the tradeoff 🤷‍♂️

cmd/infracost/comment_github.go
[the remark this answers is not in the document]
aliscott: The main existing reason for `--dry-run` is for the CI examples E2E tests. 

We want to replace lines like this in the examples:

```sh
script:
  ...
  - infracost comment gitlab --repo my-org/my-gitlab-repo --merge-request 3 --path infracost.json --gitlab-token $GITLAB_TOKEN
```

With something that will output the generated comment so we can compare it to the golden file.

So I think we have a few options:

1. `--dry-run` performs as normal but skips write actions (posting, hiding, deleting comments), and outputs the comment to stdout.
2. Same as 1, but instead of `--dry-run` we use `INFRACOST_COMMENT_DRY_RUN` env var, since there's no user use-case for `--dry-run` yet.
3. Same as 1, but doesn't output the comment, we use a combination of `--dry-run` and `--out-file` for that
4. `--dry-run` Only outputs the comment to stdout

I'm (weakly) leaning towards number 1 or 2. The issue with 4 is that it doesn't test that we pass the correct credentials to the GitHub/GitLab API.

I don't know the use case for `--out-file` on its own yet, so I say we leave it out until we hear from users.

cmd/infracost/comment_github.go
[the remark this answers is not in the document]
vdmgolub: Agree, it's interesting where this thing line lies: when to abstract and when to keep it explicit. I think the additional cognitive load can be eliminated with thoughtful naming: my example is not great `events.Send` is really vague, maybe something like `telemetry.SendEvent` would be more descriptive (or `analytics.Track` similar to JS Segment library). I would also argue that too much explicitness can add more cognitive load too, as this event sending thing is not the most important part of the function, but you still need to read through it anyway in multiple places. What's clear is that I need to educate myself more in this Go's department :)

cmd/infracost/comment_github.go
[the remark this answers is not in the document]
tim775: good point about no use case for dry-run, I'll make it an env so opt 2

PR #1303, opened 2022-01-22
feat: add hcl only provider

hugorut:
This PR adds **HCL Only** parsing method for `infracost breakdown`. HCL only means that infracost cmds run by parsing the raw HCL code without relying on terraform to provide an output. 

This means users can run infracost cmds without providing credentials or connecting to remote providers. This not only simplifies running but significantly reduces the time to run a `breakdown`.

This work is *Experimental*


hugorut:
@aliscott not sure we need to give this a full review, but just wanted to make sure we're happy to merge this

aliscott (approved):

Issue #1288, opened 2022-01-18
Migrate Remaining AWS Resources


sinabakh:
Here are the steps to migrate a resource:
1. Pull latest changes from my `refactor-migration-t1` branch.
2. Set the var `SINGLE_RESOURCE_NAME` in `cmd/migrator/main.go` to the resource filename.
3. Run `go run cmd/migrator/main.go`, the resource file would be edited and a new resource file will be created in `internal/resources/aws`
4. All things that the script was unable to migrate are changed to the `UNKNOWN` keyword. There might even be need for logic changes.
5. Read the old logic and code and try to fix things as smoothly as possible.
6. Create a separate PR for your changes.

Important notes:
1. Let me know if you had any idea that the script could make things easier
2. Don't push resources on the aforementioned branch
3. Please sync the branch with the master whenever a migration pr is merged to master
4. There are cases where resources are dependent on each other, in these cases please convert all of them in one place

sinabakh:
Here's a list of remaining resources:
- [ ] internal/providers/terraform/aws/cloudfront_distribution.go
- [ ] internal/providers/terraform/aws/cloudwatch_metric_alarm.go
- [ ] internal/providers/terraform/aws/dx_connection.go
- [ ] internal/providers/terraform/aws/dx_gateway_association.go
- [ ] internal/providers/terraform/aws/ebs_snapshot.go
- [ ] internal/providers/terraform/aws/ebs_snapshot_copy.go
- [ ] internal/providers/terraform/aws/ec2_transit_gateway_peering_attachment.go
- [ ] internal/providers/terraform/aws/ec2_transit_gateway_vpc_attachment.go
- [ ] internal/providers/terraform/aws/ecs_service.go
- [ ] internal/providers/terraform/aws/efs_file_system.go
- [ ] internal/providers/terraform/aws/elasticache_cluster.go
- [ ] internal/providers/terraform/aws/elasticache_replication_group.go
- [ ] internal/providers/terraform/aws/kinesis_firehose_delivery_stream.go
- [ ] internal/providers/terraform/aws/neptune_cluster_snapshot.go
- [ ] internal/providers/terraform/aws/route53_record.go
- [ ] internal/providers/terraform/aws/route53_resolver_endpoint.go
- [ ] internal/providers/terraform/aws/vpc_endpoint.go
- [ ] internal/providers/terraform/aws/waf_web_acl.go
- [ ] internal/providers/terraform/aws/wafv2_web_acl.go

[...]

PR #1309, opened 2022-01-27
refactor: Migrate AWS cloudfront/cloudwatch resources to new structure

vdmgolub:
- [x] internal/providers/terraform/aws/cloudfront_distribution.go
- [x] internal/providers/terraform/aws/cloudwatch_metric_alarm.go

Related to https://github.com/infracost/infracost/issues/1288


sinabakh:
I think `cloudfront_distribution` and `dx_connection` are the hardest resources of aws for migration. I personally don't know what the right structure for them is.

vdmgolub:
@sinabakh Yeah, agree, this one is tough cookie. I have it in not-working state right now, but didn't have time to finish. Hope to finish it on a weekend :)

internal/resources/aws/cloudfront_distribution.go
```diff
@@ -0,0 +1,659 @@
...
+		Name:            "Field level encryption requests",
+		Unit:            "10k requests",
+		UnitMultiplier:  decimal.NewFromInt(10000),
+		MonthlyQuantity: quantity,
+		ProductFilter: &schema.ProductFilter{
+			VendorName: strPtr("aws"),
+			Service:    strPtr("AmazonCloudFront"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "requestDescription", Value: strPtr("HTTPS Proxy requests with Field Level Encryption")},
+				{Key: "location", Value: strPtr("Europe")},
```
vdmgolub, line 595: The "Europe" location has been used in original code. The pricing appears to be the same for all regions.

internal/providers/terraform/aws/cloudfront_distribution.go
```diff
@@ -1,527 +1,38 @@
...
-		return nil
-	}
-
-	// we need to find the legacy usage key that's defined in the usage file
-	// this is done so we support backwards compatibility and don't nuke old usage files.
-	var usageKey string
-	if v, ok := regionShieldMapping[region]; ok {
-		usageKey = v
-	}
+	isOriginShieldEnabled := d.Get("origin.0.origin_shield.0.enabled").Bool() || false
```
aliscott, line 16:
```suggestion
	isOriginShieldEnabled := d.Get("origin.0.origin_shield.0.enabled").Bool()
```
Not important, but this should default to false anyway so `|| false` is not needed.

internal/providers/terraform/aws/cloudfront_distribution.go
```diff
@@ -1,527 +1,38 @@
...
-	if d.IsEmpty("logging_config.0.bucket") {
-		return nil
-	}
-
-	var quantity *decimal.Decimal
-	if u != nil && !u.IsEmpty("monthly_log_lines") {
-		quantity = decimalPtr(decimal.NewFromInt(u.Get("monthly_log_lines").Int()))
+	var originShieldRegion string
+	if !d.IsEmpty("origin.0.origin_shield.0.origin_shield_region") {
+		originShieldRegion = d.Get("origin.0.origin_shield.0.origin_shield_region").String()
```
aliscott, line 23:
```suggestion
       originShieldRegion := d.Get("origin.0.origin_shield.0.origin_shield_region").String()
```
This defaults to an empty string either way so no need for the `IsEmpty` check

internal/providers/terraform/aws/cloudfront_distribution.go
```diff
@@ -1,527 +1,38 @@
...
-
-	// we need to find the legacy usage key that's defined in the usage file
-	// this is done so we support backwards compatibility and don't nuke old usage files.
-	var usageKey string
-	if v, ok := regionShieldMapping[region]; ok {
-		usageKey = v
-	}
+	isOriginShieldEnabled := d.Get("origin.0.origin_shield.0.enabled").Bool() || false
+	isSSLSupportMethodVIP := d.Get("viewer_certificate.0.ssl_support_method").String() == "vip"
+	hasLoggingConfigBucket := !d.IsEmpty("logging_config.0.bucket")
```
aliscott, line 18:
```suggestion
	hasLoggingConfigBucket := d.Get("logging_config.0.bucket").String() != ""
```
Should we check that this is not an empty string as well?

internal/providers/terraform/aws/cloudfront_distribution.go
```diff
@@ -1,527 +1,38 @@
...
+	hasFieldLevelEncryptionID := !d.IsEmpty("default_cache_behavior.0.field_level_encryption_id")
```
aliscott, line 19:
```suggestion
	hasFieldLevelEncryptionID := d.Get("default_cache_behavior.0.field_level_encryption_id").String() != ""
```
Should we check that this is not an empty string as well?

internal/resources/aws/cloudfront_distribution.go
```diff
@@ -0,0 +1,659 @@
...
+
+type CloudfrontDistribution struct {
+	Address string
+	Region  string
+
+	IsOriginShieldEnabled     bool
+	IsSSLSupportMethodVIP     bool
+	HasLoggingConfigBucket    bool
+	HasFieldLevelEncryptionID bool
+	OriginShieldRegion        *string
```
aliscott, line 25:
```suggestion
	OriginShieldRegion        string
```
I think this could just be a string and we can check later if it is non-empty.

internal/resources/aws/cloudfront_distribution.go
```diff
@@ -0,0 +1,659 @@
...
+				components,
+				r.buildDataOutCostComponent(tierNames[i], fromLocation, tierStarts[i], &tiers[i]),
+			)
+		}
+	}
+
+	return components
+}
+
+func (r *CloudfrontDistribution) buildDataOutCostComponent(usageName, fromLocation string, startUsage int, quantity *decimal.Decimal) *schema.CostComponent {
```
aliscott, line 322: Not suggesting we change this, just wondering why you're using `build` prefix for this cost component method?

internal/resources/aws/cloudwatch_metric_alarm.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+
+	"github.com/shopspring/decimal"
+)
+
+type CloudwatchMetricAlarm struct {
+	Address            string
+	Region             string
+	ComparisonOperator string
+	MetricCount        int64
```
aliscott, line 16:
```suggestion
	Metrics        int64
```
Not  got a strong opinion on this, this is just how we handle this case elsewhere.

aliscott (approved):
LGTM!!! Thanks @vdmgolub

internal/providers/terraform/aws/cloudfront_distribution.go
vdmgolub: `IsEmpty` checks that already:
```go
// Return true if the key doesn't exist, is null, or is an empty string.
// Needed because gjson.Exists returns true as long as a key exists, even if it's empty or null.
func (d *ResourceData) IsEmpty(key string) bool {
	g := d.RawValues.Get(key)
}
```

internal/resources/aws/cloudfront_distribution.go
vdmgolub: Mostly to distinguish `dataOutToInternetCostComponents` function name and `buildDataOutCostComponent` that is used in the first one. It should not be used as standalone cost components. Otherwise `dataOutToInternetCostComponents` and `dataOutCostComponent` would look pretty similar and it would be hard to understand which one to include in resource's cost components.

internal/resources/aws/cloudwatch_metric_alarm.go
vdmgolub: Updated 👍

internal/resources/aws/cloudfront_distribution.go
vdmgolub: Updated 👍

internal/providers/terraform/aws/cloudfront_distribution.go
vdmgolub: `IsEmpty` checks that as well 👍

internal/providers/terraform/aws/cloudfront_distribution.go
vdmgolub: Updated 👍

internal/providers/terraform/aws/cloudfront_distribution.go
aliscott: 👍 makes sense

internal/resources/aws/cloudfront_distribution.go
aliscott: 👍

PR #1312, opened 2022-01-28
feat: Add gh action to update download redirects on release

tim775:
@aliscott This is the GH action method of udpating the release redirects which would require add IAM permissions to use amplify update-app.  I think it's the way to go, but I looked into doing this in the amplify pipeline and it is possible by adding something like this to the build script:
```
    postBuild:
      commands:
        - | 
          latestCLI=$(curl -s https://github.com/infracost/infracost/releases/latest/download/infracost-linux-amd64.tar.gz | pcregrep -o1 'https://github.com/infracost/infracost/releases/download/(v\d+\.\d+\.\d+)/infracost-linux-amd64.tar.gz')
          majorMinor=$(echo $latestCLI | pcregrep -o1 '(v\d+\.\d+)\.\d+')
          updatedRules=$(aws amplify get-app --app-id $AMPLIFY_ID | \
            jq -r '.app.customRules |
               del( .[] | select( .source == "/downloads/'$majorMinor'/<*>" ) ) |
               . += [{ "source": "/downloads/'$majorMinor'/<*>",
                       "target": "https://github.com/infracost/infracost/releases/download/'$latestCLI'/<*>",
                       "status": "302" }]')
```
The main issue I had with the pipeline is that `jq` and `pcregrep` aren't installed on the build image so we would either have to roll a custom image or rewrite this using commands that are available (probably `awk` or `sed`).  It kinda seems like that's  getting too wonky, but I'm interested to hear what you think.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -87,3 +87,14 @@ jobs:
         with:
           token: ${{secrets.HOMEBREW_RELEASE_TOKEN}}
           formula: infracost
+
+  update-download-redirects:
+    name: Upload Download Redirects
+    needs: build
+    uses: ./.github/workflows/update-download-redirects.yml
+    with:
+      version: ${{ github.ref_name }}
+    secrets:
+      AMPLIFY_APP_ID: ${{ secrets.AMPLIFY_APP_ID }}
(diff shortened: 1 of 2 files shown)


aliscott (approved):
+1 on this way. It's cleaner and more maintainable.

PR #1315, opened 2022-01-30
docs: simplify help texts

alikhajeh1:
@vdmgolub can you please review this? This keeps the style of our help texts the same as previous commands. 
I included update behavior in the top-level comment help so it's clear there is an option for it, but I left it out in the sub-command as that help text shows that it's the default.

If you're ok with these, can you please update the golden files and merge?


vdmgolub (approved):

alikhajeh1:
Thanks @vdmgolub!

Issue #1175, opened 2021-12-03
Make links in CLI output simpler to read


alikhajeh1:
The current blue links are sometimes not easy to read, e.g. see the first screenshot from a user vs the second screenshot from my terminal (iTerm2):

![image](https://user-images.githubusercontent.com/638577/144635984-7ff5f334-b07a-433e-9e0d-4d1c2b22aa4e.png)

![Screen Shot 2021-12-03 at 7 17 55 PM](https://user-images.githubusercontent.com/638577/144636188-c01e185d-c4eb-4ad8-bfbc-0a8cd8b40fee.png)

## Note for community

Please 👍 if you'd like to upvote this or receive updates.

vdmgolub:
@alikhajeh1 What if we use underlined styling instead of the colors?
![image](https://user-images.githubusercontent.com/87082/155324322-0c577c96-3440-499f-893c-36a2c4a18b87.png)

alikhajeh1:
LGTM! thank you

alikhajeh1:
@emocharnik, @penja and @PatMyron just FYI that as of [v0.9.19](https://github.com/infracost/infracost/releases), released today, the link colors are white and bold, can you please try that and let us know if that helped?

emocharnik:
thanks, @alikhajeh1!

PR #1316, opened 2022-01-30
chore: ci/cd help messaging

hugorut:
- [x]  additional data processing clarification
- [x]  Expand CI/CD detection
- [x]  Mention CI/CD in CLI runs
- [x]  Update Infracost --help to include link to CI/CD

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -327,7 +327,7 @@ func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config
 		NoColor:       runCtx.Config.NoColor,
 		Indent:        "  ",
 	}
-	spinner := ui.NewSpinner("Extracting only cost-related params from terraform plan\nRetrieving cloud prices to calculate costs", spinnerOpts)
+	spinner := ui.NewSpinner("Retrieving cloud prices to calculate costs", spinnerOpts)
 	defer spinner.Fail()
 
 	for _, project := range projects {
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -157,6 +157,7 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 		return err
 	}
 
+	r.IsCIRun = runCtx.IsCIRun()
 	r.Currency = runCtx.Config.Currency
 
 	dashboardClient := apiclient.NewDashboardAPIClient(runCtx)
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -326,7 +326,7 @@ func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config
 		NoColor:       runCtx.Config.NoColor,
 		Indent:        "  ",
 	}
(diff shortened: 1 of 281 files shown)


hugorut:
ah interesting the PR is failing as it's detecting the CI system and not appending the help comment. I'll fix to detect if in test as well

cmd/infracost/run.go
```diff
@@ -326,7 +327,7 @@ func runProjectConfig(cmd *cobra.Command, runCtx *config.RunContext, ctx *config
 		NoColor:       runCtx.Config.NoColor,
 		Indent:        "  ",
 	}
-	spinner := ui.NewSpinner("Calculating monthly cost estimate", spinnerOpts)
+	spinner := ui.NewSpinner("Extracting only cost-related params from terraform plan\nRetrieving cloud prices to calculate costs", spinnerOpts)
```
aliscott, line 330: Should this be two separate spinners? The first one (`Extracting only cost-related params...`) inside the `LoadResources` call starting after the plan JSON has been generated, and the second one (`Retrieving cloud prices`) here.

hugorut: hmm in the implementation docs it's as such:

```
  ✔ Extracting only cost-related params from terraform plan
     Retrieving cloud prices to calculate costs
 ```
 
 paging dr @alikhajeh1 - thoughts?

aliscott: Ah okay, I left a comment on the original docs that suggested this option as two separate spinners, but it looks that comment was resolved so wasn't appearing any more.

I don't think having multi-lines works with our current spinner library anyway.

hugorut: 👍  - I'll alter

aliscott:
Screenshot 2022-01-30 at 15 45 22

@alikhajeh1 @hugorut seems like a lot of links at the bottom now. I'm wondering if we want a different color for primary and secondary links, but maybe we do this when we fix https://github.com/infracost/infracost/issues/1175.

aliscott (approved):
LGTM pending the spinner changes!

alikhajeh1:
@hugorut can you please review my latest commits? If that wording reads ok for you, can you possibly update the tests and merge this?

@aliscott / @hugorut 4 links colored the same does indeed seem like a lot (usage-file, new resource issue, share link, CI/CD). I made the top 2 links white so they don't stand out just now as those are not the primary actions. How's that?
Screen Shot 2022-01-30 at 10 58 12 AM

hugorut:
> @aliscott / @hugorut 4 links colored the same does indeed seem like a lot (usage-file, new resource issue, share link, CI/CD). I made the top 2 links white so they don't stand out just now as those are not the primary actions. How's that?

Looks good 👍

alikhajeh1:
@hugorut infracost.io/cicd is live so this can be shipped, I'll change the URL to goto a new dedicated CI page Mon/Tue

hugorut:
Merging this - the failing test is to do with the frontdoor issue, we're following here: https://github.com/hashicorp/terraform-provider-azurerm/issues/15170

PR #1340, opened 2022-02-02
fix: add in hcl only identifiers so that resources can reference 

hugorut:
fixes golden file tests for a number of resources that use `resource.my_resource.id` and then Infracost looks up a reference based on this `id`


aliscott (approved):
LGTM

PR #1341, opened 2022-02-02
fix: recursive block evaluation in output

hugorut:
1. Fixes test/parsing so that HCL blocks are marshalled recursively. This solves Infracost lookup on resources that have many nested blocks. e.g. `aws_cloudfront_distribution`, where Infracost needs to fetch `origin_shield.enabled`/`origin_shield_region`

```
resource "aws_cloudfront_distribution" "s3_distribution_0_with_shield" {
  origin {
    domain_name = aws_s3_bucket.b.bucket_regional_domain_name
    origin_id   = local.s3_origin_id
   ...
    origin_shield {
      enabled              = true
      origin_shield_region = "sa-east-1"
    }
  }

  enabled             = true
  ...  
}
```

2. Fixes issue where attributes referenced which were non `id` based, e.g. name, were not being evaluated correctly. We now add the `expressions` into the HCL plan JSON output so Infracost can lookup attributes based on this. e.g.:

```
  launch_template {
    name    = aws_launch_template.foo3.name # now supported in HCL
    version = "default_version"
  }
```

### DIFF
--- a/internal/hcl/attribute.go
+++ b/internal/hcl/attribute.go
@@ -129,6 +129,32 @@ func getIndexValue(part hcl.TraverseIndex) string {
 	}
 }
 
+// Reference returns the pointer to a Reference struct that holds information about the Attributes
+// referenced block. Reference achieves this by traversing the Attribute Expression in order to find the
+// parent block. E.g. with the following HCL
+//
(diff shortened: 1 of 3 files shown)


aliscott:
Sorry for the late review. This looks good.

PR #1342, opened 2022-02-03
fix: provider region specification and provide sane defaults

hugorut:
* add top provider region information to HCL output so that prices reflect both top level provider and aliased providers
* remove "count" attribute from output which was causing duplication of prices
* add defaults to a number of resources so that the HCL output is in line with what the Terraform schema displays in docs.
* changes the method for azure `location` retrieval to format the strings as they are returned from the azure provider
---

With these changes only one test is failing across all the provider golden files (`aws_autoscaling_group`) 🍾


aliscott:
LGTM!

PR #1343, opened 2022-02-03
chore: Generate share links for comment command

vdmgolub:
If Dashboard support is enabled the `comment` command will combine the
provided reports and send them to Infracost Dashboard similar to
`output` command's behavior.

### DIFF
--- a/cmd/infracost/comment.go
+++ b/cmd/infracost/comment.go
@@ -3,6 +3,7 @@ package main
 import (
 	"github.com/infracost/infracost/internal/config"
 	"github.com/infracost/infracost/internal/output"
+	"github.com/infracost/infracost/internal/ui"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/infracost/comment.go
```diff
@@ -35,7 +36,7 @@ func commentCmd(ctx *config.RunContext) *cobra.Command {
 	return cmd
 }
 
-func buildCommentBody(ctx *config.RunContext, paths []string, mdOpts output.MarkdownOptions) ([]byte, error) {
+func buildCommentBody(cmd *cobra.Command, ctx *config.RunContext, paths []string, mdOpts output.MarkdownOptions) ([]byte, error) {
```
vdmgolub, line 39: I tried different approaches but this makes the most sense as it's part of the command anyway:
- Introduce "commentOptions" with `DryRun` attribute and pass it here, but the self-hosted warning would be tricky to print.
- Sending the run in each github/gitlab/azure comment functions seemed too repetitive.

cmd/infracost/comment_azure_repos.go
```diff
@@ -2,15 +2,16 @@ package main
...
+
 	"github.com/infracost/infracost/internal/apiclient"
 	"github.com/infracost/infracost/internal/comment"
 	"github.com/infracost/infracost/internal/config"
 	"github.com/infracost/infracost/internal/output"
 	"github.com/infracost/infracost/internal/ui"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"strconv"
-	"strings"
```
vdmgolub, line 13: Auto-formatting doesn't pollute this PR too much, so decided to keep these changes.

cmd/infracost/comment.go
tim775: My only thought is should dryRun be part of the the RunContext, but we're not using it elsewhere so it seems ok to me like this.

tim775 (approved):

PR #1346, opened 2022-02-05
HCL remote module loading

aliscott:
What this does:
* Recursively loops through all Terraform modules found in the HCL code
* Downloads any modules from a Terraform registry or remote source to `.infracost/terraform_modules`. This supports the [same remote sources](https://www.terraform.io/language/modules/sources) as `terraform init`
* Saves the downloaded module's location in a `.infracost/terraform_modules/manifest.json`. This file is the same format as the `.terraform/modules/modules.json` file.
* The next time it runs, it checks if any new modules need to be downloaded by checking the modules against the manifest file.

What this doesn't do:
* If the Terraform code has already been init'd we still download the modules again into the `.infracost` directory. I think this simplifies the process/debugging, by having one way Infracost works which is independent from the Terraform environment.


internal/hcl/modules/loader.go
```diff
@@ -0,0 +1,217 @@
...
+		// This can happen if the directory the module was downloaded to has been deleted and moved
+		// so the existing manifest.json is out-of-date.
+		_, diags := tfconfig.LoadModule(path.Join(m.Path, manifestModule.Dir))
+		if diags.HasErrors() {
+			log.Debugf("Module %s cannot be loaded, re-loading: %s", key, diags.Err())
+		} else {
+			return manifestModule, err
+		}
+	} else {
+		log.Debugf("Module %s needs loaded: %s", key, err.Error())
```
hugorut, line 137:
```suggestion
		log.Debugf("Module %s needs loading: %s", key, err.Error())
```

internal/hcl/modules/loader.go
```diff
@@ -0,0 +1,217 @@
...
+
+		// Test if we can actually load the module. If not, then we should try re-loading it.
+		// This can happen if the directory the module was downloaded to has been deleted and moved
+		// so the existing manifest.json is out-of-date.
+		_, diags := tfconfig.LoadModule(path.Join(m.Path, manifestModule.Dir))
+		if diags.HasErrors() {
+			log.Debugf("Module %s cannot be loaded, re-loading: %s", key, diags.Err())
+		} else {
+			return manifestModule, err
+		}
```
hugorut, line 135:
```suggestion
		if !diags.HasErrors() {
		        return manifestModule, err
		}
		
		log.Debugf("Module %s cannot be loaded, re-loading: %s", key, diags.Err())
```

internal/hcl/modules/loader.go
```diff
@@ -0,0 +1,217 @@
...
+
+	var lookupResult *RegistryLookupResult
+	var registryErr error
+	registrySource, registryErr := normalizeRegistrySource(moduleAddr)
+	if registryErr == nil {
+		lookupResult, registryErr = registryLoader.lookupModule(registrySource, moduleCall.Version)
+	}
+
+	if registryErr == nil {
+		log.Debugf("Downloading module %s from registry URL %s", key, lookupResult.DownloadURL)
```
hugorut, line 183:
```suggestion
	if registryErr == nil {
		lookupResult, registryErr = registryLoader.lookupModule(registrySource, moduleCall.Version)

		log.Debugf("Downloading module %s from registry URL %s", key, lookupResult.DownloadURL)
```

internal/hcl/modules/registry.go
```diff
@@ -0,0 +1,250 @@
...
+	dest           string
+}
+
+// NewRegistryLoader constructs a registry loader
+func NewRegistryLoader(packageFetcher *PackageFetcher, dest string) *RegistryLoader {
+	return &RegistryLoader{
+		packageFetcher: packageFetcher,
+		dest:           dest,
+	}
+}
```
hugorut, line 41: I'd change the `dest` to move to an input of the `downloadModule` method. Then we can have `RegistryLoader` as part of the top level `Loader` and not have to be constructed in a method.

hugorut: e.g.:

```go
        fetcher :=  NewPackageFetcher()
	return &ModuleLoader{
		Path:           path,
		cache:          NewCache(),
		packageFetcher: packageFetcher,
		registryLoader: RegistryLoader{packageFetcher: packageFetcher}
	}
```

internal/hcl/module.go
```diff
@@ -58,34 +56,8 @@ func (b *Block) getModuleBlocks(modulePath string) (Blocks, error) {
...
-	var metadata ModulesMetadata
-	if err := json.NewDecoder(f).Decode(&metadata); err != nil {
-		return nil, fmt.Errorf("could not decode metadata file %w", err)
-	}
+func loadModules(fullPath string) (*modules.Manifest, error) {
+	moduleLoader := modules.NewModuleLoader(fullPath)
 
-	return &metadata, nil
+	return moduleLoader.Load()
 }
```
hugorut, line 63: I'd probably remove this `loadModules` function and just set the `modules.Loader` on the `Parser` as a struct. Then move the `fullPath` to the `Load` method. And whilst we're doing this let's move the `initialPath` variable to a argument on `ParseDirectory`, doesn't make a lot of sense to have it at the top level.

```
func New(initialPath string, options ...Option) *Parser {
	p := &Parser{
		workspaceName: "default",
		moduleLoader: modules.NewModuleLoader(),
	}
        ...
}


func (parser *Parser) ParseDirectory(initialPath string) ([]*Module, error) {
      ....
      modulesMetadata, err := p.moduleLoader.Load(initialPath)
      ...
}
```

Easier for unit testing in the future

hugorut (commented):
looks good 🚀 🚀  💯  - just minor things with the base structure/interface of the structs. Small tweaks though.

My one big question is how this relates to the output. I know you're piping a lot of stuff to `log.Debug` but I do think we should have an info line or spinner for an initial load, as this could take a decent chunk of time. Maybe this spinner/log line only needs to appear if handling remote modules (as local ones will be fast)

internal/hcl/modules/loader.go
aliscott: TIL that this is called "infinitival copula deletion" which is common in Scotland and some other places, but not elsewhere.

internal/hcl/modules/loader.go
hugorut: ah you're right, I was confused with the shadowed `registryErr`

internal/hcl/module.go
aliscott: The `ModuleLoader` creates a cache which should be unique per initial path. Would it make sense to construct that each time `Load` is called as well if we're moving the path to be an argument of that function? My only concern with this is that it makes the code a bit messier having to pass the path and cache to each function in the `ModuleLoader`. But if I get rid of the `downloadDir` and `manifestFilePath` functions then it's probably alright.

aliscott:
> My one big question is how this relates to the output.
Yes, it's a question I have as well. I think this should have a spinner, but not sure what to do at the moment because it depends on how we expose it to the user initially. We could add an `EnableSpinner` flag to the `ModuleLoader`.

internal/hcl/module.go
hugorut: spoken about this over zoom and decision is that we leave as is as this the way Terraform does it with. Further docs to be added.

aliscott:
@hugorut can you re-review this please?

hugorut (approved):
looks good - just tested on the autoscaling_group_test and it works a charm ✅.

PR #1347, opened 2022-02-05
fix: solve issue where combined output has double diff monthly total

hugorut:
removes lines which were duplicated: https://github.com/infracost/infracost/blob/master/internal/output/combined.go#L140-L153


aliscott (approved):
Thanks! LGTM

PR #1348, opened 2022-02-07
Feat/add module calls

tim775:
@hugorut Adding the module calls blocks in does help.  There's still some work to do because it's not picking up the detailed monitoring costs.

### DIFF
--- a/internal/hcl/block.go
+++ b/internal/hcl/block.go
@@ -258,6 +258,44 @@ func (b *Block) HasModuleBlock() bool {
 	return b.moduleBlock != nil
 }
 
+// ModuleName returns the address of the module associated with this Block or "" if it is part of the root Module
+func (b *Block) ModuleAddress() string {
+	if b == nil || !b.HasModuleBlock() {
+		return ""
+	}
+
+	return b.moduleBlock.FullName()
+}
+
+// ModuleName returns the name of the module associated with this Block or "" if it is part of the root Module
+func (b *Block) ModuleName() string {
+	if b == nil || !b.HasModuleBlock() {
+		return ""
+	}
+
+	return b.moduleBlock.TypeLabel()
+}
+
+// ModuleSource returns the "source" attribute from the associated Module or "" if it is part of the root Module
+func (b *Block) ModuleSource() string {
+	if b == nil || !b.HasModuleBlock() {
+		return ""
+	}
+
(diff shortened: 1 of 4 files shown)


internal/hcl/block.go
```diff
@@ -258,6 +258,65 @@ func (b *Block) HasModuleBlock() bool {
...
+			return value.AsString()
+		}
+	}
+
+	// there's no explicit provider so get the provider implied as the prefix from the type
+	parts := strings.Split(b.TypeLabel(), "_")
+	if len(parts) > 0 {
+		return parts[0]
+	}
+	return ""
```
hugorut, line 317: I'm not sure this will work for all cases. It might be best to provide the top level provider from the module into this method. Or just return an empty string and then at the module level use the base `provider` block.

hugorut (commented):
Nice, minor thing about the `provider` logic - but I think the way you have done it at the moment will work right now. So no biggie

hugorut:
@tim775 part of the reason this isn't picking up the metrics is because monitoring is enabled by default in launch configuration: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_configuration#enable_monitoring

So we need to change this: 
https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/autoscaling_group.go#L97 to have true as the default, e.g. `GetBoolOrDefault` - that leave's one failing line in the test

internal/hcl/block.go
tim775: I hear you but I didn't exactly make this up 🙂  https://github.com/hashicorp/terraform/blob/d1ac8b71d4c1d6cfae4c250454a1c6f893a31231/internal/addrs/resource.go#L68-L77

hugorut: haha nice 👍

internal/hcl/block.go
```diff
@@ -258,6 +258,65 @@ func (b *Block) HasModuleBlock() bool {
 	return b.moduleBlock != nil
 }
 
+// ModuleName returns the address of the module associated with this Block or "" if it is part of the root Module
```
hugorut, line 261:
```suggestion
// ModuleAddress returns the address of the module associated with this Block or "" if it is part of the root Module
```

hugorut (approved):

PR #1351, opened 2022-02-08
feat: add sha checksum to binaries

hugorut:
Adds checksums to release artefacts.

Screenshot 2022-02-08 at 14 14 04

We then recommend users check these as part of install process, as outlined here: https://github.com/infracost/docs/pull/161

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -4,10 +4,39 @@ set -e
 
 os=$(uname | tr '[:upper:]' '[:lower:]')
 arch=$(uname -m | tr '[:upper:]' '[:lower:]' | sed -e s/x86_64/amd64/)
+if [ "$arch" = "aarch64" ]; then
+  arch="arm64"
+fi
+
+url="https://infracost.io/downloads/latest"
+tar="infracost-$os-$arch.tar.gz"
 echo "Downloading latest release of infracost-$os-$arch..."
-curl -sL https://github.com/infracost/infracost/releases/latest/download/infracost-$os-$arch.tar.gz | tar xz -C /tmp
+curl -sL $url/$tar -o /tmp/$tar
+echo
+
+code=$(curl -s -L -o /dev/null -w "%{http_code}" $url/$tar.sha256)
+if [ "$code" = "404" ]; then
+  echo "Checksum for infracost-$os-$arch release not found, skipping checksum validation"
+else
+  if [ -x "$(command -v shasum)" ]; then
+    echo "Validating checksum for infracost-$os-$arch..."
+    curl -sL $url/$tar.sha256 -o /tmp/$tar.sha256
+    (cd /tmp/;shasum -sqc $tar.sha256)
+    rm /tmp/$tar.sha256
+  else
+    echo "The shasum command could not be found, skipping checksum validation"
+  fi
+fi
+
+tar xzf /tmp/$tar -C /tmp
+rm /tmp/$tar
+
 echo
 echo "Moving /tmp/infracost-$os-$arch to /usr/local/bin/infracost (you might be asked for your password due to sudo)"
-sudo mv /tmp/infracost-$os-$arch /usr/local/bin/infracost
+if [ -x "$(command -v sudo)" ]; then
+  sudo mv /tmp/infracost-$os-$arch /usr/local/bin/infracost
+else
+  mv /tmp/infracost-$os-$arch /usr/local/bin/infracost
+fi
 echo
 echo "Completed installing $(infracost --version)"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -26,23 +26,29 @@ curl -sL "$url/$tar" -o "/tmp/$tar"
 echo
 
 code=$(curl -s -L -o /dev/null -w "%{http_code}" "$url/$tar.sha256")
-if [ "$code" != "404" ]; then
+if [ "$code" = "404" ]; then
+    echo "Skipping checksum validation as the sha for the release could not be found, no action needed."
+else
   if [ -x "$(command -v shasum)" ]; then
     echo "Validating checksum for infracost-$os-$arch..."
     curl -sL "$url/$tar.sha256" -o "/tmp/$tar.sha256"
 
     if ! check_sha "$tar.sha256"; then
       echo
-      echo "Installation checksum failed! This could be a security issue so please email hello@infracost.io for help."
-      exit 1
+      read -r -p "Installation checksum failed. This could be a security issue. Would you like to continue? (y/n) " answer
+      if [ "$answer" != "y" ]; then
+        echo
+        echo "Exiting, please email hello@infracost.io for help."
+        exit 1
+      fi
     fi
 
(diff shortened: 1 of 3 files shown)


scripts/install.sh
```diff
@@ -4,10 +4,39 @@ set -e
 
 os=$(uname | tr '[:upper:]' '[:lower:]')
 arch=$(uname -m | tr '[:upper:]' '[:lower:]' | sed -e s/x86_64/amd64/)
+if [ "$arch" = "aarch64" ]; then
+  arch="arm64"
```
hugorut, line 8: solves issue where `arrch64` architecture machines can't get a valid binary

scripts/install.sh
```diff
@@ -4,10 +4,39 @@ set -e
...
+rm /tmp/$tar
+
 echo
 echo "Moving /tmp/infracost-$os-$arch to /usr/local/bin/infracost (you might be asked for your password due to sudo)"
-sudo mv /tmp/infracost-$os-$arch /usr/local/bin/infracost
+if [ -x "$(command -v sudo)" ]; then
+  sudo mv /tmp/infracost-$os-$arch /usr/local/bin/infracost
+else
+  mv /tmp/infracost-$os-$arch /usr/local/bin/infracost
+fi
```
hugorut, line 40: I've removed the reliance on `sudo` as when I was testing this in `ubuntu` container I didn't actually have access to it. Just mv the binary to the local bin was fine perms wise.

hugorut:
@aliscott could you check the `install.sh` changes here: https://github.com/infracost/infracost/pull/1351/commits/ee868919e91cf5f51bf3d1d6eb4e877dfd8f874e as we discussed yesterday.

aliscott:
LGTM!

scripts/install.sh
```diff
@@ -4,10 +4,39 @@ set -e
...
+if [ "$code" = "404" ]; then
+  echo "Checksum for infracost-$os-$arch release not found, skipping checksum validation"
+else
+  if [ -x "$(command -v shasum)" ]; then
+    echo "Validating checksum for infracost-$os-$arch..."
+    curl -sL $url/$tar.sha256 -o /tmp/$tar.sha256
+    (cd /tmp/;shasum -sqc $tar.sha256)
+    rm /tmp/$tar.sha256
+  else
+    echo "The shasum command could not be found, skipping checksum validation"
```
alikhajeh1, line 27: I'm wondering if we should avoid logging here and on line 19 if there's a 404. In both cases, there's nothing the user can do (since the install proceeds anyway), so why warn them? The one we want to flag-up is when the checksum fails. What does that log look like? I'd even consider not logging if the checksum passes.

I say this based on the `brew install` experience, it's so painless/unalarming that I didn't even know it was checking the checksum.

scripts/install.sh
```diff
@@ -4,10 +4,39 @@ set -e
...
+  fi
+fi
+
+tar xzf /tmp/$tar -C /tmp
+rm /tmp/$tar
```
alikhajeh1, line 32: Should we get rid of the `rm` commands? it could alarm users if the `rm` asks for confirmation, and we're deleting stuff in /tmp so it doesn't matter as it'll get cleaned up eventually if storage is low.

alikhajeh1 (commented):
Good stuff!

alikhajeh1:
it might be worth checking https://www.shellcheck.net (I use to use that on the old diff script). Also, I suppose we should merge this after the next release as the docs repo uses `curl -fsSL https://raw.githubusercontent.com/infracost/infracost/master/scripts/install.sh | sh` so if it goes into master it gets used.

scripts/install.sh
hugorut: I don't have a strong opinion of this - however, I remember seeing a brew issue (can't find it at the moment) where someone mentioned they wanted to have a warning if the cask had disabled `sha256`. Which suggests that we might want to be as verbose as possible. 

The warning message when checksum fails is as follows:

```
infracost-darwin-amd64.tar.gz: FAILED
shasum: WARNING: 1 computed checksum did NOT match
```

This is directly from the shasum command. We could capture output and modify message if you want.

scripts/install.sh
hugorut: Personally I'd prefer scripts to cleanup artefacts they created. It just seems good practice.

hugorut:
> Also, I suppose we should merge this after the next release as the docs repo uses `curl -fsSL https://raw.githubusercontent.com/infracost/infracost/master/scripts/install.sh | sh` so if it goes into master it gets used. 

This is safe to go into master at the moment - hence `404` checking. But we can wait until the release.

scripts/install.sh
alikhajeh1: I'd optimize this flow for getting to success as soon as possible. The 404 log doesn't have any actionable things for the user to do, so I'd probably skip that so we can merge this PR. The missing shasum command gives me a bit of pause - maybe "Skipping checksum validation as the shasum command could not be found, no action needed." is clearer that we don't want the user to do anything, this is just FYI.

Re the failure log, maybe it could say "Installation checksum failed! This could be a security issue so please email hello@infracost.io for help."
For now, I guess it's ok to continue with the installation even if the checksum failed until we revisit this side of things next month.

scripts/install.sh
alikhajeh1: tru dat! Overall it seems like we want this script to be non-interactive as people are copying that curl into their workflows, so if there is no risk of `rm` throwing a prompt then it's safe to ship.

hugorut:
@alikhajeh1 I've updated the script following your feedback, could you please take a look at it again:

1. updated to fail silently on 404
2. updated shasum not found messaging
3. updated failure messaging if shasum check fails
4. fixed shell linting quotes

**if shasum fails:**
Screenshot 2022-02-10 at 09 19 47

**If shasum succeeds:**
Screenshot 2022-02-10 at 09 20 10

**if 404:**
Screenshot 2022-02-10 at 09 20 26

**If shasum not exists:**
Screenshot 2022-02-10 at 09 26 52

---


> For now, I guess it's ok to continue with the installation even if the checksum failed until we revisit this side of things next month. 

I'm hesitant to do this as I think failure of checksum should be force stop for users as this could be a critical security issue. So we explicitly exit if the shasum fails. @aliscott what are your thoughts?

aliscott:
Here's my thoughts, but these aren't strong opinions.

1. Skipping checksum validation because of 404

I think we should show a similar message here as the `shasum could not be found` error. It seems inconsistent not to and I think we should be explicit.

2. Failure on mismatch checksum

One option would be to have a prompt like 'The checksum does not match. Would you like to continue?'. If we think most users would use this in CLI instead of CI then I think having one is fine, we potentially prompt anyway for `sudo`.

hugorut:
@aliscott I've added the 404 check back in with https://github.com/infracost/infracost/pull/1351/commits/b653e35ffb4360b8ff4683bbef616113c3d9a415 and added a prompt for shasum failure

PR #1355, opened 2022-02-09
feat: Add support for Premium Azure Event Hubs

### LINKED ISSUE #1354: Cost diff discrepancy on Premium Azure Event Hub service
### Cost diff discrepancy on Azure Event Hub service
Azure Event Hub change Standard to Premium SKU and Infracost shows decrease in price, pretty sure the increase is actually quite big. Result of diff:
```
Detected Terragrunt directory at workloads/development/australiaeast/eventhub
  ✔ Running terragrunt run-all terragrunt-info 
  ✔ Running terragrunt run-all plan 
  ✔ Running terragrunt show 
  ✔ Extracting only cost-related params from terragrunt plan 
  ✔ Retrieving cloud prices to calculate costs 

Project: gysiedebruyn/terraform-azure/workloads/development/australiaeast/eventhub

~ azurerm_eventhub_namespace.ns
  -$21.90 ($21.90 → $0.00)

    - Ingress event (Standard)
      Monthly cost depends on usage
        -$0.028 per 1M events

    ~ Throughput
      -$21.90 ($21.90 → $0.00)

Monthly cost change for gysiedebruyn/terraform-azure/workloads/development/australiaeast/eventhub
Amount:  -$21.90 ($29.70 → $7.80)
Percent: -74%

Key: ~ changed, + added, - removed

4 cloud resources were detected, rerun with --show-skipped to see details:
∙ 3 were estimated, 2 include usage-based costs, see https://infracost.io/usage-file
∙ 1 was free
```

This is the actual Terraform plan output:
```
  # azurerm_eventhub_namespace.ns must be replaced
-/+ resource "azurerm_eventhub_namespace" "ns" {
      ~ default_primary_connection_string         = (sensitive value)
      + default_primary_connection_string_alias   = (sensitive value)
      ~ default_primary_key                       = (sensitive value)
      ~ default_secondary_connection_string       = (sensitive value)
      + default_secondary_connection_string_alias = (sensitive value)
      ~ default_secondary_key                     = (sensitive value)
      ~ id                                        = "/.../jdb-tf-standard-hub" -> (known after apply)
      ~ maximum_throughput_units                  = 0 -> (known after apply)
        name                                      = "jdb-tf-standard-hub"
      ~ network_rulesets                          = [
          ~ {
              ~ virtual_network_rule           = [] -> null
                # (3 unchanged elements hidden)
            },
        ]
      ~ sku                                       = "Standard" -> "Premium" # forces replacement
        tags                                      = {
            "CostCentre" = "PoC"
            "Module"     = "eventhub"
            "Owner"      = "Jan de Bruyn"
        }
      ~ zone_redundant                            = false -> true # forces replacement
        # (4 unchanged attributes hidden)
    }
```

_Originally posted by @gysiedebruyn in https://github.com/infracost/infracost/discussions/1271#discussioncomment-2137601_

tim775:
Fixes #1354

### DIFF
--- a/internal/providers/terraform/azure/event_hubs_namespace.go
+++ b/internal/providers/terraform/azure/event_hubs_namespace.go
@@ -109,14 +109,11 @@ func eventHubsIngressCostComponent(region, sku string, quantity *decimal.Decimal
 }
 
 func eventHubsThroughPutCostComponent(region, sku, meterName string, capacity decimal.Decimal) *schema.CostComponent {
-	unitName := meterName
-	if !strings.HasPrefix(meterName, sku) {
-		unitName = sku + " " + meterName
-	}
+	unitName := strings.TrimPrefix(strings.ToLower(meterName), strings.ToLower(sku+" ")) + "s"
 
 	return &schema.CostComponent{
-		Name:           "Capacity",
-		Unit:           strings.ToLower(unitName) + "s",
+		Name:           fmt.Sprintf("Capacity (%s)", sku),
+		Unit:           unitName,
 		UnitMultiplier: schema.HourToMonthUnitMultiplier,
 		HourlyQuantity: decimalPtr(capacity),
 		ProductFilter: &schema.ProductFilter{
--- a/internal/providers/terraform/azure/event_hubs_namespace.go
+++ b/internal/providers/terraform/azure/event_hubs_namespace.go
@@ -19,8 +19,9 @@ func GetAzureRMEventHubsNamespaceRegistryItem() *schema.RegistryItem {
 
 func NewAzureRMEventHubs(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
 	var events, unknown *decimal.Decimal
+	var includedRetention decimal.Decimal
 	sku := "Basic"
-	meterName := "Throughput Unit"
+	meterName := ""
 	region := lookupRegion(d, []string{})
 
 	if d.Get("sku").Type != gjson.Null {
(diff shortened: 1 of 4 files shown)


aliscott (approved):
LGTM. Two questions:

1. Should we use these cost components names and units to map to the units here? https://azure.microsoft.com/en-gb/pricing/details/event-hubs/

```
Capacity (Basic): throughput unit
Capacity (Standard): throughput unit
Capacity (Premium): processing unit
Capacity (Dedicated): capacity unit
```

2. The Azure pricing calculator has a user input for 'Extended Retention', and not total storage retention. I'm wondering if we want the same instead of including the included storage retention in the number? It's what we do for other resources that have `additional_backup_storage_gb`. I don't feel strongly about this though so open to either option.

tim775:
> 1. Should we use these cost components names and units to map to the units here?

Yeah that's better

> 2. The Azure pricing calculator has a user input for 'Extended Retention', and not total storage retention...

Hmmm... I think since this already exists for the dedicated sku I'm inclided to leave it as is.  

I'm not sure which one is the right way to go -- if we start going deeper into the usage estimation stuff we may decide that it's better to put the total in the usage file and let the calculator decide what the difference is.  I think the main reason we don't is because of free tier stuff, but maybe even that we should just have a way to specify is_free_tier in the usage file.

aliscott:
> Hmmm... I think since this already exists for the dedicated sku I'm inclided to leave it as is.

👍 makes sense

PR #1363, opened 2022-02-09
feat: Add commenting for Bitbucket

vdmgolub:


cmd/infracost/testdata/comment_bitbucket_help/comment_bitbucket_help.golden
```diff
@@ -0,0 +1,32 @@
...
+                                        delete-and-new    Delete previous matching comments and create a new comment (default "update")
+      --bitbucket-server-url string   Bitbucket Server URL
+      --bitbucket-token string        Bitbucket access token
+      --commit string                 Commit SHA to post comment on, mutually exclusive with pull-request. Not available when bitbucket-server-url is set
+      --dry-run                       Generate comment without actually posting to Azure Repos
+  -h, --help                          help for bitbucket
+  -p, --path stringArray              Path to Infracost JSON files, glob patterns need quotes
+      --pull-request int              Pull request number to post comment on
+      --repo string                   Repository in format owner/repo
+      --tag string                    Customize text used to detect comments posted by Infracost. It is placed at the bottom of a comment
```
vdmgolub, line 28: @alikhajeh1 Please review the copy, specifically `--commit` and `--tag` ones. Should we also clarify anyhow the `--bitbucket-token` that it needs `user:password` combination for Cloud and HTTP Access Token for Server? We'll explain that in documentation too.

internal/output/templates.go
```diff
@@ -365,3 +356,56 @@ This comment will be replaced when the cost estimate changes
...
+{{- end }}
+{{- if .MarkdownOptions.WillReplace }}
+
+This comment will be replaced when the cost estimate changes
+{{- end }}
+{{- if .MarkdownOptions.IncludeFeedbackLink }}
+
+Is this comment useful? [Yes](https://www.infracost.io/feedback/submit/?value=yes), [No](https://www.infracost.io/feedback/submit/?value=no)
+{{- end }}
+`
```
vdmgolub, line 411: @alikhajeh1 Here's the Bitbucket comment structure. Does it look okay?

cmd/infracost/comment_bitbucket.go
```diff
@@ -0,0 +1,133 @@
...
+	cmd.Flags().String("bitbucket-token", "", "Bitbucket access token")
+	_ = cmd.MarkFlagRequired("bitbucket-token")
+	cmd.Flags().String("commit", "", "Commit SHA to post comment on, mutually exclusive with pull-request. Not available when bitbucket-server-url is set")
+	cmd.Flags().StringArrayP("path", "p", []string{}, "Path to Infracost JSON files, glob patterns need quotes")
+	_ = cmd.MarkFlagRequired("path")
+	_ = cmd.MarkFlagFilename("path", "json")
+	cmd.Flags().Int("pull-request", 0, "Pull request number to post comment on")
+	cmd.Flags().String("repo", "", "Repository in format owner/repo")
+	_ = cmd.MarkFlagRequired("repo")
+	cmd.Flags().String("tag", "", "Customize text used to detect comments posted by Infracost. It is placed at the bottom of a comment")
```
alikhajeh1, line 129:
```suggestion
	cmd.Flags().String("tag", "", "Customize special text used to detect comments posted by Infracost (placed at the bottom of a comment).")
```

cmd/infracost/comment_bitbucket.go
```diff
@@ -0,0 +1,133 @@
...
+	cmd.Flags().Bool("dry-run", false, "Generate comment without actually posting to Azure Repos")
```
alikhajeh1, line 130:
```suggestion
	cmd.Flags().Bool("dry-run", false, "Generate comment without actually posting to Bitbucket")
```

cmd/infracost/comment_bitbucket.go
```diff
@@ -0,0 +1,133 @@
...
+
+	cmd.Flags().String("behavior", "update", `Behavior when posting comment, one of:
+  update (default)  Update latest comment
+  new               Create a new comment
+  delete-and-new    Delete previous matching comments and create a new comment`)
+	_ = cmd.RegisterFlagCompletionFunc("behavior", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return validCommentBitbucketBehaviors, cobra.ShellCompDirectiveDefault
+	})
+	cmd.Flags().String("bitbucket-server-url", "", "Bitbucket Server URL")
+	cmd.Flags().String("bitbucket-token", "", "Bitbucket access token")
```
alikhajeh1, line 120:
```suggestion
	cmd.Flags().String("bitbucket-token", "", "Bitbucket access token. Use 'user:password' for Bitbucket Cloud and HTTP access token for Bitbucket Server.")
```

Is this right? We could also link to https://support.atlassian.com/bitbucket-cloud/docs/app-passwords/ in the docs repo and the Bitbucket readme

cmd/infracost/comment_bitbucket.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	cmd.Flags().String("behavior", "update", `Behavior when posting comment, one of:
+  update (default)  Update latest comment
+  new               Create a new comment
+  delete-and-new    Delete previous matching comments and create a new comment`)
+	_ = cmd.RegisterFlagCompletionFunc("behavior", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return validCommentBitbucketBehaviors, cobra.ShellCompDirectiveDefault
+	})
+	cmd.Flags().String("bitbucket-server-url", "", "Bitbucket Server URL")
```
alikhajeh1, line 119: Does the URL default to something? Maybe that's defined elsewhere?

cmd/infracost/testdata/comment_bitbucket_help/comment_bitbucket_help.golden
alikhajeh1: Done.

internal/output/templates.go
alikhajeh1: Do you have a screenshot or a link to an example comment?

alikhajeh1 (commented):
This is looking really good! I added my comments

cmd/infracost/comment_bitbucket.go
vdmgolub: Yep, `password` is the App password, maybe we can show it as `user:app-password`.

cmd/infracost/comment_bitbucket.go
vdmgolub: Bitbucket Cloud is `bitbucket.org`, so I use it in the code to specify API endpoints (the hostname is hardcoded here: https://github.com/infracost/infracost/pull/1363/files#diff-b209df5ca9c5860317f0907a53d70e6889cbd10f3e352ae6bd8674becb38fe75R144). If this flag is specified though, it tells that users are using Bitbucket Server.

internal/output/templates.go
aliscott: We could default tag to "", and if they set it then the last line could be _(generated by Infracost, tag: X)_.

tim775 (approved):

cmd/infracost/comment_bitbucket.go
alikhajeh1: Sorry what I meant is, why does the help text for github/lab show the default but this one doesn't? it would be nice if it could show the default but if that's not possible to do automatically, then changing the text to `Bitbucket Server URL, defaults to https://bitbucket.org` is enough. I think the text should include an example so the user knows if they should put https://, or any forward slashes at the end etc

```
      --github-api-url string   GitHub API URL (default "https://api.github.com")
      --gitlab-server-url string   GitLab Server URL (default "https://gitlab.com")
      --bitbucket-server-url string   Bitbucket Server URL
```

cmd/infracost/comment_bitbucket.go
alikhajeh1: good idea, I updated my original suggestion

internal/output/templates.go
alikhajeh1: Looks great! I'd keep it as is since there is a use-case for matrix builds wanting to customize the tag.

internal/output/templates.go
```diff
@@ -365,3 +356,56 @@ This comment will be replaced when the cost estimate changes
 </sub>
 {{- end}}
 `
+
+var CommentMarkdownTemplate = `
```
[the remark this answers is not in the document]
vdmgolub: I had to introduce `BasicSyntax: true/false` markdown option for that. Which means that default always includes HTML bits now.

Is its purpose clear for you from the name? Or maybe we could rename it to something better.

internal/output/templates.go
vdmgolub: @aliscott Great idea! I'll do that, thanks! 👍

internal/output/templates.go
```diff
@@ -288,36 +288,35 @@ iVBORw0KGgoAAAANSUhEUgAAAMAAAADACAMAAABlApw1AAAABGdBTUEAALGPC/xhBQAAAAFzUkdCAK7O
...
-var CommentMarkdownTemplate = `
+var CommentMarkdownWithHTMLTemplate = `
 {{- define "summaryRow"}}
-    <tr>
-      <td>{{ truncateMiddle .Name 64 "..." }}</td>
-      <td align="right">{{ formatCost .PastCost }}</td>
-      <td align="right">{{ formatCost .Cost }}</td>
-      <td>{{ formatCostChange .PastCost .Cost }}</td>
-    </tr>
+  <tr>
```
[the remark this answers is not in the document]
vdmgolub: Aaah, I see what you mean, the tabs were inside `tbody`. Yep, I indented it based on the `define` block, not the table. Thanks for catching, I'll update 👍

internal/output/templates.go
vdmgolub: Updated: https://github.com/infracost/infracost/pull/1363/files#diff-b209df5ca9c5860317f0907a53d70e6889cbd10f3e352ae6bd8674becb38fe75R162

internal/output/templates.go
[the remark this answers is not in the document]
vdmgolub: Reverted the changes 👍

cmd/infracost/comment_bitbucket.go
alikhajeh1: Thanks!

PR #1369, opened 2022-02-10
feat: Add support for Amazon RDS Performance Insights charges.

### LINKED ISSUE #1195: No price change when changing `db_instance.performance_insights_retention_period`
Hey! I just noticed today that infracost doesn't report a price change when changing `db_instance.performance_insights_retention_period`, I would expect it to pick up the change between `7` and `731` and report a price change according to https://aws.amazon.com/rds/performance-insights/pricing/

Here's a link to the slack thread I opened earlier: https://infracost-community.slack.com/archives/C01G63Q4HS7/p1638969459121200

Let me know if you need more info!

I've attached a screenshot of the report from GitHub actions:

infracost-price-change

tim775:
Fixes #1195


internal/resources/aws/db_instance.go
```diff
@@ -231,9 +235,70 @@ func (r *DBInstance) BuildResource() *schema.Resource {
...
+			Service:       strPtr("AmazonRDS"),
+			ProductFamily: strPtr("Performance Insights"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "usagetype", ValueRegex: regexPtr("PI_LTR:" + strings.ToUpper(instanceFamily) + "$")},
+			},
+		},
+	}
+}
+
+func PerformanceInsightsAPIRequestCostComponent(region string, additionalRequests *int64) *schema.CostComponent {
```
vdmgolub, line 288: Both resources have access to them within one package, and editor can help navigating to the definition, but I wonder if these two functions should be placed in a dedicated file or something, like `s3_storage_class_helpers.go` does. @hugorut, maybe you know Go's any good practices on the code placement?

internal/resources/aws/db_instance.go
```diff
@@ -231,9 +235,70 @@ func (r *DBInstance) BuildResource() *schema.Resource {
...
 
+	if r.PerformanceInsightsEnabled {
+		if r.PerformanceInsightsLongTermRetention {
+			costComponents = append(costComponents, PerformanceInsightsLongTermRetentionCostComponent(r.Region, r.InstanceClass))
+		}
+
+		if r.MonthlyAdditionalPerformanceInsightsRequests == nil || *r.MonthlyAdditionalPerformanceInsightsRequests > 0 {
+			costComponents = append(costComponents,
+				PerformanceInsightsAPIRequestCostComponent(r.Region, r.MonthlyAdditionalPerformanceInsightsRequests))
+		}
```
vdmgolub, line 246: Just a thought, this already looks good. As alternative, we could move the conditions inside the cost component function and return an empty list. But this way it's more explicit, I guess.
```go
costComponents = append(costComponents, PerformanceInsightsAPIRequestCostComponent(r.Region, r.MonthlyAdditionalPerformanceInsightsRequests)...)
```
```go
func PerformanceInsightsAPIRequestCostComponent(region string, additionalRequests *int64) *schema.CostComponent {
	costComponents := []*schema.CostComponent{}
	if additionalRequests == nil || additionalRequests > 0 {
		return costComponents
	}
	// ...
}
```

vdmgolub (approved):

internal/resources/aws/db_instance.go
hugorut: I think it's fine to have it in this file. I would mark it as un-exported though e.g. `performanceInsightsAPIRequestCostComponent` as it's not consumed outside this package. I think there's only real value to extracting to a dedicated file when there's a ton more specific `s3_storage` helper functions. If there are some lying around already, then maybe we can.

internal/resources/aws/db_instance.go
tim775: Yeah, I thought about that.  It seemed like it would be better to have the answer to the "When does this component actually get added?" question in resource itself 🤷‍♂️

PR #1370, opened 2022-02-11
fix(aws): Update tests to support terraform-provider-aws v4

vdmgolub:


internal/providers/terraform/aws/s3_bucket_lifecycle_configuration.go
```diff
@@ -0,0 +1,69 @@
...
+
+func getS3BucketLifecycleConfigurationRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "aws_s3_bucket_lifecycle_configuration",
+		RFunc: newS3BucketLifecycleConfigurationResource,
+	}
+}
+
+func newS3BucketLifecycleConfigurationResource(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	storageClassNames := map[string]string{
```
vdmgolub, line 17: I kept standard class on the S3 bucket: I couldn't find a way to detect the provider version, and standard class is added on the resource by default, I couldn't add any conditions on that. It looks weird in the output though, so maybe you have any better ideas? When we decide to stop supporting v3, we'll need to remove all that code from S3 bucket resource, along with moving that standard class here.

internal/providers/terraform/dir_provider.go
```diff
@@ -354,7 +354,7 @@ func (p *DirProvider) runInit(opts *CmdOptions, spinner *ui.Spinner) error {
 	if p.IsTerragrunt {
 		args = append(args, "run-all", "--terragrunt-ignore-external-dependencies")
 	}
-	args = append(args, "init", "-input=false", "-no-color")
+	args = append(args, "init", "-input=false", "-no-color", "-upgrade")
```
vdmgolub, line 357: Had to add this flag, otherwise the locked v4 is used for v3 tests.

infracost-usage-example.yml
```diff
@@ -272,15 +272,70 @@ resource_usage:
...
+  aws_s3_bucket.my_bucket:
+    object_tags: 10000000 # Total object tags. Only for AWS provider V3.
     standard: # Usages of S3 Standard:
       storage_gb: 10000 # Total storage in GB.
       monthly_tier_1_requests: 1000000 # Monthly PUT, COPY, POST, LIST requests (Tier 1).
       monthly_tier_2_requests: 100000 # Monthly GET, SELECT, and all other requests (Tier 2).
       monthly_select_data_scanned_gb: 10000 # Monthly data scanned by S3 Select in GB.
       monthly_select_data_returned_gb: 1000 # Monthly data returned by S3 Select in GB.
-    intelligent_tiering: # Usages of S3 Intelligent - Tiering:
+    intelligent_tiering: # Usages of S3 Intelligent - Tiering (Only for AWS provider V3):
```
aliscott, line 338: I think you should still be able to model different storage classes using the usage file for v4. We shouldn't show the prices by default but if a user is using that storage class with the bucket and they add usage, we should show it since you don't need top use lifecycle policies to use different storage classes.

internal/providers/terraform/aws/s3_bucket_lifecycle_configuration.go
aliscott: Should we have it on both? An S3 bucket doesn't need a lifecycle configuration so the costs might be on either.

internal/providers/terraform/dir_provider.go
aliscott: We might want to do this conditionally just for the tests. I don't think we should be upgrading the user's modules by default.

aliscott (commented):
Added some initial comments

internal/providers/terraform/dir_provider.go
vdmgolub: Yep, you're right. I'll add it when `IsTest()` is true only. 👍

infracost-usage-example.yml
vdmgolub: Do I understand correctly the logic?
1. If there are rules on the TF resource - it's v3 and we need to use them for cost components.
2. If there are no any rules, but the usage is present - it's _maybe_ v4 and we can use usage params to show the cost components? <- I see an issue for v3 here when there are no rules, but there are usage params.

internal/providers/terraform/aws/s3_bucket_lifecycle_configuration.go
aliscott: Since they're all usage-based costs would it not just show the prices unless you add usage data? And so it would only show costs against the resource you added the usage data to.

vdmgolub: Yep, it's up to the user to decide where to add the usage params. Sounds good 👍

infracost-usage-example.yml
vdmgolub: Sounds good, I'll update the PR. Thanks! 👍

internal/providers/terraform/aws/testdata/s3_bucket_test/s3_bucket_test.golden
```diff
@@ -2,70 +2,14 @@
...
-    ├─ Retrieval requests (standard)        Monthly cost depends on usage: $0.10 per 1k requests    
-    ├─ Retrievals (standard)                Monthly cost depends on usage: $0.02 per GB             
-    ├─ Retrieval requests (bulk)            Monthly cost depends on usage: $0.025 per 1k requests   
-    ├─ Retrievals (bulk)                    Monthly cost depends on usage: $0.0025 per GB           
-    └─ Early delete (within 180 days)       Monthly cost depends on usage: $0.00099 per GB          
+    ├─ Select data scanned                  Monthly cost depends on usage: $0.002 per GB            
+    └─ Select data returned                 Monthly cost depends on usage: $0.0007 per GB           
                                                                                                     
  aws_s3_bucket.bucket_withUsage                                                                     
- ├─ Object tagging                                        0.1  10k tags                       $0.00
```
vdmgolub, line 68: @aliscott You were right, the default behavior is to use either usage params or lifecycle options, so I haven't changed anything in the logic and it just works™. However, the Object tagging cost component relies on `lifecycle_rule` blocks, so it's not shown even when the usage param is set.

internal/providers/terraform/aws/s3_bucket_lifecycle_configuration.go
vdmgolub: Fixed in https://github.com/infracost/infracost/pull/1370/commits/addefa9c6486951666d492d08fb03c4de96d1a05. Will fixup after code review.

internal/providers/terraform/dir_provider.go
vdmgolub: Fixed in https://github.com/infracost/infracost/pull/1370/commits/9f7fc048cf65ad74a6c458e4d1d355f387c5031f.

infracost-usage-example.yml
vdmgolub: I misunderstood how the code works, it's already exactly how you described 😅. Thank you! Updated the tests in https://github.com/infracost/infracost/pull/1370/commits/3557f33f7fea344c6b291568c574898de87e8c8b.

infracost-usage-example.yml
```diff
@@ -272,15 +272,76 @@ resource_usage:
...
```
aliscott, line 344: Should the `(Only for AWS provider V3)` comments be removed?

aliscott (approved):
LGTM

PR #1376, opened 2022-02-12
refactor(google): resources migration batch 9

sinabakh:


sinabakh:
@aliscott Would you please review?

internal/resources/google/network_egress_internal.go
```diff
@@ -0,0 +1,306 @@
...
+	{ValueType: schema.Float64, DefaultValue: 0, Key: "china"},
+	{ValueType: schema.Float64, DefaultValue: 0, Key: "australia"},
+}
+
+type egressRegionData struct {
+	gRegion             string
+	apiDescription      string
+	apiDescriptionRegex string
+	usageKey            string
+	fixedRegion         string
```
aliscott, line 50: Although this isn't related to these changes, it might be good to document what `gRegion` and `fixedRegion` are. It's not immediately clear to me looking through the code.

internal/resources/google/network_egress_internal.go
```diff
@@ -0,0 +1,306 @@
...
+
+	for _, regData := range regionsData {
+		gRegion := regData.gRegion
+		usageKey := regData.usageKey
+
+		// TODO: Reformat to use tier helpers.
+		var usage float64
+		var used float64
+		var lastEndUsageAmount float64
+		usage = GetFloatFieldValueByUsageTag(usageKey, *r)
```
aliscott, line 98:
```suggestion
		var used float64
		var lastEndUsageAmount float64
		usage := GetFloatFieldValueByUsageTag(usageKey, *r)
```

internal/resources/google/network_egress_internal.go
```diff
@@ -0,0 +1,306 @@
...
+	case ComputeExternalVPNGateway, ComputeVPNGateway:
+		return false
+	default:
+		return true
+	}
+}
+
+func (r *NetworkEgressUsage) getEgressRegionsData(prefixName string, egressResourceType EgressResourceType) []*egressRegionData {
+	switch egressResourceType {
+	case ComputeVPNGateway:
```
aliscott, line 172: Now that we have the `NetworkEgressUsage` struct I'm wondering if it enables different approaches for handling the different egress resource types.

One option could be to use composition to extend the NetworkEgressUsage type, e.g:

```go
type ComputeVPNGatewayNetworkEgressUsage struct {
  NetworkEgressUsage
}

func (r *ComputeVPNGatewayNetworkEgressUsage) getEgressRegionsData...
```

Another option could be to have an interface for these and pass it to this struct instead of `egressResourceType`, e.g:

```go
type EgressResource interface {
  includesSameContinent() bool
  getEgressRegionsData []*EgressRegionData
}
```

What do you think?

aliscott (commented):
LGTM

sinabakh:
@aliscott Ok so your comments started a rabbit hole of refactoring :))
I want to give a summary of the changes and their reason.

## A. Why the previous structure was wrong
In the previous refactoring, we had a single struct named `NetworkEgress` and all usage data were its attributes. The problem was that different resources have different usage keys. For instance, `VPNGateway` has no `same_continent` usage key. So we should probably have different structs for different types of resources that use egress.

## B. Why composition didn't work
Each egress resource has about 5-6 usage keys. To make working with them easier I use reflect to get their values so I can iterate over them. The helper function is called `GetFloatFieldValueByUsageTag`.
Let's say we have a core struct named `NetworkEgress` and another one that encapsulates it named `ComputeVPNGatewayNetworkEgressUsage`. Since the core logic is placed on the `NetworkEgress` it can't access its parent fields unless the fields are defined on the core struct which then the problem `A` will arise.

So I had to split most of the code and introduce some duplication. **I'm not sure if that's the best or even the right way so I would be more than happy to get your ideas.**

internal/resources/google/network_egress_internal.go
```diff
@@ -0,0 +1,508 @@
...
+		usageKey := regData.usageKey
+		usage := GetFloatFieldValueByUsageTag(usageKey, *r)
+		newCostComponents := stepPricingHelper(usage, usageFiltersData, regData, defaultAPIRegionName, serviceName)
+		resource.CostComponents = append(resource.CostComponents, newCostComponents...)
+	}
+
+	return resource
+}
+
+func stepPricingHelper(usage float64, usageFiltersData []*egressRegionUsageFilterData, regData *egressRegionData, defaultAPIRegionName, serviceName *string) []*schema.CostComponent {
```
aliscott, line 227:
```suggestion
func egressStepPricingHelper(usage float64, usageFiltersData []*egressRegionUsageFilterData, regData *egressRegionData, defaultAPIRegionName, serviceName string) []*schema.CostComponent {
```

1. I'd make the name more explicit since the function is global in this package
2. I'd suggest making making the region and service strings, since we're just wrapping them in a ptr to pass them in here which seems pointless.

internal/resources/google/util.go
```diff
@@ -153,3 +154,20 @@ func (r RegionsUsage) Values() []RegionUsage {
...
+	if rt.Kind() != reflect.Struct {
+		return 0
+	}
+	for i := 0; i < rt.NumField(); i++ {
+		f := rt.Field(i)
+		v := strings.Split(f.Tag.Get("infracost_usage"), ",")[0] // use split to ignore tag "options" like omitempty, etc.
+		if v == tagValue {
+			r := reflect.ValueOf(s)
+			field := reflect.Indirect(r).FieldByName(f.Name)
+			return field.Elem().Float()
```
aliscott, line 169: Can `field.Elem()` be nil? We might want to check that just in case.

internal/resources/google/compute_external_vpn_gateway.go
```diff
@@ -0,0 +1,39 @@
...
+	},
+}
+
+func (r *ComputeExternalVPNGateway) PopulateUsage(u *schema.UsageData) {
+	resources.PopulateArgsWithUsage(r, u)
+}
+
+func (r *ComputeExternalVPNGateway) BuildResource() *schema.Resource {
+	region := r.Region
+	r.MonthlyEgressDataTransferGB.Region = region
```
aliscott, line 30: Do we need a check for this being nil, e.g.:

```go
if r.MonthlyEgressDataTransferGB == nil {
        r.MonthlyEgressDataTransferGB = &computeExternalVPNGatewayNetworkEgressUsage{}
}
```

aliscott (commented):
This looks good. I think separating it out makes it more explicit even if it is duplicated. One thought I had was if we should inline the usage subresources in each resource. I've written out an example here since it was easier to express as code: https://gist.github.com/aliscott/6d7f8973f29f2d4a5b1dabda3bd4dc7d. I feel that putting it inline means less jumping around when looking at the code.

Other suggestions I have:

1. Remove the `getEgressAPIRegionName` and `getEgressAPIServiceName` functions if we're just inlining these
2. Same with the `Prefix` and `Address` properties of the usage struct. And if we remove them we don't need the `NetworkEgressUsage` struct anymore, we can just inline `Region` where we need it.

Let me know what you think.

sinabakh:
> This looks good. I think separating it out makes it more explicit even if it is duplicated. One thought I had was if we should inline the usage subresources in each resource. I've written out an example here since it was easier to express as code: https://gist.github.com/aliscott/6d7f8973f29f2d4a5b1dabda3bd4dc7d. I feel that putting it inline means less jumping around when looking at the code.
> 
> Other suggestions I have:
> 
>     1. Remove the `getEgressAPIRegionName` and `getEgressAPIServiceName` functions if we're just inlining these
> 
>     2. Same with the `Prefix` and `Address` properties of the usage struct. And if we remove them we don't need the `NetworkEgressUsage` struct anymore, we can just inline `Region` where we need it.
> 
> 
> Let me know what you think.

I'm a big fan of making things simple so I completely agree with you.

internal/resources/google/util.go
sinabakh: Checks on previous parts makes it logically impossible for it to panic.

internal/resources/google/network_egress_internal.go
sinabakh: Makes total sense

internal/resources/google/compute_external_vpn_gateway.go
sinabakh: I think the call to `PopulateArgsWithUsage` will take care of initializing it.

vdmgolub:
@sinabakh Just FYI in case you plan on the next batch, I'm working on migrating `container_cluster`/`container_node_pool` Google resources and some other related to them.

aliscott (approved):
LGTM thanks Sina!

PR #1377, opened 2022-02-15
feat: add hcl variable inputs from terraform plan flags

hugorut:
Adds variable file and flag support for hcl, through:

* parsing `terraform-plan-flags`
* `var` and `var-file` flags
* parsing default `terraform.tfvars` file and `.auto.tfvars` files

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -60,8 +60,9 @@ func addRunFlags(cmd *cobra.Command) {
 
 	cmd.Flags().Bool("sync-usage-file", false, "Sync usage-file with missing resources, needs usage-file too (experimental)")
 
-	cmd.Flags().Bool("hcl-only", false, "Run infracost without terraform, parsing hcl files at the given path. This option does not need credentials and is much quicker to run.")
-	cmd.Flags().StringSlice("tfvar-files", nil, "Path to .tfvars file, can be used multiple times and evaluated in order of specification")
+	cmd.Flags().Bool("hcl-only", false, "Run infracost without terraform, parsing hcl files at the given path. This option does not need credentials and is much quicker to run. (experimental)")
+	cmd.Flags().StringSlice("var-file", nil, "Can only be used with --hcl-only. Load variable values from the given file, in addition\nto the default files terraform.tfvars and *.auto.tfvars. Use this option more than once to\ninclude more than one variables file. (experimental)")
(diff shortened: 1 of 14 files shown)


cmd/infracost/run.go
```diff
@@ -60,8 +60,9 @@ func addRunFlags(cmd *cobra.Command) {
 
 	cmd.Flags().Bool("sync-usage-file", false, "Sync usage-file with missing resources, needs usage-file too (experimental)")
 
-	cmd.Flags().Bool("hcl-only", false, "Run infracost without terraform, parsing hcl files at the given path. This option does not need credentials and is much quicker to run.")
-	cmd.Flags().StringSlice("tfvar-files", nil, "Path to .tfvars file, can be used multiple times and evaluated in order of specification")
+	cmd.Flags().Bool("hcl-only", false, "Run infracost without terraform, parsing hcl files at the given path. This option does not need credentials and is much quicker to run. (experimental)")
+	cmd.Flags().StringSlice("var-file", nil, "Can only be used with --hcl-only. Load variable values from the given file, in addition\nto the default files terraform.tfvars and *.auto.tfvars. Use this option more than once to\ninclude more than one variables file. (experimental)")
+	cmd.Flags().StringSlice("var", nil, "Can only be used with --hcl-only. Set a value for one of the input variables in the root module\nof the terraform configuration. Use this option more than once to set more than one variable. (experimental)")
```
aliscott, line 65: I'm not reviewing these just now, assuming that finalizing these is part of a different story.

aliscott (approved):
LGTM

internal/hcl/parser.go
```diff
@@ -32,31 +48,39 @@ func OptionStopOnHCLError() Option {
...
+				continue
+			}
 
-func TfVarsToOption(varsPaths ...string) (Option, error) {
-	for _, p := range varsPaths {
-		tfvp, _ := filepath.Abs(p)
-		_, err := os.Stat(tfvp)
-		if err != nil {
-			return nil, fmt.Errorf("passed tfvar file does not exist")
+			ctyVars[pieces[0]] = cty.StringVal(pieces[1])
```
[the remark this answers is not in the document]
hugorut: Nope this is what the terraform source does with input vars too

cmd/infracost/run.go
hugorut: 👍

aliscott:
> I didn't think about this tbh - do you think we should handle this case. It might be wise to cover all bases?

I think we should start without it. Currently the HCL parsing only consumes what is in code, whereas this looks at the ambient runtime environment. So I'm hesitant to add it unless we find a compelling reason.

PR #1382, opened 2022-02-17
Test/hcl multi project

tim775:


hugorut (approved):
LGTM

cmd/infracost/cmd_test.go
```diff
@@ -43,6 +43,12 @@ func DefaultOptions() *GoldenFileOptions {
 }
 
 func GoldenFileCommandTest(t *testing.T, testName string, args []string, testOptions *GoldenFileOptions, ctxOptions ...func(ctx *config.RunContext)) {
+	if testOptions == nil || !testOptions.OnlyRunHCL {
+		t.Run("Terraform CLI", func(t *testing.T) {
+			goldenFileCommandTest(t, testName, args, testOptions, ctxOptions)
+		})
+	}
+
```
hugorut, line 51: I'm not sure if this assumption is true or not but I believe that having the `cli` tests last means that `-update` flag will update using the CLI output - which should be the bearer of the truth. Theoretically you could get round this by filtering the tests to the sub test, but 🤷 .

tim775: I thought I needed to do this so the CLI test would load modules that the HCL test needed to be present, but it seems like that's not true, so: reverted.

PR #1383, opened 2022-02-17
Log out-file message if log-level is set

aliscott:
We shouldn't output this to the stderr if the user has specified a log level. Instead we should log it as an info level log.

### DIFF
--- a/cmd/infracost/main.go
+++ b/cmd/infracost/main.go
@@ -243,18 +243,22 @@ func loadGlobalFlags(ctx *config.RunContext, cmd *cobra.Command) error {
 }
 
 // saveOutFile saves the output of the command to the file path past in the `--out-file` flag
-func saveOutFile(cmd *cobra.Command, outFile string, b []byte) error {
-	return saveOutFileWithMsg(cmd, outFile, fmt.Sprintf("Output saved to %s\n", outFile), b)
+func saveOutFile(ctx *config.RunContext, cmd *cobra.Command, outFile string, b []byte) error {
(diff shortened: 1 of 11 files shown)


cmd/infracost/main.go
```diff
@@ -243,18 +243,22 @@ func loadGlobalFlags(ctx *config.RunContext, cmd *cobra.Command) error {
...
 	if err != nil {
 		return errors.Wrap(err, "Unable to save output")
 	}
 
-	cmd.PrintErr(successMsg)
+	if ctx.Config.IsLogging() {
+		log.Info(successMsg)
+	} else {
+		cmd.PrintErrf("%s\n", successMsg)
+	}
```
aliscott, line 261: We have this in a few places now, so at some point we need a better way of abstracting the logging vs print to terminal logic, but I didn't want to do that as part of this change.

cmd/infracost/run.go
```diff
@@ -225,17 +225,16 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
...
-		err = saveOutFile(cmd, outFile, b)
+		err = saveOutFile(runCtx, cmd, outFile, b)
 		if err != nil {
 			return err
 		}
 	} else {
+		// Print a new line to separate the logs from the output
+		if runCtx.Config.IsLogging() {
+			cmd.PrintErrln()
+		}
```
aliscott, line 237: We only need this blank line in between logs and the Infracost output being printed, so I moved it here.

hugorut (approved):
LGTM

Issue #1265, opened 2022-01-05
CLI glitches on Windows + Git Bash


3dbrows:
```
$ infracost -v
Infracost v0.9.16
```

Using `Git Bash 2.28.0.windows.1` on Windows 10 (a common way to use Git, and common Linux CLI utilities) produces the following output. The program quits immediately with no opportunity to supply a y/N answer.
```
$ infracost register
You already have an Infracost API key saved in C:\Users\Duncan\.config\infracost/credentials.yml
? Would you like to generate a new API key? [y/N] Ôûê

Error: ^D
```

This is avoided by using `winpty`, but on entering `y` some output is doubled.
```
$ winpty infracost register
You already have an Infracost API key saved in C:\Users\Duncan\.config\infracost/credentials.yml
Would you like to generate a new API key: y
Would you like to generate a new API key: y
```

Of note: `kubectl exec` detects and objects to being run without `winpty` in Git Bash:
```
$ kubectl exec -it ubuntu-846b56b68c-qtfxn -- sh
Unable to use a TTY - input is not a terminal or the right kind of file
```
```
$ winpty kubectl exec -it ubuntu-846b56b68c-qtfxn -- sh
#
```

vdmgolub:
This is fixed and will be released with v0.9.19.

tim775:
@3dbrows This is live in v0.9.19

PR #1388, opened 2022-02-22
fix: Use simple prompts for register commands on Windows

### LINKED ISSUE #924: infracost register outputs lines twice on Windows
I noticed the name/email output was doubled:
```
C:\infracost>infracost.exe register
Please enter your name and email address to get an API key.
See our FAQ (https://www.infracost.io/docs/faq) for more details.
Name: Txxxxr
Name: Txxxxr
Email: xxxx@gmail.com
Email: xxxx@gmail.com

Thank you Txxxxxr!
Your API key is: xxxx
```

vdmgolub:
The prompt libraries like promptui or survey produce double output on Windows.
Using simple prompts solves the issue, but doesn't alter experience for
Linux/macOS users.

Fixes #924 and #1265.

### DIFF
--- a/cmd/infracost/register.go
+++ b/cmd/infracost/register.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"runtime"
 	"strings"
 
 	"github.com/manifoldco/promptui"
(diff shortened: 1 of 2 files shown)


internal/ui/promts.go
```diff
@@ -0,0 +1,57 @@
...
+		if input != "" {
+			break
+		}
+	}
+
+	return input
+}
+
+func YesNoPrompt(label string) bool {
+	choices := "y/N"
```
hugorut, line 37:
```suggestion
	choices := "y/n"
```

internal/ui/promts.go
```diff
@@ -0,0 +1,57 @@
...
+	input := ""
```
hugorut, line 38:
```suggestion

```

internal/ui/promts.go
```diff
@@ -0,0 +1,57 @@
...
+	reader := bufio.NewReader(os.Stdin)
+
+	for {
+		fmt.Fprintf(os.Stdout, "%s [%s] ", label, choices)
+		input, _ = reader.ReadString('\n')
```
hugorut, line 43:
```suggestion
		input, _ := reader.ReadString('\n')
```

internal/ui/promts.go
```diff
@@ -0,0 +1,57 @@
+package ui
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+type ValidateFn func(input string) error
```
hugorut, line 10: docs on exported types please - also `StringPrompt` and `YesNoPrompt`

cmd/infracost/register.go
```diff
@@ -143,88 +144,73 @@ func registerCmd(ctx *config.RunContext) *cobra.Command {
...
-	}
-	name, err := p.Run()
-	name = strings.TrimSpace(name)
+	name, err := stringPrompt("Name", func(input string) error {
+		input = strings.TrimSpace(input)
+		if input == "" {
+			return errors.New("Please enter a name")
+		}
+		return nil
+	})
```
hugorut, line 153: Can we extract these funcs to the ui pkg so they're inline with the interface

hugorut (commented):
few tweaks

cmd/infracost/register.go
hugorut: by that i mean the validate func

hugorut: so:

```
func(input string) error {
		input = strings.TrimSpace(input)
		if input == "" {
			return errors.New("Please enter a name")
		}
		return nil
	}
```

internal/ui/promts.go
vdmgolub: `N` means it's the default behavior if left empty. Same as promptui shows it.

cmd/infracost/register.go
vdmgolub: I'm not sure I follow. `ui` provides prompts to be used anywhere, but these validate functions are unique for `register` command. I would see such function having "Please enter a <any field name>" that takes an argument with the error message, and it seems redundant. It may also make sense to introduce `required` flag for the `StringPrompt`.

internal/ui/promts.go
vdmgolub: Updated.

internal/ui/promts.go
vdmgolub: Updated.

cmd/infracost/register.go
hugorut: my preference would be to have this as exported funcs in the ui pkg like so:

```
var (
   ValidateName ValidateFn = func(input string) error {
		input = strings.TrimSpace(input)
		if input == "" {
			return errors.New("Please enter a name")
		}
		return nil
    }
    ValidateEmail ValidateFn = func(input string) error {
		if input == "" {
			return errors.New("Please enter an email")
		}
		match, err := regexp.MatchString(".+@.+", input)
		if err != nil {
			return errors.New("Unable to validate email")
		}
		if !match {
			return errors.New("Please enter a valid email")
		}
		return nil
	}
)

type ValidateFn func(input string) error

```

this seems quite generic to me

hugorut: but if you feel more comfortable with it stuffed in register the thats ok with me

vdmgolub: I see, but we still need to provide a `label` for the prompts. So having for example `StringPrompt("Foobar", ui.ValidateName)` will show `Please enter a name` instead of `Please enter a foobar`. With the email validation it may make more sense introducing `EmailPrompt` function so there is no disconnect between the label and error messages.

hugorut: yeah but reading `StringPrompt("Foobar", ui.ValidateName)` you'd realise validateName isn't the valid validate function. So you'd just create another function. If you really want to make it generic - which I don't feel theres too much benefit in this case then you can just have a function that decorates the validate function. e.g:


```go

func ValidateStringNotEmpty(emptyMsg string) ValidateFn {
    return func(input string) error {
		input = strings.TrimSpace(input)
		if input == "" {
			return errors.New(emptyMsg)
		}
		return nil
    }
}
```

this would be used as:

```go
ui.StringPrompt("Foobar", ui.ValidateStringNotEmpty("foobar cannot be empty")
```

hugorut: let's just stick with the original functionality tho

hugorut (approved):
✅

cmd/infracost/register.go
vdmgolub: Yeah, I guess we can iterate on that when we need these simple prompts in other places 👍

PR #1390, opened 2022-02-22
HCL flags, logs and envs

aliscott:
Docs: https://github.com/infracost/docs/pull/170


internal/hcl/parser.go
```diff
@@ -12,13 +12,16 @@ import (
...
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 	"github.com/zclconf/go-cty/cty"
 
 	"github.com/infracost/infracost/internal/config"
 	"github.com/infracost/infracost/internal/hcl/modules"
 )
 
+// This sets a global logger for this package, which is a bit of a hack. In the future we should use a context for this.
+var log = logrus.StandardLogger().WithField("parser", "terraform_hcl")
```
aliscott, line 23: This is pretty hacky, so I'm not sure about it. The other option would just be to put a `HCL parsing: ` prefix on each log line.

cmd/infracost/breakdown.go
```diff
@@ -50,6 +50,14 @@ func breakdownCmd(ctx *config.RunContext) *cobra.Command {
 	cmd.Flags().String("format", "table", "Output format: json, table, html")
 	cmd.Flags().StringSlice("fields", []string{"monthlyQuantity", "unit", "monthlyCost"}, "Comma separated list of output fields: all,price,monthlyQuantity,unit,hourlyCost,monthlyCost.\nSupported by table and html output formats")
 
+	// Hidden experimental flags
+	cmd.Flags().Bool("terraform-parse-hcl", false, "Parse the HCL directly instead of generating a Terraform plan. This option does not need credentials and is faster (experimental)")
+	cmd.Flags().StringSlice("terraform-var-file", nil, "Load variable files from the given file, similar to Terraform's -var-file flag. Only supported with --terraform-parse-hcl (experimental)")
```
hugorut, line 55:
```suggestion
	cmd.Flags().StringSlice("terraform-var-file", nil, "Load variables from the given files, similar to Terraform's -var-file flag. Only supported with --terraform-parse-hcl (experimental)")
```

internal/hcl/parser.go
hugorut: think this works for now, perhaps we can actually set it on the `Parser` as a property. However, really not necessary for no

hugorut (approved):
:shipit:

cmd/infracost/breakdown.go
```diff
@@ -50,6 +50,14 @@ func breakdownCmd(ctx *config.RunContext) *cobra.Command {
...
+	cmd.Flags().StringSlice("terraform-var", nil, "Set a value for one of the input variables, similar to Terraform's -var flag. Only supported with --terraform-parse-hcl (experimental)")
+	_ = cmd.Flags().MarkHidden("terraform-parse-hcl")
+	_ = cmd.Flags().MarkHidden("terraform-var-file")
+	_ = cmd.Flags().MarkHidden("terraform-var")
```
hugorut, line 59: when will users see these if they're hidden flags?

aliscott: They're not shown in the help text at all for the first cut.

internal/hcl/parser.go
aliscott: I thought about that but it doesn't cover all the logs for this package, so would need to be passed around a fair amount, which I think we'd want to use a context for in the future.

cmd/infracost/breakdown.go
hugorut: Is there anyway the user can see them at all then. How can they see the flag description if they're not part of `--help`. Is there a `--help --show-hidden`? 

So for the first draft we're essentially relying totally on the infracost docs to get people to find out about this right?

aliscott: > Is there anyway the user can see them at all then. 

Not currently.

For the first draft we're not making it very discoverable. We'll be reaching out to users directly to get them to try it and pointing them to the docs.

hugorut: 👍

PR #1391, opened 2022-02-22
fix: Validate PR/MR number in comment commands

vdmgolub:
Current behavior when the flag is followed by other flags:
```bash
Error: invalid argument "--bitbucket-token" for "--pull-request" flag: strconv.ParseInt: parsing "--bitbucket-token": invalid syntax
exit status 1
make: *** [run] Error 1
```

The updated error:
```bash
Error: invalid argument "--bitbucket-token" for "--pull-request" flag: must be integer
exit status 1
make: *** [run] Error 1
```

When the flag goes last (current behavior with or without the fix):
```bash
Error: flag needs an argument: --pull-request
exit status 1
make: *** [run] Error 1
```

### DIFF
--- a/cmd/infracost/comment.go
+++ b/cmd/infracost/comment.go
@@ -1,9 +1,12 @@
 package main
 
 import (
+	"strconv"
+
 	"github.com/infracost/infracost/internal/config"
 	"github.com/infracost/infracost/internal/output"
 	"github.com/infracost/infracost/internal/ui"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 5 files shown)


tim775:
I don't think this hurts anything but I'm not sure `Error: required flag(s) "bitbucket-token" not set` is better than the current error since that kind of points you in the wrong direction since the problem is with the pull-request flag.

I guess ideally we would want`Error: flag needs an argument: --pull-request` in all cases, but if that's not possible or easy I'm ok with whatever seems best to you.

vdmgolub:
@tim775 Thank you! My reasoning comes from the nature of the `Error: invalid argument "--bitbucket-token" for "--pull-request" flag: strconv.ParseInt: parsing "--bitbucket-token": invalid syntax` error: it looks like an unhandled error by exposing Go's `ParseInt`.

> I guess ideally we would want `Error: flag needs an argument: --pull-request` in all cases, but if that's not possible or easy I'm ok with whatever seems best to you.

Unfortunately, I don't think it's possible as the empty env var will be ignored and the next word (in this case the token flag) will become a value.

 Maybe I'm inventing a problem here and we can close this PR without merging 😕

tim775 (approved):

PR #1397, opened 2022-02-24
Adding in a switch for Bitbucket Server.

### LINKED ISSUE #1396: Commenting for Bitbucket Server is broken
The new support for commenting in Bitbucket Server is broken. Right now the URL that is built for the API only works with the Bitbucket 2.0 API that is used in Bitbucket cloud.

In 2.0 the path to pull requests uses "pullrequests" in the URL but for the 1.0 API it uses "pull-requests".

Link to 1.0 API:
https://docs.atlassian.com/bitbucket-server/rest/7.20.0/bitbucket-rest.html#idp332

Link to 2.0 API:
https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/#api-repositories-workspace-repo-slug-pullrequests-pull-request-id-comments-post

There are two lines where this needs to be fixed.
https://github.com/infracost/infracost/blob/master/internal/comment/bitbucket.go#L171
https://github.com/infracost/infracost/blob/master/internal/comment/bitbucket.go#L239

lazzurs:
The 1.0 API that Bitbucket Server uses has a different path for pull
request comments than the 2.0 API so adding in a switch that will
generate the right path for each API version.

Closes #1396.

### DIFF
--- a/internal/comment/bitbucket.go
+++ b/internal/comment/bitbucket.go
@@ -168,7 +168,11 @@ func NewBitbucketPRHandler(ctx context.Context, repo string, targetRef string, e
 // CallFindMatchingComments calls the Bitbucket API to find the pull request
 // comments that match the given tag, which has been embedded in the comment.
 func (h *bitbucketPRHandler) CallFindMatchingComments(ctx context.Context, tag string) ([]Comment, error) {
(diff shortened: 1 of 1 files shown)


lazzurs:
This was a misunderstanding of how the switch happens in the code. Closing PR.

PR #1400, opened 2022-02-24
chore: report if tf_var is set if running hcl

hugorut:
Adds reporting data if tf_var env is found when using hcl. This can help us understand potential discrepancies between hcl and tf runs.


aliscott (approved):
LGTM

PR #1401, opened 2022-02-24
Feat/add app autoscaling target

tim775:
This adds support for reverse references.  It might be useful to look at the commits separately when reviewing.


aliscott (approved):
LGTM! Does this need updates to the contributing docs - maybe just pointing to this resource as an example.

tim775:
> LGTM! Does this need updates to the contributing docs - maybe just pointing to this resource as an example.

Yes, good point.  I've updated the docs.

PR #1407, opened 2022-02-25
fix: cleanup hcl debug for verbose messages

hugorut:
* Removes attribute evaluation messages caused by partial context at time of evaluation
* Adds verbose env var `INFRACOST_HCL_DEBUG_VERBOSE` that can show these messages, as this could help us track a bug down in the future
* Change context evaluation log message to be more clear
* Changes fake "arn" attribute on hcl resources to be the format AWS defines so we skip the : `Unexpected ARN format` debug line


aliscott (approved):
LGTM

PR #1408, opened 2022-02-25
feat: Add application autoscaling support for AWS MSK clusters

tim775:


internal/providers/terraform/aws/testdata/msk_cluster_test/msk_cluster_test.usage.yml
```diff
@@ -0,0 +1,4 @@
+version: 0.1
+resource_usage:
+  aws_msk_cluster.cluster-autoscaling-usage:
+    capacity: 1789
```
vdmgolub, line 4: Noticed that an example of this parameter usage is missing in `infracost-usage-example.yml`, but I guess it can be added in another PR.

vdmgolub (approved):
👍

internal/providers/terraform/aws/testdata/msk_cluster_test/msk_cluster_test.usage.yml
tim775: No that is actually a typo I will fix that and make it more obvious when the test is failing

PR #1409, opened 2022-02-26
feat: Add application autoscaling support for ElastiCache replication groups

tim775:
The first commit adds support for renamed terraform attributes related to the number of shards/replicas.  The second commit adds app autoscaling support.

### DIFF
--- a/internal/providers/terraform/aws/elasticache_replication_group.go
+++ b/internal/providers/terraform/aws/elasticache_replication_group.go
@@ -11,7 +11,7 @@ func getElastiCacheReplicationGroupItem() *schema.RegistryItem {
 		RFunc:               NewElastiCacheReplicationGroup,
 		ReferenceAttributes: []string{"aws_appautoscaling_target.resource_id"},
 		CustomRefIDFunc: func(d *schema.ResourceData) []string {
-			// returns a table name that will match the custom format used by aws_appautoscaling_target.resource_id
+			// returns a name that will match the custom format used by aws_appautoscaling_target.resource_id
 			name := d.Get("replication_group_id").String()
 			if name != "" {
 				return []string{"replication-group/" + name}
--- a/internal/providers/terraform/aws/elasticache_replication_group.go
+++ b/internal/providers/terraform/aws/elasticache_replication_group.go
@@ -3,7 +3,6 @@ package aws
 import (
 	"github.com/infracost/infracost/internal/resources/aws"
 	"github.com/infracost/infracost/internal/schema"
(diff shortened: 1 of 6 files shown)


internal/providers/terraform/aws/elasticache_replication_group.go
```diff
@@ -3,27 +3,57 @@ package aws
...
+			return nil
+		},
 	}
 }
 func NewElastiCacheReplicationGroup(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	cacheClusters := d.Get("num_cache_clusters").Int()
+	if d.IsEmpty("num_cache_clusters") {
+		// check for deprecated attribute
+		cacheClusters = d.Get("number_cache_clusters").Int()
+	}
```
hugorut, line 28:
```suggestion
	cacheClusters := d.GetInt64OrDefault("num_cache_clusters", 1)
	if d.IsEmpty("num_cache_clusters") {
		// check for deprecated attribute
		cacheClusters = d.Get("number_cache_clusters").Int()
	}
```
`num_cache_clusters` defaults to `1` if not specified and this won't be picked up by hcl

internal/providers/terraform/aws/elasticache_replication_group.go
```diff
@@ -3,27 +3,57 @@ package aws
...
+
+	clusterNodeGroups := d.Get("num_node_groups").Int()
+	if d.IsEmpty("num_node_groups") {
+		// check for deprecated attribute
+		clusterNodeGroups = d.Get("cluster_mode.0.num_node_groups").Int()
+	}
```
hugorut, line 34:
```suggestion
	clusterNodeGroups := d. GetInt64OrDefault("num_node_groups", 1)
	if d.IsEmpty("num_node_groups") {
		// check for deprecated attribute
		clusterNodeGroups = d.Get("cluster_mode.0.num_node_groups").Int()
	}
```

same as above

hugorut (commented):
Few things for HCL specifics, other than that looks good. FYI hcl is now in master so I'd recommend rebasing and rerunning the tests to make sure this works with HCL.

internal/providers/terraform/aws/elasticache_replication_group.go
tim775: Hey just to double check, why do you think num_node_groups has a default of 1?  I'm not seeing that in the docs.

hugorut: yeah you're right - I must have seen misread the propery

PR #1411, opened 2022-02-28
feat(aws): add elastic beanstalk environment

### LINKED ISSUE #1410: Resource to support: `aws_elastic_beanstalk_environment`
Please can you add support for `aws_elastic_beanstalk_environment`

lukeadams84:
## Objective:

Add support for `aws_elastic_beanstalk_environment` and `aws_elastic_beanstalk_application` as FREE. Fixes #1410.

## Pricing details:

Price breakdown for elastic beanstalk environments is ec2 instances associated with the environment multiplied by the minimum ASG numbers.

Additional costs are associated for the load balancer type. If StreamLogs is enabled, there can be usage costs for cloudwatch metrics, including ingest, scanning and storage in GB.

## Status:

- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/pull/175)

## Issues:

None

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -161,6 +161,11 @@ resource_usage:
     monthly_cpu_credit_hrs: 350 # Number of hours in the month where the instance is expected to burst. Only applicable with t2, t3 & t4 Instance types. T2 requires credit_specification to be unlimited.
     vcpu_count: 2 # Number of the vCPUs for the instance type. Only applicable with t2, t3 & t4 Instance types. T2 requires credit_specification to be unlimited.
 
+  aws_elastic_beanstalk_environment.my_eb_environment:
+    monthly_data_ingested_gb: 50  # Cloudwatch log stream ingest in GB
+    monthly_data_scanned_gb: 50   # Cloudwatch data scanned in GB
+    storage_gb: 100               # Cloudwatch storage used in GB
+
   aws_elasticache_cluster.my_redis_snapshot:
     snapshot_storage_size_gb: 10000 # Size of Redis snapshots in GB.
 
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -162,14 +162,24 @@ resource_usage:
     vcpu_count: 2 # Number of the vCPUs for the instance type. Only applicable with t2, t3 & t4 Instance types. T2 requires credit_specification to be unlimited.
 
   aws_elastic_beanstalk_environment.my_eb_environment:
-    monthly_data_ingested_gb: 50  # Cloudwatch log stream ingest in GB
-    monthly_data_scanned_gb: 50   # Cloudwatch data scanned in GB
-    storage_gb: 100               # Cloudwatch storage used in GB
-    new_connections: 500000    # Loadbalancer newly established connections per second on average.
-    active_connections: 100000 # Loadbalancer active connections per minute on average.
-    processed_bytes_gb: 25000  # The number of bytes processed by the load balancer for HTTP(S) requests and responses in GB.
-    rule_evaluations: 10000 # Loadbalancer rule evaulations
-
+    db:
+      additional_backup_storage_gb: 1000  # Amount of backup storage used that is in excess of 100% of the storage size for all databases in GB.
+      monthly_standard_io_requests: 10000 # Monthly number of input/output requests for database.
+      monthly_additional_performance_insights_requests: 10000 # Monthly Performance Insights API requests above the 1000000 requests included in the free tier.
+    ec2:
+      monthly_cpu_credit_hrs: 350 # Number of hours in the month where the instance is expected to burst. Only applicable with t2, t3 & t4 Instance types. T2 requires credit_specification to be unlimited.
+    cloudwatch:
+      storage_gb: 1000               # Total data stored by CloudWatch logs in GB.
+      monthly_data_ingested_gb: 1000 # Monthly data ingested by CloudWatch logs in GB.
+      monthly_data_scanned_gb: 200   # Monthly data scanned by CloudWatch logs insights in GB.
+    lb:
+      new_connections: 500000    # Number of newly established connections per second on average.
+      active_connections: 100000 # Number of active connections per minute on average.
+      processed_bytes_gb: 25000  # The number of bytes processed by the load balancer for HTTP(S) requests and responses in GB.
+      rule_evaluations: 10000    # Number of rule evaluations on application/network load balancers
+    elb:
+      monthly_data_processed_gb: 10000 # Monthly data processed on classic loadbalaner
+    
   aws_elasticache_cluster.my_redis_snapshot:
     snapshot_storage_size_gb: 10000 # Size of Redis snapshots in GB.
 
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -165,6 +165,10 @@ resource_usage:
     monthly_data_ingested_gb: 50  # Cloudwatch log stream ingest in GB
     monthly_data_scanned_gb: 50   # Cloudwatch data scanned in GB
     storage_gb: 100               # Cloudwatch storage used in GB
+    new_connections: 500000    # Loadbalancer newly established connections per second on average.
+    active_connections: 100000 # Loadbalancer active connections per minute on average.
+    processed_bytes_gb: 25000  # The number of bytes processed by the load balancer for HTTP(S) requests and responses in GB.
+    rule_evaluations: 10000 # Loadbalancer rule evaulations
 
   aws_elasticache_cluster.my_redis_snapshot:
     snapshot_storage_size_gb: 10000 # Size of Redis snapshots in GB.
--- a/internal/providers/terraform/aws/elastic_beanstalk_environment.go
+++ b/internal/providers/terraform/aws/elastic_beanstalk_environment.go
@@ -13,10 +13,9 @@ func getElasticBeanstalkEnvironmentRegistryItem() *schema.RegistryItem {
 }
 
 func newElasticBeanstalkEnvironment(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
-
 	var region = d.Get("region").String()
-	var StreamLogs = false
-	var DBIncluded = false
+	var streamLogs = false
+	var dBIncluded = false
 
 	r := &aws.ElasticBeanstalkEnvironment{
 		Address: d.Address,
@@ -29,22 +28,27 @@ func newElasticBeanstalkEnvironment(d *schema.ResourceData, u *schema.UsageData)
 		Region:  region,
 		Tenancy: "dedicated",
 	}
+
 	volume := &aws.EBSVolume{
 		Address: "aws_ebs_volume",
 		Region:  region,
 	}
+
 	db := &aws.DBInstance{
 		Address: "aws_db_instance",
 		Region:  region,
 	}
+
 	cwlg := &aws.CloudwatchLogGroup{
 		Address: "aws_cloudwatch_log_group",
 		Region:  region,
 	}
+
 	elb := &aws.ELB{
 		Address: "aws_elb",
 		Region:  region,
 	}
+
 	lb := &aws.LB{
 		Address: "aws_loadbalancer",
 		Region:  region,
@@ -65,9 +69,9 @@ func newElasticBeanstalkEnvironment(d *schema.ResourceData, u *schema.UsageData)
 		case "LoadBalancerType":
 			r.LoadBalancerType = setting.Get("value").String()
 		case "StreamLogs":
-			StreamLogs = true
+			streamLogs = true
 		case "DBInstanceClass":
-			DBIncluded = true
+			dBIncluded = true
 			db.InstanceClass = setting.Get("value").String()
 		case "MultiAZDatabase":
 			db.MultiAZ = true
@@ -79,12 +83,15 @@ func newElasticBeanstalkEnvironment(d *schema.ResourceData, u *schema.UsageData)
 	}
 	r.LaunchConfiguration = lc
 	r.LaunchConfiguration.RootBlockDevice = volume
-	if DBIncluded {
+
+	if dBIncluded {
 		r.DBInstance = db
 	}
-	if StreamLogs {
+
+	if streamLogs {
 		r.CloudwatchLogGroup = cwlg
 	}
+
 	if r.LoadBalancerType == "classic" {
 		r.ElasticLoadBalancer = elb
 	} else {
--- a/internal/providers/terraform/aws/elastic_beanstalk_environment.go
+++ b/internal/providers/terraform/aws/elastic_beanstalk_environment.go
@@ -14,136 +14,76 @@ func getElasticBeanstalkEnvironmentRegistryItem() *schema.RegistryItem {
 
 func newElasticBeanstalkEnvironment(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
 
-	var loadBalancerType = ""
-	var instanceType = ""
-	var instanceCount int64
-	var streamLogs bool
-	var rdsIncluded bool
-	var rootVolumeSize int64
-	var rootVolumeType = "gp2"
-	var rootVolumeIOPS int64
-
-	autoscalingGroup := &schema.ResourceData{}
-	data := autoscalingGroup
-
-	for _, setting := range d.Get("setting").Array() {
-		if setting.Get("name").String() == "InstanceTypes" {
-			instanceType = setting.Get("value").String()
-		}
-		if setting.Get("name").String() == "MinSize" {
-			instanceCount = setting.Get("value").Int()
-		}
-		if setting.Get("name").String() == "RootVolumeSize" {
-			rootVolumeSize = setting.Get("value").Int()
-		}
-		if setting.Get("name").String() == "RootVolumeIOPS" {
-			rootVolumeIOPS = setting.Get("value").Int()
-		}
-		if setting.Get("name").String() == "RootVolumeType" {
-			rootVolumeType = setting.Get("value").String()
-		}
-		if setting.Get("name").String() == "LoadBalancerType" {
-			data.Set("load_balancer_type", setting.Get("value").String())
-			loadBalancerType = setting.Get("value").String()
-		}
-		if setting.Get("name").String() == "StreamLogs" {
-			streamLogs = true
-		}
-		if setting.Get("name").String() == "DBInstanceClass" {
-			rdsIncluded = true
-			data.Set("db_instance_class", setting.Get("value").String())
-		}
-		if setting.Get("name").String() == "MultiAZDatabase" {
-			data.Set("multi_az", true)
-		}
-		if setting.Get("name").String() == "DBEngine" {
-			data.Set("engine", setting.Get("value").String())
-		}
-		if setting.Get("name").String() == "DBAllocatedStorage" {
-			data.Set("allocated_storage", setting.Get("value").Float())
-		}
-	}
-	data.Set("region", d.Get("region").String())
-	data.Set("name", d.Get("name").String())
+	var region = d.Get("region").String()
 
 	r := &aws.ElasticBeanstalkEnvironment{
-		Address:          d.Address,
-		Region:           d.Get("region").String(),
-		Name:             d.Get("name").String(),
-		InstanceType:     instanceType,
-		InstanceCount:    instanceCount,
-		LoadBalancerType: loadBalancerType,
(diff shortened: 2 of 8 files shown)


hugorut:
@lukeadams84 thanks for doing this ❤️ . We've got some big planning meetings today so I might not get to review/look over this today, but i'll endeavour to get back to you as quick as I can.

hugorut:
@lukeadams84 thanks again for this. I've had a peek through the code and done some reading. I'm wasn't familiar with elastic beanstalk environments so I appreciate you bringing it to our attention. My initial thoughts though is that we might need to go a different route with this resource than the norm. 

My understanding that elastic beanstalk environment essentially simplifies provisioning a set of underlying AWS resources to run an application. In terraform you can change the base set of configuration options for this using the `setting` block. Because a lot of these `settings` are resources that Infracost already supports (ec2 instances, autoscaling group, cloudwatch, launch configuration, e.t.c.) my preference would be to use this resource fetch the configuration and then delegate pricing to those Infracost resources. 

This would mean that all `aws.ElasticBeanstalkEnvironment` would do is call other resources like `aws.Instance` to output their cost components to the CLI. This would be more maintainable in the longer term as it would mitigate code drift between resources that are, under the hood, the same.

Let me consult with the team and get back to you.

lukeadams84:
> @lukeadams84 thanks again for this. I've had a peek through the code and done some reading. I'm wasn't familiar with elastic beanstalk environments so I appreciate you bringing it to our attention. My initial thoughts though is that we might need to go a different route with this resource than the norm.
> 
> My understanding that elastic beanstalk environment essentially simplifies provisioning a set of underlying AWS resources to run an application. In terraform you can change the base set of configuration options for this using the `setting` block. Because a lot of these `settings` are resources that Infracost already supports (ec2 instances, autoscaling group, cloudwatch, launch configuration, e.t.c.) my preference would be to use this resource fetch the configuration and then delegate pricing to those Infracost resources.
> 
> This would mean that all `aws.ElasticBeanstalkEnvironment` would do is call other resources like `aws.Instance` to output their cost components to the CLI. This would be more maintainable in the longer term as it would mitigate code drift between resources that are, under the hood, the same.
> 
> Let me consult with the team and get back to you.

Yes, I understand. I think one of the headaches seems to be that the EC2 instances (for example) that elastic beanstalk uses are more constrained than if you were creating an EC2 instance normally. I believe the instances are always dedicated, and only support Linux as an operating system.

I'll have another look at your examples and see what I can work out to use the existing resources.

hugorut:
I'd recommend having a peek at [eks node group](https://github.com/infracost/infracost/blob/master/internal/resources/aws/eks_node_group.go) or [launch template](https://github.com/infracost/infracost/blob/master/internal/resources/aws/launch_template.go) as they do similiar-ish things

lukeadams84:
> I'd recommend having a peek at [eks node group](https://github.com/infracost/infracost/blob/master/internal/resources/aws/eks_node_group.go), [launch template](https://github.com/infracost/infracost/blob/master/internal/resources/aws/launch_template.go) or [neptune cluster snapshot](https://github.com/infracost/infracost/blob/3272eb3627debde54a548bd9eabd80781f297c22/internal/resources/aws/neptune_cluster_snapshot.go) as they do similiar-ish things

Okay, have updated now. It looks like it covers the bases.

hugorut:
@lukeadams84 amazing - I'll take a peek later

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,173 @@
...
+			count, err := aws.AutoscalingGetInstanceCount(ctx, r.Region, r.Name)
+			if err != nil {
+				return err
+			}
+			if count > 0 {
+				u["instances"] = int64(math.Round(count))
+			}
+		}
+		return nil
+	}
```
hugorut, line 164: I don't believe we need this as we're not using any attributes for the `AutoscalingGroup` usage params. We could potentially add these in the future but I'm leaning towards not allowing them in at this point.

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,173 @@
...
+	usageSchema := make([]*schema.UsageItem, 0, len(AutoscalingGroupUsageSchema))
+	for _, u := range AutoscalingGroupUsageSchema {
+		if u.Key == "instances" {
+			usageSchema = append(usageSchema, &schema.UsageItem{Key: "instances", DefaultValue: intVal(instanceCount), ValueType: schema.Int64})
+		} else {
+			usageSchema = append(usageSchema, u)
+		}
+	}
+	return usageSchema
+}
```
hugorut, line 96: I don't believe we need this as we're not using the `AutoscalingGroup` usage at the moment. I think this is something we could add down the line, but let's remove atm.

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+			if count > 0 {
+				u["instances"] = int64(math.Round(count))
+			}
+		}
+		return nil
+	}
+	return &schema.Resource{
+		Name:           r.Address,
+		UsageSchema:    r.getUsageSchemaWithDefaultInstanceCount(),
```
hugorut, line 167:
```suggestion
		UsageSchema:  ElasticBeanstalkEnvironmentUsageSchema,
```

hugorut (changes requested):
Starting to shape up nicely. There's a few points to resolve before we get this in though.

lukeadams84:
Can you tell I am new to go? :D Ill get on these, thanks for looking at it.

hugorut:
@lukeadams84 not at all! It's just a bit of a bizarre resource, you chose a tricky one to start! Ping here if you have any other questions. Also feel free to message me [on slack](https://www.infracost.io/community-chat) if you have any ad hoc questions that you want to run by me with a quicker response rate

infracost-usage-example.yml
```diff
@@ -161,6 +161,15 @@ resource_usage:
...
     vcpu_count: 2 # Number of the vCPUs for the instance type. Only applicable with t2, t3 & t4 Instance types. T2 requires credit_specification to be unlimited.
 
+  aws_elastic_beanstalk_environment.my_eb_environment:
+    monthly_data_ingested_gb: 50  # Cloudwatch log stream ingest in GB
+    monthly_data_scanned_gb: 50   # Cloudwatch data scanned in GB
+    storage_gb: 100               # Cloudwatch storage used in GB
+    new_connections: 500000    # Loadbalancer newly established connections per second on average.
+    active_connections: 100000 # Loadbalancer active connections per minute on average.
+    processed_bytes_gb: 25000  # The number of bytes processed by the load balancer for HTTP(S) requests and responses in GB.
+    rule_evaluations: 10000 # Loadbalancer rule evaulations
```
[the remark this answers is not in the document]
hugorut: I think you're doing this because these params directly map to the sub resources. However, it's confusing to the user. I'd recommend setting them as sub yaml object properties:

```yaml
  aws_elastic_beanstalk_environment.my_eb_environment:
    cloudwatch:
      monthly_data_ingested_gb: 50  # Cloudwatch log stream ingest in GB
      monthly_data_scanned_gb: 50   # Cloudwatch data scanned in GB
      storage_gb: 100               # Cloudwatch storage used in GB
    lb:
      new_connections: 500000    # Loadbalancer newly established connections per second on average.
      active_connections: 100000 # Loadbalancer active connections per minute on average.
      processed_bytes_gb: 25000  # The number of bytes processed by the load balancer for HTTP(S) requests and responses in GB.
      rule_evaluations: 10000 # Loadbalancer rule evaulations
```

Then in the populate usage methods you'll have to pass the sub usage data to the sub resource, see my comment on `ElasticBeanstalkEnvironment.PopulateUsage` for more info.

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,103 @@
...
+		resources.PopulateArgsWithUsage(r.ElasticLoadBalancer, u)
+	} else {
+		resources.PopulateArgsWithUsage(r.LoadBalancer, u)
+	}
+	if r.StreamLogs {
+		resources.PopulateArgsWithUsage(r.CloudwatchLogGroup, u)
+	}
+	if r.RDSIncluded {
+		resources.PopulateArgsWithUsage(r.DBInstance, u)
+	}
```
hugorut, line 69: So the reason this isn't working with the sub resources is that the sub resources have different usage params from the ones defined here. e.g.  `CloudWatchLogGroup` has a usage param called `monthly_data_ingested_gb` not `cloudwatch_monthly_data_ingested_gb` therefore the sub resource can't parse the usage.

In order to sort i'd introduce a sub object in the usage file. See my comment on `infracost-usage-example` for more info. 

In order to do this need to change the usage schema to something like this:

```yaml
  aws_elastic_beanstalk_environment.my_eb_environment:
    cloudwatch:
      monthly_data_ingested_gb: 50  # Cloudwatch log stream ingest in GB
      monthly_data_scanned_gb: 50   # Cloudwatch data scanned in GB
      storage_gb: 100               # Cloudwatch storage used in GB
```

Then the `ElasticBeanstalkEnvironmentUsageSchema` would change to something like so:

```go
var ElasticBeanstalkEnvironmentUsageSchema = []*schema.UsageItem{
	{
	    Key: "cloudwatch", 
            DefaultValue: CloudwatchLogGroupUsageSchema,
            ValueType: schema.SubResourceUsage
         }
}
```

Then in populate usage you would have to do something like:


```go
func (r *ElasticBeanstalkEnvironment) PopulateUsage(u *schema.UsageData) {
	if r.CloudwatchLogGroup != nil {
		resources.PopulateArgsWithUsage(r.CloudwatchLogGroup, u["cloudwatch"])
	}
}
```

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,103 @@
...
+	{Key: "monthly_data_processed_gb", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "lb_new_connections", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "lb_active_connections", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "lb_rule_evaluations", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "lb_processed_bytes_gb", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "cloudwatch_monthly_data_ingested_gb", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "cloudwatch_monthly_data_scanned_gb", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "cloudwatch_storage_gb", DefaultValue: 0, ValueType: schema.Float64},
+	{Key: "instances", DefaultValue: 0, ValueType: schema.Int64},
+}, LaunchConfigurationUsageSchema...)
```
hugorut, line 54: I would hesitate to put `LaunchConfigurationUsageSchema` at the same level of the `ElasticBeanstalkEnvironment` -> let's move this to a sub usage param, e.g:

```yaml
 aws_elastic_beanstalk_environment.my_eb_environment:
    ec2:
       # launch config here
```

hugorut: see my comments below on `PopulateUsage` and in `infracost-usage-example` for more info

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,103 @@
...
+
+	StreamLogs            bool
+	MonthlyDataIngestedGB *float64 `infracost_usage:"cloudwatch_monthly_data_ingested_gb"`
+	StorageGB             *float64 `infracost_usage:"cloudwatch_storage_gb"`
+	MonthlyDataScannedGB  *float64 `infracost_usage:"cloudwatch_monthly_data_scanned_gb"`
+	MonthlyDataProcessed  *float64 `infracost_usage:"monthly_data_processed_gb"`
+	LBNewConnections      *float64 `infracost_usage:"lb_new_connections"`
+	LBActiveConnections   *float64 `infracost_usage:"lb_active_connections"`
+	LBRuleEvaluations     *float64 `infracost_usage:"lb_rule_evaluations"`
+	LBProcessedBytesGB    *float64 `infracost_usage:"lb_processed_bytes_gb"`
```
hugorut, line 33: There's no need to have these usage properties on this `ElasticBeanstalkEnvironment` struct as all the usage is for subresources, see comments below

internal/providers/terraform/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,89 @@
...
+		case "RootVolumeIOPS":
+			volume.IOPS = setting.Get("value").Int()
+		case "RootVolumeType":
+			volume.Type = setting.Get("value").String()
+		case "LoadBalancerType":
+			r.LoadBalancerType = setting.Get("value").String()
+		case "StreamLogs":
+			r.StreamLogs = true
+		case "DBInstanceClass":
+			r.RDSIncluded = true
```
hugorut, line 68: I'd move this to a `var` at this level. 

Then line `80` can go to this:

```go
if dbIncluded {
	r.DBInstance = db
}
```

This means that in `ElasticBeanstalkEnvironment` you can remove the properties and just check if `DBInstance` is not nil.

e.g:

```go
if r.DBInstance != nil {
    resources.PopulateArgsWithUsage(r.DBInstance, u)
}
```

hugorut (changes requested):

hugorut:
@lukeadams84 we're getting closer, just need to fix up these usage params.

internal/providers/terraform/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,97 @@
...
+		Name:  "aws_elastic_beanstalk_environment",
+		RFunc: newElasticBeanstalkEnvironment,
+	}
+}
+
+func newElasticBeanstalkEnvironment(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+
+	var region = d.Get("region").String()
+	var StreamLogs = false
+	var DBIncluded = false
```
hugorut, line 19:
```suggestion
	var streamLogs = false
	var dbIncluded = false
```

internal/providers/terraform/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,97 @@
...
+
+func getElasticBeanstalkEnvironmentRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "aws_elastic_beanstalk_environment",
+		RFunc: newElasticBeanstalkEnvironment,
+	}
+}
+
+func newElasticBeanstalkEnvironment(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+
```
hugorut, line 16:
```suggestion
```

Not a major thing, but could we remove these new lines at the beginning of functions. No reason to have them.

internal/providers/terraform/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,97 @@
...
+		r.DBInstance = db
+	}
+	if StreamLogs {
+		r.CloudwatchLogGroup = cwlg
+	}
+	if r.LoadBalancerType == "classic" {
+		r.ElasticLoadBalancer = elb
+	} else {
+		r.LoadBalancer = lb
+	}
```
hugorut, line 92: So, something we could do to improve here is to remove `ElasticLoadBalancer` as a property and have `LoadBalancer` of type `resources.CoreResource`

we'd need to modify the `CoreResource` type to include a `BuildResource` method: 

```go
type CoreResource interface {
	PopulateUsage(u *schema.UsageData)
	BuildResource() *schema.Resource // added
}
```

this would mean that `ElasticBeanstalkEnvironment` would look like such:

```go
type ElasticBeanstalkEnvironment struct {
        ...
        LoadBalancer resources.CoreResource
        ....
}
```

Which means you could remove `LoadBalancerType` as a property and just use it as a local var here. Setting `r.LoadBalancer` with either `elb` or `lb` depending on type.

Then `PopulateUsage` and `BuildResource` in `ElasticBeanstalkEnvironment` could just call to the single property.

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,115 @@
...
+
+	RootBlockDevice     *EBSVolume
+	CloudwatchLogGroup  *CloudwatchLogGroup
+	LoadBalancer        *LB
+	ElasticLoadBalancer *ELB
+	DBInstance          *DBInstance
+	LaunchConfiguration *LaunchConfiguration
+}
+
+// ElasticBeanstalkEnvironmentUsageSchema defines a list which represents the usage schema of ElasticBeanstalkEnvironment.
```
hugorut, line 34: Could we add some more meaningful docs here mentioning that all the usage schema comes from the sub resources.

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if r.CloudwatchLogGroup != nil {
+		resources.PopulateArgsWithUsage(r.CloudwatchLogGroup, schema.NewUsageData("cloudwatch", u.Get("cloudwatch").Map()))
+	}
+	resources.PopulateArgsWithUsage(r.LaunchConfiguration, schema.NewUsageData("ec2", u.Get("ec2").Map()))
+}
+
+// BuildResource builds a schema.Resource from a valid ElasticBeanstalkEnvironment struct.
+// This method is called after the resource is initialised by an IaC provider.
+// See providers folder for more information.
```
hugorut, line 89: could we add some information here that `ElasticBeanstalkEnvironment` doesn't have any real cost components with it. They are all associated with sub resources.

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if r.CloudwatchLogGroup != nil {
+		a.SubResources = append(a.SubResources, r.CloudwatchLogGroup.BuildResource())
+	}
+	switch r.LoadBalancerType {
+	case "classic":
+		a.SubResources = append(a.SubResources, r.ElasticLoadBalancer.BuildResource())
+	default:
+		a.SubResources = append(a.SubResources, r.LoadBalancer.BuildResource())
+	}
```
hugorut, line 111: If we take my suggestion about the `LoadBalancer` using an interface this can go to:

```suggestion
	if r.LoadBalander  != nil {
		a.SubResources = append(a.SubResources, r.LoadBalancer.BuildResource())
	}
```

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,115 @@
...
+	if r.LoadBalancer != nil {
+		resources.PopulateArgsWithUsage(r.LoadBalancer, schema.NewUsageData("lb", u.Get("lb").Map()))
+	}
+	if r.DBInstance != nil {
+		resources.PopulateArgsWithUsage(r.DBInstance, schema.NewUsageData("db", u.Get("db").Map()))
+	}
+	if r.CloudwatchLogGroup != nil {
+		resources.PopulateArgsWithUsage(r.CloudwatchLogGroup, schema.NewUsageData("cloudwatch", u.Get("cloudwatch").Map()))
+	}
```
hugorut, line 83: Can we put some spacing in between these `if`s

internal/resources/aws/elastic_beanstalk_environment.go
```diff
@@ -0,0 +1,115 @@
...
+type ElasticBeanstalkEnvironment struct {
+	Address string
+	Region  string
+	Name    string
+
+	InstanceCount    *int64
+	InstanceType     string
+	RDSIncluded      bool
+	LoadBalancerType string
+	StreamLogs       bool
```
hugorut, line 24: none of these, outside of `LoadBalancerType` seem to be used now. (But hopefully `LoadBalancerType` can also be removed). Let's get rid of them. Thanks

hugorut (changes requested):
Minor code tweaks, but logic looks solid

internal/providers/terraform/aws/elastic_beanstalk_environment.go
hugorut: Capitalised variables in GO denote exported types https://www.ardanlabs.com/blog/2014/03/exportedunexported-identifiers-in-go.html

hugorut (approved):
LGTM 🚀

hugorut:
@lukeadams84

PR #1412, opened 2022-02-28
fix: add run time values into singular attribute

hugorut:
sums project specific reporting metrics into singular value


aliscott (commented):
LGTM

Issue #1433, opened 2022-03-03
HCL Parsing doesn't report `azurerm_app_service_plan`


bgrainger:
I just tried the new `infracost breakdown --path=. --terraform-parse-hcl`. It is so much faster! 🎉

However, it didn't report my `azurerm_app_service_plan` resource, so it omitted $229.95 of costs. The rest of the output was identical though.

Here's the end of the output from regular `infracost breakdown`:

```
[ 31 lines skipped]
 module.XXX.azurerm_application_insights.this

 +- Data ingested                                                             Monthly cost depends on usage: $2.30 per GB


 module.XXX.azurerm_app_service_plan.service_plan

 +- Instance usage (P1v3)                                                                    730  hours
         $229.95


 OVERALL TOTAL
         $455.87
----------------------------------
23 cloud resources were detected, rerun with --show-skipped to see details:
· 5 were estimated, 4 include usage-based costs, see https://infracost.io/usage-file
· 11 weren't estimated, report them in https://github.com/infracost/infracost
· 7 were free
```

Here's the end of the output from `infracost breakdown --terraform-parse-hcl` (the beginning lines are the same):

```
[ 31 lines skipped ]
 module.XXX.azurerm_application_insights.this

 +- Data ingested                                                             Monthly cost depends on usage: $2.30 per GB


 OVERALL TOTAL
         $225.92
----------------------------------
23 cloud resources were detected, rerun with --show-skipped to see details:
· 5 were estimated, 4 include usage-based costs, see https://infracost.io/usage-file
· 11 weren't estimated, report them in https://github.com/infracost/infracost
· 7 were free
```

As you can see, the `module.XXX.azurerm_app_service_plan.service_plan` resource is entirely missing when using `--terraform-parse-hcl`.

tim775:
@bgrainger Thanks for the report, this is exactly the kind of thing we're hoping the community can help us find!  Is there any chance you could send a sanitized tf project with the issue so we can reproduce it?

bgrainger:
Here's a file that reproduces the problem:

```hcl
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~> 2.98"
    }
  }

  required_version = ">= 1.1.0"
}

provider "azurerm" {
  features {}

  subscription_id = "[redacted]"
}

resource "azurerm_resource_group" "this" {
  name     = "Issue1433"
  location = "westus2"
}

resource "azurerm_app_service_plan" "this" {
[...]

PR #1435, opened 2022-03-07
feat(azure): Set default capacity for Azure App Service Plan

### LINKED ISSUE #1384: Azure Dedicated App Service Plan with Function App Bug
It is still possible to use "dedicated" app service hosting plans such as "PremiumV2" (P1v2) for azure functions. Currently Infracost is skipping the App Service Plan pricing if it's being used for a function app. Because it is expecting an "EP" plan (example EP1)

Current Outcome:
Skips calculating cost for function.

- "Skipping resource azurerm_function_app.app. Could not find a way to get its cost components from the resource or usage file."

Skips calculating cost for app service plan.
- "Hiding cost with no usage: azurerm_app_service_plan.app_service_plan 'Instance usage (P1v2)'"

Expected Outcome:
If an App Service plan is deployed with "PremiumV2" and azure function is using the app service plan then infracost should calculate the price of the App Service plan not the price of the Functions.

Example Terraform:

```terraform
resource "azurerm_app_service_plan" "app_service_plan" {
  name                = var.app_service_plan_name
  resource_group_name = "resource-group"
  location            = "location"
  kind                = "linux"
  reserved            = true

  sku {
    tier = "PremiumV2"
    size = "P1v2"
  }
}
```

```terraform
resource "azurerm_function_app" "app" {
  name                       = "test-app"
  resource_group_name        = "resource-group"
  location                   = "location
  app_service_plan_id        = azurerm_app_service_plan.app_service_plan.id
  os_type                    = "linux"
  version                    = "~4"
  https_only                 = true
}
```

vdmgolub:
`capacity` is an optional argument. If not specified the resource is not
estimated. The pricing calculator starts with 1 instances, thus we
assume the default capacity would be 1 as well.

Fixes #1384 and #1433.

### DIFF
--- a/internal/providers/terraform/azure/app_service_plan.go
+++ b/internal/providers/terraform/azure/app_service_plan.go
@@ -21,7 +21,7 @@ func NewAzureRMAppServicePlan(d *schema.ResourceData, u *schema.UsageData) *sche
 	sku := d.Get("sku.0.size").String()
 	skuRefactor := ""
 	os := "windows"
-	capacity := d.Get("sku.0.capacity").Int()
+	capacity := d.GetInt64OrDefault("sku.0.capacity", 1)
 	productName := "Standard Plan"
 
 	// These are used by azurerm_function_app, their costs are calculated there as they don't have prices in the azurerm_app_service_plan resource
--- a/internal/providers/terraform/azure/testdata/app_service_plan_test/app_service_plan_test.golden
+++ b/internal/providers/terraform/azure/testdata/app_service_plan_test/app_service_plan_test.golden
@@ -1,31 +1,34 @@
 
- Name                                  Monthly Qty  Unit   Monthly Cost 
-                                                                        
- azurerm_app_service_plan.basic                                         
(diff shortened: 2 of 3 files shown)


internal/providers/terraform/azure/app_service_plan.go
```diff
@@ -22,6 +22,9 @@ func NewAzureRMAppServicePlan(d *schema.ResourceData, u *schema.UsageData) *sche
 	skuRefactor := ""
 	os := "windows"
 	capacity := d.Get("sku.0.capacity").Int()
+	if capacity == 0 {
+		capacity = 1
+	}
```
tim775, line 27: There was a new helper added with the HCL stuff so I think we can do:
```suggestion
	capacity := d.GetInt64OrDefault("sku.0.capacity", 1)
```

tim775 (approved):

internal/providers/terraform/azure/app_service_plan.go
vdmgolub: This is cool! Thanks! :)

PR #1438, opened 2022-03-08
fix(aws): Set engine for AWS DB Instance replica

vdmgolub:


internal/providers/terraform/azure/testdata/app_service_plan_test/app_service_plan_test.tf
```diff
@@ -114,7 +114,7 @@ resource "azurerm_app_service_plan" "default_capacity" {
   reserved            = false
 
   sku {
-    tier     = "Basic"
-    size     = "B2"
+    tier = "Basic"
+    size = "B2"
```
vdmgolub, line 118: Cleanup for my previous fix for Azure, not related to the fix.

hugorut (approved):
LGTM - have we made sure that this works with a `module` setup like the user described?

vdmgolub:
@hugorut Good point, let me verify that before merging 👍

PR #1444, opened 2022-03-09
feat: change release workflow to use internal script

hugorut:
Changes release workflow to use GO script in the repository. Tested and working as per this release https://github.com/hugorut/infracost/releases/tag/v0.9.24 (on personal fork)

* side note - i've introduced a `tools` directory with this, where I think we should move some of the codegen and other standalone go scripts so they don't pollute the main binary

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -20,20 +20,9 @@ jobs:
           make release
 
       - name: Release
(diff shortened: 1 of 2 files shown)


hugorut:
Converting to draft to remove the un-required file path abstraction

aliscott (approved):
LGTM!

PR #1453, opened 2022-03-11
fix: panic error if policy body is invalid

hugorut:
Solves issue if a valid policy file type is provided in `--policy-path` but the contents is malformed.


hugorut:
@aliscott makes sense to me

PR #1455, opened 2022-03-11
fix: Google SQL Database Instance default values

aliscott:
* When state was applied `replica_configuration` was set to an empty value, so this was adding a subresource for the replica when it shouldn't have been
* A public v4 IP is added to the DB instances by default


hugorut (approved):
🍾

PR #1456, opened 2022-03-13
fix(google): fix possible nil exceptions in network egress

sinabakh:


sinabakh:
@aliscott Ready for review

PR #1465, opened 2022-03-15
chore: Add telemetry for HCL resources cost discrepancies

vdmgolub:


cmd/infracost/run.go
```diff
@@ -802,14 +803,116 @@ func AddHCLEnvVars(projectContexts []*config.ProjectContext, r output.Root, hclR
...
+		for _, resource := range project.Resources {
+			acc[resource.Name] = resource.MonthlyCost
+		}
+
+		for _, hclResource := range hclProject.Resources {
+			resourceCost, ok := acc[hclResource.Name]
+			if !ok {
+				continue
+			}
+
```
hugorut, line 873: I would prefer here if we merged the logic from `hclMissingResources` so that we don't have create lookup maps and loops twice. i.e. here we could just add to the `missingList`. 

We can alter the return signature of the `hclResourceDiff` to return a struct with missing resources and discrepancies, e.g:

```go
type hclProjectDiff struct {
    resourceDiffs map[string][]string
    missingResources []string
}
```

cmd/infracost/run.go
```diff
@@ -802,14 +803,116 @@ func AddHCLEnvVars(projectContexts []*config.ProjectContext, r output.Root, hclR
...
+			if resourceCost != nil {
+				cost = *resourceCost
+
+				change = percentChange(hclCost, cost)
+				abs = change.Abs()
+			} else {
+				change = decimal.NewFromInt(100)
+				abs = decimal.NewFromInt(100)
+			}
+
```
hugorut, line 896:
```suggestion
				change := decimal.NewFromInt(100)
				abs := decimal.NewFromInt(100)

			if resourceCost != nil {
				cost = *resourceCost

				change = percentChange(hclCost, cost)
				abs = change.Abs()
			} 
```

hugorut (changes requested):
Nice, this looks good - minor request to merge the two functions to make the logic more efficient.

cmd/infracost/run.go
vdmgolub: Sounds good, I'll name it `hclRunDiff` though as it collects data from all projects. Sounds good?

hugorut: 👍

cmd/infracost/run.go
```diff
@@ -802,14 +808,95 @@ func AddHCLEnvVars(projectContexts []*config.ProjectContext, r output.Root, hclR
...
+	for _, project := range projects {
+		hclProject := hclProjectsMapping[project.Name]
+
+		if hclProject == nil {
+			log.Debugf("could not find a matching HCL project '%s' for HCL run diff", project.Name)
+			continue
+		}
+
+		hclResourcesMapping := map[string]*decimal.Decimal{}
+		hclResourceNames := map[string]bool{}
```
hugorut, line 834: this lookup map is not needed any more - you can just assert that the `hclResourcesMapping`, e.g.

```go
	if _, ok := hclResourcesMapping[resource.Name]; !ok {
		missingResources[resource.ResourceType] = true
                continue
	}
```

hugorut (changes requested):
almost there

hugorut (approved):
Looks good - just fixing the tests on master and then we can rebase that and check everything green

PR #1471, opened 2022-03-17
feat: add terraform-init-flags

hugorut:
* adds `terraform-ini-flags` flag which are parsed to `terraform init` if present
* adds golden file tests using `-plugin-dir` flag to test if the flags are parsed to init.


cmd/infracost/testdata/breakdown_init_flags_plugin_path/breakdown_init_flags_plugin_path.golden
```diff
@@ -0,0 +1,15 @@
...
+  Terraform command failed with:
+    Error: Required plugins are not installed
+    The installed provider plugins are not consistent with the packages selected
+    in the dependency lock file:
+      - registry.terraform.io/hashicorp/aws: there is no package for registry.terraform.io/hashicorp/aws 4.5.0 cached in .terraform/providers
+    Terraform uses external plugins to integrate with a variety of different
+    infrastructure services. You must install the required plugins before running
+    Terraform operations.
+
+Try setting the --path to a Terraform plan JSON file. See https://infracost.io/troubleshoot for how to generate this.
```
alikhajeh1, line 15: This is good as the troubleshooting page shows how they can init manually to workaround the issue. I'm wondering if it's possible to catch init errors and tell the user about the new flag? That would be even better as it'll save them from reading the troubleshooting page...

hugorut: @alikhajeh1 I've changed the init error so that we have a string like this https://github.com/infracost/infracost/pull/1471/files#diff-e5a4e07536e98f79ffd6f93bb121fdbd1c1e1c38267780a31e18b04076956a13R11 let me know what you think

alikhajeh1: @hugorut that's awesome - thanks!

aliscott (approved):
LGTM!

PR #1474, opened 2022-03-18
feat: Set log level to 'info' when CI is detected

### LINKED ISSUE #1450: Add environment variable to disable spinner output
Suggestion to not have spinner output when setting a variable like `INFRACOST_NO_SPINNER=true`

Example to avoid in Gitlab pipeline:
image

vdmgolub:
Fixes #1450.

This automatically disables spinners which currently cause issues
for some CI users. Infracost GitHub Actions already set logging by
default, this change unifies the behavior across all CI platforms.


aliscott (approved):
LGTM!

PR #1475, opened 2022-03-18
fix: hcl panic errors

hugorut:
* Fixes issue with `aws_eks_node_group` where  `instance_types` was not set https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group we now use the default `t3.medium`
* Fixes issue where `cty.Value` was unknown and causing panic when trying to set attribute


vdmgolub (approved):
👍

PR #1476, opened 2022-03-19
chore(google): Migrate Compute disk/image/snapshot resources to new structure

vdmgolub:


internal/providers/terraform/google/compute_disk.go
```diff
@@ -84,10 +53,38 @@ func computeDiskSize(d *schema.ResourceData) *decimal.Decimal {
...
+
+	if len(d.References("source_image")) > 0 {
+		return computeImageDiskSize(d.References("source_image")[0])
+	}
+
+	if len(d.References("source_snapshot")) > 0 {
+		return computeSnapshotDiskSize(d.References("source_snapshot")[0])
+	}
+
+	return 0
```
vdmgolub, line 81: Not sure if it is okay to treat zero as "empty" value here.

aliscott: I think it makes sense. It looks like we had a nil ptr risk before: https://github.com/infracost/infracost/blob/438c7a01f56ba6065af841573d1e3e3bb1336084/internal/providers/terraform/google/compute_disk.go#L49. So this solves that.

aliscott (approved):
This looks great!

PR #1477, opened 2022-03-21
feat: Update breakdown/diff summary wording

### LINKED ISSUE #1459: Disable "Add cost estimates to your pull requests: https://infracost.io/cicd" output
Hello everyone, 

Like written in this issue, I would to know if it is possible to disable the output : `Add cost estimates to your pull requests: https://infracost.io/cicd` when I am launching infracost from my computer. 

Of course, I know that this output is disabled when infracost is launched in a CI/CD. 
I saw in this [issue](https://github.com/infracost/infracost/issues/1380) that you added this output to let people knows that adding infracost `comments` in CI/CD was possible, but it would be great to be able to disable it when we want to launch infracost from computers (like during the development of a new module or new infra).

An environment variable such as `INFRACOST_SKIP_CI_MESSAGE` would be great (same idea as `INFRACOST_SKIP_UPDATE_CHECK` for updates)

Have a nice day !

vdmgolub:
Resolves  #1459.


internal/output/output.go
```diff
@@ -346,27 +344,22 @@ func (r *Root) summaryMessage(showSkipped bool) string {
...
 
+		allUsageBased := *r.Summary.TotalUsageBasedResources == *r.Summary.TotalSupportedResources
+
 		if r.Summary.TotalUsageBasedResources != nil && *r.Summary.TotalUsageBasedResources > 0 {
-			usageLink := ui.SecondaryLinkString("https://infracost.io/usage-file")
-			if *r.Summary.TotalUsageBasedResources == 1 {
-				msg += fmt.Sprintf(", 1 includes usage-based costs, see %s", usageLink)
-			} else {
-				msg += fmt.Sprintf(", %d include usage-based costs, see %s", *r.Summary.TotalUsageBasedResources, usageLink)
+			usageBasedCount := "from which 1 includes"
```
aliscott, line 350: "from which" or "of which"? I think "of which" is correct because you'd say "two of them".

aliscott (approved):
LGTM

internal/output/output.go
vdmgolub: English is hard :) In this case we can put the number first, right? `1 of which includes`/`20 of which include`.

PR #1479, opened 2022-03-21
feat: add codeql vulnerability scanning

hugorut:
Adds codeql workflow which is run on pr, master and on a cron schedule (weekly)

### DIFF
--- a/.github/workflows/code-scanning.yml
+++ b/.github/workflows/code-scanning.yml
@@ -6,8 +6,8 @@ on:
   pull_request:
     branches: [ master ]
   schedule:
-    # run 00:00 every Monday
-    - cron: '00 00 * * 1'
+    # run 05:00 every Monday
+    - cron: '00 05 * * 1'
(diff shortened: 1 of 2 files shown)


.github/workflows/code-scanning.yml
```diff
@@ -0,0 +1,36 @@
+name: Scan code for vulnerabilities
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    # run 00:00 every Monday
+    - cron: '00 00 * * 1'
```
aliscott, line 10: Can we change this to 05:00 on Monday - that way there will be less time between a reported issue and us actioning it.

aliscott (approved):
LGTM

.github/workflows/code-scanning.yml
hugorut: 👍

PR #1492, opened 2022-03-25
feat: Add google_service_networking_connection

### LINKED ISSUE #1360: Add support for `google_service_networking_connection` resource
Terraform resource: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/service_networking_connection

tim775:
## Objective:

Add support for google_service_networking_connection. Fixes #1360.

## Pricing details:

Private service connections are essentially VPC to VPC connections and and can generate "internal" egress costs in the same manner as a VPN Gateway.  Therefore this resource has the same usage cost components for egress traffic as VPN gateway:
```
    monthly_egress_data_transfer_gb: # Monthly VM-VM data transfer from VPN gateway to the following, in GB:
      same_region: 250                # VMs in the same Google Cloud region.
      us_or_canada: 100               # From a Google Cloud region in the US or Canada to another Google Cloud region in the US or Canada.
      europe: 70                      # Between Google Cloud regions within Europe.
      asia: 50                        # Between Google Cloud regions within Asia.
      south_america: 100              # Between Google Cloud regions within South America.
      oceania: 50                     # Indonesia and Oceania to/from any Google Cloud region.
      worldwide: 200                  # to a Google Cloud region on another continent.
```

## Status:

- [ ] ~~Updated the internal/resources file~~
- [ ] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/pull/191)

### DIFF
--- a/internal/providers/terraform/google/service_networking_connection.go
+++ b/internal/providers/terraform/google/service_networking_connection.go
@@ -1,0 +1,22 @@
+package google
+
+import (
+	"github.com/infracost/infracost/internal/schema"
+)
+
+func getServiceNetworkingConnectionRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "google_service_networking_connection",
+		RFunc: newServiceNetworkingConnection,
+	}
+}
+
+func newServiceNetworkingConnection(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := d.Get("region").String()
+	return &schema.Resource{
+		Name: d.Address,
+		SubResources: []*schema.Resource{
+			networkEgress(region, u, "Network egress", "Traffic", ComputeVPNGateway),
+		},
+	}
+}
--- a/internal/providers/terraform/google/service_networking_connection_test.go
+++ b/internal/providers/terraform/google/service_networking_connection_test.go
@@ -1,0 +1,16 @@
+package google_test
+
+import (
+	"testing"
+
+	"github.com/infracost/infracost/internal/providers/terraform/tftest"
+)
+
+func TestServiceNetworkingConnection(t *testing.T) {
+	t.Parallel()
+	if testing.Short() {
+		t.Skip("skipping test in short mode")
+	}
+
+	tftest.GoldenFileResourceTests(t, "service_networking_connection_test")
+}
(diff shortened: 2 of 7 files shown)


internal/providers/terraform/google/service_networking_connection_test.go
```diff
@@ -0,0 +1,16 @@
+package google_test
+
+import (
+	"testing"
+
+	"github.com/infracost/infracost/internal/providers/terraform/tftest"
+)
+
+func TestServiceNetworkingConnection(t *testing.T) {
```
vdmgolub, line 9: FYI resourcegen command generates tests having `GoldenFile` postfix. I wonder if we should do that or actually drop it. What do you think? It makes the test explicit, but may be redundant, as it's still a regular test.
```suggestion
func TestServiceNetworkingConnectionGoldenFile(t *testing.T) {
```

internal/providers/terraform/google/service_networking_connection.go
```diff
@@ -0,0 +1,22 @@
...
+		RFunc: newServiceNetworkingConnection,
+	}
+}
+
+func newServiceNetworkingConnection(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := d.Get("region").String()
+	return &schema.Resource{
+		Name: d.Address,
+		SubResources: []*schema.Resource{
+			networkEgress(region, u, "Network egress", "Traffic", ComputeVPNGateway),
```
vdmgolub, line 19: We could copy this function to `internal/resources/google` for now and introduce a new resource there. It may help someone who'll be migrating the other related resources.

vdmgolub (approved):

vdmgolub:
FYI, fixes for Azure tests are here: https://github.com/infracost/infracost/pull/1493

internal/providers/terraform/google/service_networking_connection_test.go
tim775: Oops, I'll stick with the convention for now.  I guess the suffix was more important when we were stilling converting everything to golden file, so it should be ok to remove now if we want.

internal/providers/terraform/google/service_networking_connection.go
tim775: Yeah I considered that and I think we should but I noticed that @sinabakh has already started that process recently (~2 weeks) so I figured I see how that plays out.

P.S. Thanks for the networkEgress work/refactor @sinabakh, it made adding this resource super easy!

internal/providers/terraform/google/service_networking_connection_test.go
vdmgolub: Sounds good, I'll change the resourcegen then. 👍

PR #1494, opened 2022-03-25
fix: get the correct instance type for EKS node groups

aliscott:
This fixes two issues:
1. When the instance type is not specified and the Terraform is applied, it was being set as an empty string instead of null so we weren't using the default.
2. The instance type specified in the EKS node group wasn't overriding the instance type in the launch template


vdmgolub (approved):
👍

PR #1498, opened 2022-03-28
chore: Update test naming in resourcegen templates

vdmgolub:
As we test resources using golden files by default, having `GoldenFile`
in the test names seems redundant.


vdmgolub:
@tim775 As discussed in your PR.

hugorut (approved):
👍

PR #1500, opened 2022-03-28
fix: use backwards refs to determine if eip has been allocated

hugorut:
Fixes HCL issue where EIP was being marked as unallocated. Now as well as checking properties that are populated from the cloud API we use the HCL references to infer if the EIP is in use.


internal/providers/terraform/parser.go
```diff
@@ -493,7 +493,13 @@ func (p *Parser) parseReferences(resData map[string]*schema.ResourceData, conf g
 func (p *Parser) parseConfReferences(resData map[string]*schema.ResourceData, conf gjson.Result, d *schema.ResourceData, attr string, registryMap *ResourceRegistryMap) bool {
 	// Check if there's a reference in the conf
 	resConf := getConfJSON(conf, d.Address)
-	refResults := resConf.Get("expressions").Get(attr).Get("references").Array()
+	exps := resConf.Get("expressions").Get(attr)
+	lookupStr := "references"
+	if exps.IsArray() {
+		lookupStr = "#.references"
+	}
```
hugorut, line 500: This fixes issue where nested blocks which contained references could not be parsed. Using the following expression JSON

```json
  "expressions": {
      "subnet_mapping": [
        {
          "allocation_id": {
            "references": [
              "aws_eip.eip4.id",
              "aws_eip.eip4"
            ]
          },
        }
      ]
    }
```

If a reference of `subnet_mapping.#.allocation_id` would fail to return the references.

internal/schema/resource_data.go
```diff
@@ -78,8 +78,14 @@ func (d *ResourceData) IsEmpty(key string) bool {
 	return g.Type == gjson.Null || len(g.Raw) == 0 || g.Raw == "\"\"" || emptyObjectOrArray(g)
 }
 
-func (d *ResourceData) References(key string) []*ResourceData {
-	return d.referencesMap[key]
+func (d *ResourceData) References(keys ...string) []*ResourceData {
```
hugorut, line 81: Changed to provide simpler way to fetch more than one reference

aliscott: 🙌

aliscott (approved):
LGTM

PR #1506, opened 2022-03-30
fix: Honor cache enabled flag for API Gateway

### LINKED ISSUE #1505: AWS cache memory billed for API Gateway even when disabled.
Hi there, 

Great project, thanks for making it, it seems to give us quite a bit of info. I've tried it on one of our new projects, and it seems there might be a small bug in the cost estimation.

We have an AWS API Gateway stage resource, which doesn't have caching enabled (this can be seen in the costs breakdown too). However, still the costs are calculated in the overview.

This is our config:

```
resource "aws_api_gateway_stage" "heartbeat_stage" {
  deployment_id = aws_api_gateway_deployment.heartbeat_deployment.id
  description   = "${terraform.workspace}-gateway-stage"
  rest_api_id   = aws_api_gateway_rest_api.heartbeat_api.id
  stage_name    = terraform.workspace

  depends_on = [aws_api_gateway_deployment.heartbeat_deployment]

  tags = var.custom_tags
}
```

And this is the (partial) output:

```
 Name                                                                                Monthly Qty  Unit                  Monthly Cost 
                                                                                                                                     
 module.apigateway.aws_api_gateway_rest_api.heartbeat_api                                                                    
 └─ Requests (first 333M)                                                       Monthly cost depends on usage: $3.50 per 1M requests 
                                                                                                                                     
 module.apigateway.aws_api_gateway_stage.heartbeat_stage                                                                     
 └─ Cache memory (0 GB)                                                                      730  hours                       $14.60 
                                                                                                                                     
```

So, it seems the cache memory is detected correctly, but still the monthly costs is calculated. We also tried explicitly disabling the caching (even though the default is disabled), but that has no effect on the costs.

tim775:
Fixes #1505 - Api Gateway should not estimate charges for cache unless the cache is enabled.

### DIFF
--- a/internal/providers/terraform/aws/api_gateway_stage.go
+++ b/internal/providers/terraform/aws/api_gateway_stage.go
@@ -16,6 +16,7 @@ func NewAPIGatewayStage(d *schema.ResourceData, u *schema.UsageData) *schema.Res
 		Address:          d.Address,
 		Region:           d.Get("region").String(),
 		CacheClusterSize: d.Get("cache_cluster_size").Float(),
+		CacheEnabled:     d.GetBoolOrDefault("cache_cluster_enabled", false),
 	}
 
 	r.PopulateUsage(u)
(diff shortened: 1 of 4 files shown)


aliscott (approved):
LGTM

PR #1529, opened 2022-04-03
refactor(google): resource migrations batch 11

sinabakh:
Migrate sql_database_instance alongside few case fix in registry file.

### DIFF
--- a/internal/providers/terraform/google/sql_database_instance.go
+++ b/internal/providers/terraform/google/sql_database_instance.go
@@ -1,305 +1,38 @@
 package google
 
 import (
-	"fmt"
-	"strconv"
-	"strings"
-
+	"github.com/infracost/infracost/internal/resources/google"
 	"github.com/infracost/infracost/internal/schema"
-	"github.com/shopspring/decimal"
 )
 
-type SQLInstanceDBType int
-
-const (
-	MySQL SQLInstanceDBType = iota
-	PostgreSQL
-	SQLServer
-)
-
-func GetSQLInstanceRegistryItem() *schema.RegistryItem {
+func getSQLInstanceRegistryItem() *schema.RegistryItem {
 	return &schema.RegistryItem{
 		Name:  "google_sql_database_instance",
-		RFunc: NewSQLInstance,
+		RFunc: NewSQLDatabaseInstance,
 		Notes: []string{
 			"Cloud SQL network, SQL Server license, 1-3 years commitments costs are not yet supported.",
 		},
 	}
 }
-
-func NewSQLInstance(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
-	var resource *schema.Resource
-
-	replica := false
+func NewSQLDatabaseInstance(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	r := &google.SQLDatabaseInstance{Address: d.Address, DatabaseVersion: d.Get("database_version").String(), Tier: d.Get("settings.0.tier").String(), Region: d.Get("region").String()}
+	if !d.IsEmpty("settings.0.ip_configuration.0.ipv4_enabled") {
+		r.UseIPV4 = d.Get("settings.0.ip_configuration.0.ipv4_enabled").Bool()
+	} else {
+		r.UseIPV4 = true // Should use ipv4 if the ipv4_enabled is empty
+	}
 	if !d.IsEmpty("replica_configuration") {
-		replica = true
+		r.ReplicaConfiguration = d.Get("replica_configuration").String()
 	}
-
-	resource = sqlDatabaseInstanceCostComponents(d, u, false, d.Address)
-	if replica {
-		resource.SubResources = append(resource.SubResources, sqlDatabaseInstanceCostComponents(d, u, true, "Replica"))
+	if !d.IsEmpty("settings.0.availability_type") {
+		r.AvailabilityType = d.Get("settings.0.availability_type").String()
 	}
-
-	return resource
-}
-
-func sqlDatabaseInstanceCostComponents(d *schema.ResourceData, u *schema.UsageData, replica bool, name string) *schema.Resource {
-	var costComponents []*schema.CostComponent
-	tier := d.Get("settings.0").Get("tier").String()
-
-	availabilityType := "ZONAL"
-	if d.Get("settings.0").Get("availability_type").Exists() && !replica {
-		availabilityType = d.Get("settings.0").Get("availability_type").String()
+	if !d.IsEmpty("settings.0.disk_type") {
+		r.DiskType = d.Get("settings.0.disk_type").String()
 	}
-
-	region := d.Get("region").String()
-	dbVersion := d.Get("database_version").String()
-	dbType := sqlInstanceDBVersionToDBType(dbVersion)
-
-	diskType := "PD_SSD"
-	if d.Get("settings.0").Get("disk_type").Exists() {
-		diskType = d.Get("settings.0").Get("disk_type").String()
+	if !d.IsEmpty("settings.0.disk_size") {
+		r.DiskSize = d.Get("settings.0.disk_size").Int()
 	}
-
-	var diskSizeGB int64 = 10
-	if d.Get("settings.0").Get("disk_size").Exists() {
-		diskSizeGB = d.Get("settings.0").Get("disk_size").Int()
-	}
-
-	if sqlInstanceTierToResourceGroup(tier) != "" && dbType != SQLServer {
-		costComponents = append(costComponents, sharedSQLInstance(tier, availabilityType, dbType, region))
-	} else if sqlInstanceTierToResourceGroup(tier) == "" && strings.Contains(tier, "db-custom-") {
-		cpu, _ := strconv.ParseInt(strings.Split(tier, "-")[2], 10, 32)
-		vCPU := decimalPtr(decimal.NewFromInt32(int32(cpu)))
-
-		costComponents = append(costComponents, cpuCostComponent(region, tier, availabilityType, dbType, vCPU))
-
-		ram, _ := strconv.ParseInt(strings.Split(tier, "-")[3], 10, 32)
-		memory := decimalPtr(decimal.NewFromInt32(int32(ram)).Div(decimal.NewFromInt(1024)))
-
-		costComponents = append(costComponents, memoryCostComponent(region, tier, availabilityType, dbType, memory))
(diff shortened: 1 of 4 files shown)


sinabakh:
The failing test is not related to the refactor

internal/resources/google/sql_database_instance.go
```diff
@@ -0,0 +1,316 @@
...
+
+type SQLInstanceDBType int
+
+const (
+	MySQL SQLInstanceDBType = iota
+	PostgreSQL
+	SQLServer
+)
+
+func sqlDatabaseInstanceCostComponents(r *SQLDatabaseInstance, replica bool, name string) *schema.Resource {
```
aliscott, line 57: I think this can be a method on the struct? Looks like there's nothing else using it.

internal/providers/terraform/google/sql_database_instance.go
```diff
@@ -1,305 +1,38 @@
...
-	description := fmt.Sprintf("%s: %s", dbTypeString, availabilityTypeString)
-
-	return description
-}
-
-func sqlInstanceStorage(region string, dbType SQLInstanceDBType, availabilityType, diskType string, diskSizeGB int64) *schema.CostComponent {
-	diskTypeHumanReadableNames := map[string]string{
-		"PD_SSD": "SSD",
-		"PD_HDD": "HDD",
+	if !d.IsEmpty("replica_configuration") {
```
aliscott, line 24: Does this need the check or can it just be included when creating the struct since it defaults to empty anyway?

internal/providers/terraform/google/sql_database_instance.go
```diff
@@ -1,305 +1,38 @@
...
+		r.ReplicaConfiguration = d.Get("replica_configuration").String()
 	}
-
-	diskTypeAPIResourceGroup := map[string]string{
-		"PD_SSD": "SSD",
-		"PD_HDD": "PDStandard",
+	if !d.IsEmpty("settings.0.availability_type") {
```
aliscott, line 27: As above, I don't think this empty check is needed.

internal/providers/terraform/google/sql_database_instance.go
```diff
@@ -1,305 +1,38 @@
...
+		r.AvailabilityType = d.Get("settings.0.availability_type").String()
 	}
-
-	if dbType == SQLServer {
-		diskType = "PD_SSD"
+	if !d.IsEmpty("settings.0.disk_type") {
```
aliscott, line 30: As above, I don't think this empty check is needed.

internal/providers/terraform/google/sql_database_instance.go
```diff
@@ -1,305 +1,38 @@
...
-		ProductFilter: &schema.ProductFilter{
-			VendorName: strPtr("gcp"),
-			Region:     strPtr("global"),
-			Service:    strPtr("Cloud SQL"),
-			AttributeFilters: []*schema.AttributeFilter{
-				{Key: "resourceGroup", Value: strPtr("IpAddress")},
-				{Key: "description", ValueRegex: strPtr("/IP address idling - hour/")},
-			},
-		},
+	if !d.IsEmpty("settings.0.disk_size") {
```
aliscott, line 33: As above, I don't think this empty check is needed, since `.Int()` will default to 0, which is what `DiskSize` defaults to anyway.

internal/resources/google/sql_database_instance.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+	}
+
+	return &schema.Resource{
+		Name:           name,
+		CostComponents: costComponents,
+	}
+}
+
+func memoryCostComponent(region string, tier string, availabilityType string, dbType SQLInstanceDBType, memory *decimal.Decimal) *schema.CostComponent {
```
aliscott, line 116: Can all the below functions be on the struct so there's no risk of conflict on the package level? If they can then I think we can tidy up the arguments that are passed to it, since most of them soon be available as struct properties.

aliscott (commented):
Thanks @sinabakh! 🚀

internal/providers/terraform/google/sql_database_instance.go
```diff
@@ -1,305 +1,38 @@
...
-type SQLInstanceDBType int
-
-const (
-	MySQL SQLInstanceDBType = iota
-	PostgreSQL
-	SQLServer
 )
 
-func GetSQLInstanceRegistryItem() *schema.RegistryItem {
+func getSQLInstanceRegistryItem() *schema.RegistryItem {
```
aliscott, line 8:
```suggestion
func getSQLDatabaseInstanceRegistryItem() *schema.RegistryItem {
```

internal/resources/google/sql_database_instance.go
sinabakh: Right, fixed.

internal/providers/terraform/google/sql_database_instance.go
sinabakh: Nice catch! Fixed it. I think I might have missed it in previous PRs so we might be able to fix them later.

internal/providers/terraform/google/sql_database_instance.go
sinabakh: Applied

internal/providers/terraform/google/sql_database_instance.go
sinabakh: Applied

internal/providers/terraform/google/sql_database_instance.go
sinabakh: Applied

internal/resources/google/sql_database_instance.go
sinabakh: Great idea, applied and removed the redundant params.

internal/providers/terraform/google/sql_database_instance.go
sinabakh: applied

sinabakh:
@aliscott All comments are resolved and ready for review.

internal/resources/google/sql_database_instance.go
```diff
@@ -0,0 +1,316 @@
...
+)
+
+func (r *SQLDatabaseInstance) sqlDatabaseInstanceCostComponents(replica bool, name string) *schema.Resource {
+	var costComponents []*schema.CostComponent
+	tier := r.Tier
+
+	availabilityType := "ZONAL"
+	if r.AvailabilityType != "" && !replica {
+		availabilityType = r.AvailabilityType
+	}
```
aliscott, line 64:
```suggestion
	if r.AvailabilityType == "" || replica {
		r.AvailabilityType = "ZONAL"
	}
```
Should we change this so it sets the default struct property? That way we don't need to pass it to the the methods like `r.memoryCostComponent` and `r.cpuCostComponent` and we can just access it directly from them?

internal/resources/google/sql_database_instance.go
```diff
@@ -0,0 +1,316 @@
...
+			Service:       strPtr("Cloud SQL"),
+			ProductFamily: strPtr("ApplicationServices"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "description", ValueRegex: strPtr(description)},
+			},
+		},
+	}
+}
+
+func cpuCostComponent(region string, tier string, availabilityType string, dbType SQLInstanceDBType, vCPU *decimal.Decimal) *schema.CostComponent {
```
aliscott, line 138: I think this function and all the functions below should be methods on the struct. That way they don't pollute the package namespace and we don't need to pass `region`, `tier` and `availabilityType` as args.

internal/resources/google/sql_database_instance.go
```diff
@@ -0,0 +1,316 @@
...
+	tier := r.Tier
+
+	availabilityType := "ZONAL"
+	if r.AvailabilityType != "" && !replica {
+		availabilityType = r.AvailabilityType
+	}
+
+	region := r.Region
+	dbVersion := r.DatabaseVersion
+	dbType := sqlInstanceDBVersionToDBType(dbVersion)
```
aliscott, line 68: We could make this a struct method `r.dbType()` so we don't need to pass it around to all the other methods.

internal/resources/google/sql_database_instance.go
sinabakh: Nice idea! I change this and few other fields like this.

internal/resources/google/sql_database_instance.go
sinabakh: Applied

internal/resources/google/sql_database_instance.go
sinabakh: Applied

internal/resources/google/sql_database_instance.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	if r.DiskSize == 0 {
+		r.DiskSize = 10
+	}
+
+	if r.sqlInstanceTierToResourceGroup() != "" && r.dbType() != SQLServer {
+		costComponents = append(costComponents, r.sharedSQLInstance())
+	} else if r.sqlInstanceTierToResourceGroup() == "" && strings.Contains(tier, "db-custom-") {
+		cpu, _ := strconv.ParseInt(strings.Split(tier, "-")[2], 10, 32)
```
aliscott, line 80: Should we check for errors here? I'm assuming `[2]` won't cause any issues because we validate there's 2 `-`s in the string above, but the `ParseInt` could cause an error.

internal/resources/google/sql_database_instance.go
```diff
@@ -0,0 +1,314 @@
...
+		vCPU := decimalPtr(decimal.NewFromInt32(int32(cpu)))
+
+		costComponents = append(costComponents, r.cpuCostComponent(vCPU))
+
+		ram, _ := strconv.ParseInt(strings.Split(tier, "-")[3], 10, 32)
```
aliscott, line 85: Should we check for errors here as well? Looks like the `[3]` could cause an `index out of range` panic and the `ParseInt` could return an error.

aliscott (commented):
Looks good, thanks @sinabakh . Two small comments about error checking. Let me know what you think.

internal/resources/google/sql_database_instance.go
sinabakh: Fixed

aliscott (approved):
LGTM

aliscott:
@sinabakh I've rebased this and fixed a conflict so it can be merged.

Issue #1524, opened 2022-04-01
HCL parsing does not work with provider aliases


aliscott:
This might be related to the $0 prices shown in https://github.com/infracost/infracost/issues/1446 but I've been unable to reproduce that yet.

**Terraform code:**

```hcl
provider "aws" {
  region = "ap-northeast-1"
}

provider "aws" {
  region = "eu-west-2"
  alias = "europe"
}

# Storage price should be $0.025 per GB for ap-northeast-1
resource "aws_s3_bucket" "bucket-1" {
  bucket = "alistair-test-438ygth-bucket-1"
}

# Storage price should be $0.024 per GB for eu-west-2
resource "aws_s3_bucket" "bucket-2" {
  provider = aws.europe
  bucket = "alistair-test-09tgri-bucket-1"
}
```

**Expected result (what running against the Terraform plan JSON shows):**

```
 aws_s3_bucket.bucket-1
 └─ Standard
    ├─ Storage                              Monthly cost depends on usage: $0.025 per GB
    ├─ PUT, COPY, POST, LIST requests       Monthly cost depends on usage: $0.0047 per 1k requests
    ├─ GET, SELECT, and all other requests  Monthly cost depends on usage: $0.00037 per 1k requests
    ├─ Select data scanned                  Monthly cost depends on usage: $0.00225 per GB
    └─ Select data returned                 Monthly cost depends on usage: $0.0008 per GB

 aws_s3_bucket.bucket-2
 └─ Standard
    ├─ Storage                              Monthly cost depends on usage: $0.024 per GB
    ├─ PUT, COPY, POST, LIST requests       Monthly cost depends on usage: $0.0053 per 1k requests
    ├─ GET, SELECT, and all other requests  Monthly cost depends on usage: $0.00042 per 1k requests
    ├─ Select data scanned                  Monthly cost depends on usage: $0.00225 per GB
    └─ Select data returned                 Monthly cost depends on usage: $0.0008 per GB

 OVERALL TOTAL                                                                                $0.00
```

**Actual result:**

```
 aws_s3_bucket.bucket-1
 └─ Standard
    ├─ Storage                              Monthly cost depends on usage: $0.025 per GB
    ├─ PUT, COPY, POST, LIST requests       Monthly cost depends on usage: $0.0047 per 1k requests
    ├─ GET, SELECT, and all other requests  Monthly cost depends on usage: $0.00037 per 1k requests
[...]

PR #1532, opened 2022-04-04
fix: provider alias references

hugorut:
Closes: #1524

Fixes issues in HCL where provider aliases were not being resolved as they were being parsed (correctly) as references in HCL


aliscott (approved):
LGTM

PR #1533, opened 2022-04-04
fix: Detect ECS Fargate launch type from capacity providers

### LINKED ISSUE #1349: `aws_ecs_service.capacity_provider_strategy` is not parsed as FARGATE deployment type
I am deploying a service and the prices that infracost is giving out seem wrong.
The main issue is it seems to think that the `aws_ecs_service` is free. Am I doing something wrong here or expecting something that cannot be done?
The resource is listed as a paid service here - https://www.infracost.io/docs/supported_resources/aws/

---

*additional context added after investigation*

Infracost uses `aws_ecs_service.launch_type` to determine what type of deployment the ecs service is (EC2/FARGATE). EC2 deployments are prices through `auto_scaling_group` and `aws_ecs_service` is marked as free.

When `aws_ecs_service.launch_type` is not specified it defaults to EC2 even if `capacity_provider_strategy` is set to a FARGATE type strategy.

We should change the `aws_ecs_service` to check `capacity_provider_strategy` block before determining the deployment type

tim775:
Fixes #1349

### DIFF
--- a/internal/providers/terraform/aws/ecs_service.go
+++ b/internal/providers/terraform/aws/ecs_service.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"github.com/tidwall/gjson"
 	"regexp"
 	"strconv"
 	"strings"
@@ -13,7 +14,7 @@ func getECSServiceRegistryItem() *schema.RegistryItem {
 	return &schema.RegistryItem{
 		Name:                "aws_ecs_service",
 		RFunc:               NewECSService,
-		ReferenceAttributes: []string{"task_definition"},
+		ReferenceAttributes: []string{"cluster", "task_definition"},
 	}
 }
 
@@ -36,7 +37,7 @@ func NewECSService(d *schema.ResourceData, u *schema.UsageData) *schema.Resource
 	r := &aws.ECSService{
 		Address:                        d.Address,
 		Region:                         d.Get("region").String(),
-		LaunchType:                     d.Get("launch_type").String(),
+		LaunchType:                     calcLaunchType(d),
 		DesiredCount:                   d.Get("desired_count").Int(),
 		MemoryGB:                       memoryGB,
 		VCPU:                           vcpu,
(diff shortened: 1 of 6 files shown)


internal/providers/terraform/aws/ecs_service.go
```diff
@@ -47,6 +48,88 @@ func NewECSService(d *schema.ResourceData, u *schema.UsageData) *schema.Resource
...
+	if launchType != "" {
+		return launchType
+	}
+
+	clusterRefs := d.References("cluster")
+	if len(clusterRefs) > 0 {
+		cluster := clusterRefs[0]
+
+		// check the cluster for a default capacity provider
+
```
hugorut, line 74:
```suggestion
		// check the cluster for a default capacity provider
```

hugorut, line 74:
```suggestion
		// check the cluster for a default capacity provider
```

internal/providers/terraform/aws/ecs_service.go
```diff
@@ -47,6 +48,88 @@ func NewECSService(d *schema.ResourceData, u *schema.UsageData) *schema.Resource
...
+		for _, capProvider := range cluster.References("aws_ecs_cluster_capacity_providers.cluster_name") {
+			defaultStrategies := capProvider.Get("default_capacity_provider_strategy").Array()
+			if len(defaultStrategies) > 0 {
+				lt := getCapacityProviderLaunchType(defaultStrategies)
+				if lt == "FARGATE" {
+					return lt
+				}
+				if lt != "" {
+					launchType = lt
+				}
```
hugorut, line 99:
```suggestion
				if lt != "" {
					return lt
				}
```

hugorut (approved):
🚀

internal/providers/terraform/aws/ecs_service.go
tim775: I think this suggestion isn't right because there can be multiple `aws_ecs_cluster_capacity_providers` and we want to keep looking to see if at least one supports fargate.

hugorut: makes sense

internal/providers/terraform/aws/ecs_service.go
hugorut: just new line removal - but no biggie

PR #1539, opened 2022-04-05
feat(azure): Add MySQL Flexible Server resources

### LINKED ISSUE #792: Add support for Azure Database for MySQL Flexible Server
Resource information: https://docs.microsoft.com/en-gb/azure/mysql/flexible-server/
Pricing information: https://azure.microsoft.com/en-gb/pricing/details/mysql/flexible-server/

### Paid resource

`azurerm_mysql_flexible_server`

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/mysql_flexible_server

Cost components:

| Name | Unit | Usage param | Notes|
| -- | -- | -- | -- |
| `Compute (<sku name>)` | `hours` | | ex. `Compute (B_Standard_B1ms)` |
| `Storage` | `GB` | | If not specified, defaults to 20GB minimum |
| `Additional IOPS` | `IOPS` | | Comes from TF resource, https://docs.microsoft.com/en-us/azure/mysql/flexible-server/concepts-compute-storage#iops. If not specified, defaults to 360 minimum. Also, first 360 IOPS are free, show only  the additional value `IOPS - 360` |
| `Additional backup storage` | `GB` | `monthly_additional_backup_storage_gb` | |

### Free resources

- `azurerm_mysql_flexible_database`
- `azurerm_mysql_flexible_server_configuration`
- `azurerm_mysql_flexible_server_firewall_rule`

vdmgolub:
## Objective:

Add support for Azure MySQL Flexible Server. Fixes #792.

## Pricing details:

https://azure.microsoft.com/en-gb/pricing/details/mysql/flexible-server/

## Status:

- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/pull/198)

## Issues:

None

### DIFF
--- a/internal/providers/terraform/azure/mysql_flexible_server.go
+++ b/internal/providers/terraform/azure/mysql_flexible_server.go
@@ -1,0 +1,70 @@
+package azure
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/infracost/infracost/internal/resources/azure"
+	"github.com/infracost/infracost/internal/schema"
+	log "github.com/sirupsen/logrus"
+)
+
+func getMySQLFlexibleServerRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_mysql_flexible_server",
+		RFunc: newMySQLFlexibleServer,
+	}
+}
+
+func newMySQLFlexibleServer(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := d.Get("location").String()
+	sku := d.Get("sku_name").String()
+	storage := d.GetInt64OrDefault("storage.0.size_gb", 0)
+	iops := d.GetInt64OrDefault("storage.0.iops", 0)
+
+	var tier, size, version string
+
+	s := strings.Split(sku, "_")
+	if len(s) < 3 || len(s) > 4 {
+		log.Warnf("Unrecognised MySQL Flexible Server SKU format for resource %s: %s", d.Address, sku)
+		return nil
+	}
+
+	if len(s) > 2 {
+		tier = strings.ToLower(s[0])
+		size = s[2]
+	}
+
+	if len(s) > 3 {
+		version = s[3]
+	}
+
+	supportedTiers := []string{"b", "gp", "mo"}
+	if !contains(supportedTiers, tier) {
+		log.Warnf("Unrecognised MySQL Flexible Server tier prefix for resource %s: %s", d.Address, sku)
+		return nil
+	}
+
+	if tier != "b" {
+		coreRegex := regexp.MustCompile(`(\d+)`)
+		match := coreRegex.FindStringSubmatch(size)
+		if len(match) < 1 {
+			log.Warnf("Unrecognised MySQL Flexible Server size for resource %s: %s", d.Address, sku)
+			return nil
+		}
+	}
+
+	r := &azure.MySQLFlexibleServer{
+		Address:         d.Address,
+		Region:          region,
+		SKU:             sku,
+		Tier:            tier,
+		InstanceType:    size,
+		InstanceVersion: version,
+		Storage:         storage,
+		IOPS:            iops,
+	}
+	r.PopulateUsage(u)
+
+	return r.BuildResource()
+}
--- a/internal/providers/terraform/azure/postgresql_flexible_server.go
+++ b/internal/providers/terraform/azure/postgresql_flexible_server.go
@@ -1,30 +1,27 @@
 package azure
 
 import (
-	"fmt"
 	"regexp"
 	"strings"
 
+	"github.com/infracost/infracost/internal/resources/azure"
 	"github.com/infracost/infracost/internal/schema"
 	log "github.com/sirupsen/logrus"
-	"github.com/tidwall/gjson"
-
-	"github.com/shopspring/decimal"
 )
 
-func GetAzureRMPostgreSQLFlexibleServerRegistryItem() *schema.RegistryItem {
+func getAzureRMPostgreSQLFlexibleServerRegistryItem() *schema.RegistryItem {
 	return &schema.RegistryItem{
 		Name:  "azurerm_postgresql_flexible_server",
-		RFunc: NewAzureRMPostrgreSQLFlexibleServer,
+		RFunc: newPostgreSQLFlexibleServer,
 	}
 }
 
-func NewAzureRMPostrgreSQLFlexibleServer(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
-	var costComponents []*schema.CostComponent
-
+func newPostgreSQLFlexibleServer(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
 	region := d.Get("location").String()
 	sku := d.Get("sku_name").String()
-	var tier, size, skuName, meterName, version, series string
+	storage := d.Get("storage_mb").Int()
+
+	var tier, size, version string
 
 	s := strings.Split(sku, "_")
 	if len(s) < 3 || len(s) > 4 {
@@ -33,7 +30,7 @@ func NewAzureRMPostrgreSQLFlexibleServer(d *schema.ResourceData, u *schema.Usage
 	}
 
 	if len(s) > 2 {
-		tier = s[0]
+		tier = strings.ToLower(s[0])
 		size = s[2]
 	}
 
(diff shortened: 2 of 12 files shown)


internal/resources/azure/mysql_flexible_server.go
```diff
@@ -0,0 +1,161 @@
...
+				{Key: "meterName", Value: strPtr("Storage Data Stored")},
+			},
+		},
+	}
+}
+
+// iopsCostComponent returns a cost component for additional IOPS.
+func (r *MySQLFlexibleServer) iopsCostComponent() *schema.CostComponent {
+	var quantity *decimal.Decimal
+	if r.IOPS > 0 {
```
vdmgolub, line 116: Not sure how to calculate the IOPS value:

1. Terraform has `storage.iops` https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/mysql_flexible_server#iops. Similar to it `storage_gb` represents the provisioned storage.
2. Pricing page mentions that each server includes free IOPS https://azure.microsoft.com/en-gb/pricing/details/mysql/flexible-server/:
   > the included IOPS of 3 IOPS/GB per total storage provisioned and free 300 IOPS available per server

   So, for example, the provisioned storage is 10GB, then the free IOPS would be 3*10 + 300 = 330. If the `storage.iops = 360` in TF resource, does it mean that the "additional IOPS" are 360 - 330 = 30 and we should show cost for these 30? NB: This formula doesn't apply for "Basic" burstable instances.
  There is another, but similar formula from [here](https://docs.microsoft.com/en-us/azure/mysql/flexible-server/concepts-compute-storage#iops) for "Complimentary IOPS": 
   > Complimentary IOPS are equal to MINIMUM("Max uncached disk throughput: IOPS/MBps" of compute size, 300 + storage provisioned in GiB * 3)

   Not sure if we can get this throughput value.

Or should we ignore the `storage.iops` value entirely and introduce a usage param`monthly_additional_iops` instead?

internal/resources/azure/mysql_flexible_server.go
```diff
@@ -0,0 +1,161 @@
...
+			},
+		},
+	}
+}
+
+// backupCostComponent returns a cost component for additional backup storage.
+func (r *MySQLFlexibleServer) backupCostComponent() *schema.CostComponent {
+	var quantity *decimal.Decimal
+	if r.AdditionalBackupStorageGB != nil {
+		quantity = decimalPtr(decimal.NewFromFloat(*r.AdditionalBackupStorageGB))
```
vdmgolub, line 142: According to pricing page if geo-redundancy is enabled, the storage is doubled:

> GRS requires twice the amount of storage capacity because it creates a replica. If you have configured the back up storage to be geo-redundant (GRS), the price will be 2x this price. Price per GB remains the same.

Should the usage parameter include that? Or should we rely on [geo_redundant_backup_enabled](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/mysql_flexible_server#geo_redundant_backup_enabled) flag and double the storage (and cost) here if it is set to true?

If yes, we'll need to fix it for PostgreSQL Flexible Server as well.

internal/providers/terraform/azure/util.go
```diff
@@ -114,3 +114,12 @@ func locationNameMapping(l string) string {
 func intPtr(i int64) *int64 {
 	return &i
 }
+
+func contains(arr []string, e string) bool {
```
vdmgolub, line 118: FYI, I tried to use `slices.Contains` from Go 1.18, but the golint failed in CI. Rolled back to this function for now. I guess we need to wait for this fix https://github.com/golangci/golangci-lint/issues/2649.

internal/providers/terraform/azure/postgresql_flexible_server.go
```diff
@@ -33,110 +30,39 @@ func NewAzureRMPostrgreSQLFlexibleServer(d *schema.ResourceData, u *schema.Usage
...
 	}
 
-	tierName := map[string]string{
-		"B":  "Burstable",
-		"GP": "General Purpose",
-		"MO": "Memory Optimized",
-	}[strings.ToUpper(tier)]
-
-	if tierName == "" {
-		log.Warnf("Unrecognised PostgreSQL tier prefix for resource %s: %s", d.Address, tierName)
```
vdmgolub, line 51: Decided to use the full `sku` value in the error messages as it reads better than separate substrings, the text already scopes them.

internal/resources/azure/mysql_flexible_server.go
aliscott: Good question. I can see benefits of both.

1. Assuming the usage parameter includes it would probably map better to any cost and usage reports.
2. Doing the doubling ourself might be simpler to the user for modelling.

I'm leaning towards 1 just now because it's less ambiguous, but we'd need to add a comment to the usage file to make it clear that for GRS you should double it yourself. What do you think?

vdmgolub: Agree, ambiguity is the main problem here. I'd assume users know what they are doing when they set the geo-redundancy flag, so something like maybe enough for now:
```yaml
  azurerm_mysql_flexible_server.my_flexible_server:
    additional_backup_storage_gb: 5000 # Additional consumption of backup storage in GB. This includes the replica storage when geo-redundancy is enabled.
```

Is it clear enough?

internal/resources/azure/mysql_flexible_server.go
aliscott: > So, for example, the provisioned storage is 10GB

Provisioned storage seems to have a minimum of 20GB for all instance types when I try to launch in the Azure portal. So IOPS minimum is 360, which is free. Apart from that I think this formula makes sense.

> This formula doesn't apply for "Basic" burstable instances.

Yeah, in the Azure portal I'm seeing burstable instance types have a max IOPS listed. I don't know if we need to worry about representing this. We can test if the Terraform plan fails if it goes above that max IOPS. If so, then I think it's okay for us just to use the same formula for these.

internal/resources/azure/mysql_flexible_server.go
aliscott: Maybe this:

> Additional backup storage in GB. If geo-redundancy is enabled, you should set this to twice the required storage capacity.

vdmgolub: Yep, that's better 👍 . Thanks!

internal/resources/azure/mysql_flexible_server.go
vdmgolub: Updated.

internal/resources/azure/mysql_flexible_server.go
vdmgolub: Updated, also for PostgreSQL Flexible Server.

internal/providers/terraform/azure/mysql_flexible_server.go
```diff
@@ -0,0 +1,70 @@
...
+		Name:  "azurerm_mysql_flexible_server",
+		RFunc: newMySQLFlexibleServer,
+	}
+}
+
+func newMySQLFlexibleServer(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := d.Get("location").String()
+	sku := d.Get("sku_name").String()
+	storage := d.GetInt64OrDefault("storage.0.size_gb", 20)
+	iops := d.GetInt64OrDefault("storage.0.iops", 360)
```
vdmgolub, line 23: If values are not defined (they are optional), we default them the minimums.

internal/resources/azure/mysql_flexible_server.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+// storageCostComponent returns a cost component for server's storage. If
+// storage is not defined, it is assumed it is a minimum default of 20GB.
+func (r *MySQLFlexibleServer) storageCostComponent() *schema.CostComponent {
+	storage := r.Storage
+	if storage == 0 {
+		storage = 20 // minimum default
+	}
```
vdmgolub, line 98: The storage value is optional. Defaulting it to 20GB here instead of IaC resource.

internal/resources/azure/postgresql_flexible_server.go
```diff
@@ -0,0 +1,175 @@
...
+	Tier            string
+	InstanceType    string
+	InstanceVersion string
+	Storage         int64
+
+	AdditionalBackupStorageGB *float64 `infracost_usage:"additional_backup_storage_gb"`
+}
+
+// PostgresqlFlexibleServerUsageSchema defines a list which represents the usage schema of PostgresqlFlexibleServer.
+var PostgresqlFlexibleServerUsageSchema = []*schema.UsageItem{
```
aliscott, line 30:
```suggestion
var PostgreSQLFlexibleServerUsageSchema = []*schema.UsageItem{
```

aliscott (approved):
LGTM! Thanks for fixing up PostgreSQL Flexible Server as well!

PR #1542, opened 2022-04-06
fix(hcl): properly evaluate count expressions

hugorut:
Fixes issues where `count` traversal expressions were not being properly evaluated. This was for two reasons:

1. When a block was cloned due to a count the context value was still an object. e.g:

```go
map[string]interface{
    "aws_launch_configuration": map[string]interface{
           "test_count": map[string]interface{
               "id": ...
            },
            "test_count[1]": map[string]interface{
               "id": ...
            }
    }
}
```

It should instead look like:

```go
map[string]interface{
    "aws_launch_configuration": map[string]interface{
           "test_count": []map[string]interface{
                {
                   "id": ...
                },
                {
                   "id": ...
                }
            }
        }
    }
}
```

This meant that only the first resource in the `count` was being pulled as a reference. All references that used a count > 0 were returning null as there were no items in the context value list past 1 (under the hood the hcl libraries were turning the object into a list of length 1).
2. The cloned block still had the id from the original block. So when we were evaluating the references in Infracost, the `idMap` was sometimes referencing incorrect resources.

---

I've solved the above by:

* adding greater support for expression evaluation in HCL
* when blocks are expanded as part of a count setting the context to be a Tuple type
* adding a dynamic `count.index` expression to the uuid we generate on the block

### DIFF
--- a/internal/hcl/attribute.go
+++ b/internal/hcl/attribute.go
@@ -162,31 +162,12 @@ func (attr *Attribute) Reference() (*Reference, error) {
 	if attr == nil {
 		return nil, fmt.Errorf("attribute is nil")
 	}
-	switch t := attr.HCLAttr.Expr.(type) {
-	case *hclsyntax.RelativeTraversalExpr:
-		if s, ok := t.Source.(*hclsyntax.IndexExpr); ok {
-			collectionRef, err := createDotReferenceFromTraversal(s.Collection.Variables()...)
-			if err != nil {
-				return nil, err
-			}
-			key, _ := s.Key.Value(attr.Ctx.Inner())
-			collectionRef.SetKey(key)
-
-			return collectionRef, nil
-		}
-
-		return createDotReferenceFromTraversal(t.Source.Variables()...)
-	case *hclsyntax.ScopeTraversalExpr:
-		return createDotReferenceFromTraversal(t.Traversal)
-	case *hclsyntax.TemplateExpr:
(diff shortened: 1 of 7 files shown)


aliscott (approved):
LGTM. I'm wondering if we need similar for name-indexed resources? Here's a case that doesn't work for me for both the plan JSON and HCL parsing, but there's probably others as well that work for plan JSON but not HCL parsing.

```hcl
provider "aws" {
  region                      = "us-east-1" # <<<<< Try changing this to eu-west-1 to compare the costs
  skip_credentials_validation = true
  skip_requesting_account_id  = true
  access_key                  = "mock_access_key"
  secret_key                  = "mock_secret_key"
}

locals {
  instance_types = {
    "micro": "t2.micro",
    "medium": "t2.medium"
  }
}

resource "aws_autoscaling_group" "test_named" {
  for_each = toset(["micro", "medium"])
  desired_capacity     = 2
  max_size             = 3
  min_size             = 1
  launch_configuration = aws_launch_configuration.test_named[each.key].id
}

resource "aws_launch_configuration" "test_named" {
  for_each = toset(["micro", "medium"])
  image_id      = "fake_ami"
  instance_type = local.instance_types[each.key]
}
```

hugorut:
@aliscott nice catch, let's investigate that a little further and open a new pr for it. I'd like to keep these hcl bug fixing prs small so they're easier to understand.

PR #1543, opened 2022-04-06
fix(hcl): module outputs evaluated before count expansion

hugorut:
fixes issue where module outputs used in count evaluation were not available when expanding counts. This meant that the resources that were marked as disabled were being falsely generated. e.g:

```
module "this" {
  source  = "./modules/test"
  enabled = false
}

resource "aws_eip" "t" {
  count = module.this.enabled ? 1 : 0

  tags = {
    "test" : module.this.enabled
  }
}
```

in this case `module.this.enabled` would not be in the evaluation context meaning that we would use a fallback and expand the block to a count of 1.

---

Depends on https://github.com/infracost/infracost/pull/1542

### DIFF
--- a/internal/hcl/evaluator.go
+++ b/internal/hcl/evaluator.go
@@ -107,6 +107,7 @@ func (e *Evaluator) Run() ([]*Module, error) {
 
 	// let's load the modules now we have our top level context.
 	e.moduleCalls = e.loadModules()
+	e.evaluate(lastContext)
 
 	// expand out resources and modules via count and evaluate again so that we can include
 	// any module outputs and or count references.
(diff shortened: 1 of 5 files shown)


internal/hcl/evaluator.go
```diff
@@ -107,6 +107,7 @@ func (e *Evaluator) Run() ([]*Module, error) {
 
 	// let's load the modules now we have our top level context.
 	e.moduleCalls = e.loadModules()
+	e.evaluate(lastContext)
```
hugorut, line 110: evaluate the context again before expanding the blocks. So that module outputs are available for expressions.

PR #1548, opened 2022-04-07
fix(hcl): count references not evaluating

hugorut:
This fixes issue where resources built with HCL were not able to fetch references for a non `id` or `arn` attribute if the resource had a count. For example a resource like:

```
resource "aws_ecs_service" "test" {
  count                              = 2
  name                               = "test"
  task_definition                = "${join("", aws_ecs_task_definition.default.*.family)}:${join("", aws_ecs_task_definition.default.*.revision)}"
  ...
}
```

would be unable to resolve the `aws_ecs_task_definition`. This was because the Plan JSON generated by the `hcl_provider` had some discrepancies. Especially in the `configuration` section of the plan JSON. 

1. Resources were being added to the `resources` list in the configuration block and not being deduped for their index:

```
    [
        {
          "address": "aws_eip.test[0]",
          "mode": "managed",
          "type": "aws_eip",
          "name": "test[0]",
          "provider_config_key": "aws",
          "schema_version": 0,
          "count_expression": {
            "constant_value": 2
          }
        },
        {
          "address": "aws_eip.test[1]",
          "mode": "managed",
          "type": "aws_eip",
          "name": "test[1]",
          "provider_config_key": "aws",
          "schema_version": 0,
          "count_expression": {
            "constant_value": 2
          }
        }
   ]
  ```  
  
  should instead be:
  
  ```
  [        
        {
          "address": "aws_eip.test",
          "mode": "managed",
          "type": "aws_eip",
          "name": "test",
          "provider_config_key": "aws",
          "schema_version": 0,
          "count_expression": {
            "constant_value": 2
          }
        }
 ]
```

2. Expression which used a count would not contain the count in the references array:

```
  "expressions": {
    "launch_configuration": {
      "references": [
        "aws_launch_configuration.test[0]"
      ]
    }
  },
 ```
 
 should be:
 
```
  "expressions": {
    "launch_configuration": {
      "references": [
        "aws_launch_configuration.test",
        "count.index"
      ]
    }
  },
```
3. recursive modules were not nesting in the planned resources or configuration sections. So they were all being assigned to the top-level module. This wasn't an issue per-say, however it was hard to compare Terraform JSON and HCL JSON to look for discrepancies 
---

In order to fix the above I've:

1. added function expression evaluation, as this was not properly covered when evaluating references. Meaning expressions like `coalesce(var.task_definition, "${join("", aws_ecs_task_definition.default.*.family)}:${join("", aws_ecs_task_definition.default.*.revision)}")` would not evaluate to a reference.
2. Refactored the `Evaluator` to return a single `Module` with child `Modules` to more accurately represent the HCL config tree. This makes it easier to generate the plan JSON structure in the `HCLProvider`
3. Refactored the `HCLProvider` to generate the proper plan JSON structure.
4. Added tests to cover the `HCLProvider` to make sure that we are generating the correct output. This meant refactoring the `Block` generation process a little into a `BlockBuilder` struct that we can set a dynamic list of `SetAttributesFunc` on. This enables us to mock the uuid generation to keep it static for the test files. It also enables us in the future to build a list of `SetAttributesFunc` with each being specific to certain resources. E.g. `aws_ecs_task_definition` has an output attribute of `revision` which is generated on apply. We could generate this with a `SetAttributesFunc` which is scoped to just that resource.

### DIFF
--- a/internal/hcl/block.go
+++ b/internal/hcl/block.go
@@ -3,6 +3,8 @@ package hcl
 import (
 	"fmt"
 	"os"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/google/uuid"
@@ -111,8 +113,13 @@ type Block struct {
 	verbose bool
 }
 
-// NewHCLBlock returns a Block with Context and child Blocks initialised.
-func NewHCLBlock(hclBlock *hcl.Block, ctx *Context, moduleBlock *Block) *Block {
+// BlockBuilder handles generating new Blocks as part of the parsing and evaluation process.
+type BlockBuilder struct {
+	SetAttributes []SetAttributesFunc
+}
+
+// NewBlock returns a Block with Context and child Blocks initialised.
+func (b BlockBuilder) NewBlock(hclBlock *hcl.Block, ctx *Context, moduleBlock *Block) *Block {
 	if ctx == nil {
 		ctx = NewContext(&hcl.EvalContext{}, nil)
 	}
@@ -120,21 +127,12 @@ func NewHCLBlock(hclBlock *hcl.Block, ctx *Context, moduleBlock *Block) *Block {
 	isLoggingVerbose := strings.TrimSpace(os.Getenv("INFRACOST_HCL_DEBUG_VERBOSE")) == "true"
 	var children Blocks
 	if body, ok := hclBlock.Body.(*hclsyntax.Body); ok {
-		for _, b := range body.Blocks {
-			children = append(children, NewHCLBlock(b.AsHCLBlock(), ctx, moduleBlock))
+		for _, bb := range body.Blocks {
+			children = append(children, b.NewBlock(bb.AsHCLBlock(), ctx, moduleBlock))
 		}
 
-		if hclBlock.Type == "resource" || hclBlock.Type == "data" {
-			_, withCount := body.Attributes["count"]
-			// add commonly used identifiers to the block so that if it's referenced by other
-			// blocks in context evaluation.
-			if _, ok := body.Attributes["id"]; !ok {
-				body.Attributes["id"] = newUniqueAttribute("id", withCount)
-			}
-
-			if _, ok := body.Attributes["arn"]; !ok {
-				body.Attributes["arn"] = newArnAttribute("arn", withCount)
-			}
+		for _, f := range b.SetAttributes {
+			f(moduleBlock, hclBlock)
 		}
 
 		return &Block{
@@ -162,7 +160,7 @@ func NewHCLBlock(hclBlock *hcl.Block, ctx *Context, moduleBlock *Block) *Block {
 	}
 
 	for _, hb := range content.Blocks {
-		children = append(children, NewHCLBlock(hb, ctx, moduleBlock))
+		children = append(children, b.NewBlock(hb, ctx, moduleBlock))
 	}
 
 	return &Block{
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Comment by hugorut** [comment]
Failing azure tests seem to be with an API change, not due to these changes.

File: internal/hcl/evaluator.go [hunk 1]
```diff
@@ -111,22 +107,19 @@ func (e *Evaluator) Run() ([]*Module, error) {
...
 }
 
-func (e *Evaluator) collectModules() []*Module {
-	rootModule := &Module{Blocks: e.blocks, RootPath: e.projectRootPath, ModulePath: e.modulePath}
-	modules := []*Module{rootModule}
-
+func (e *Evaluator) collectModules() *Module {
 	for _, definition := range e.moduleCalls {
-		modules = append(modules, definition.Modules...)
+		e.module.Modules = append(e.module.Modules, definition.Module)
```
vdmgolub, line 119: It seems currently `e.module.Modules` can be nil, if we don't always default it to an empty array, we probably should check it here. What do you think is the better approach?

internal/hcl/evaluator.go
```diff
@@ -184,17 +177,24 @@ func (e *Evaluator) evaluateModules() {
 
 		vars := module.Definition.Values().AsValueMap()
 		moduleEvaluator := NewEvaluator(
-			e.projectRootPath,
-			module.Path,
+			Module{
+				Name:       module.Definition.FullName(),
```
vdmgolub, line 181: Maybe better to rename this variable to `moduleCall` as `module.Module` looks strange? Or does it look okay for you?

internal/providers/terraform/hcl_provider.go
```diff
@@ -390,22 +494,35 @@ type ResourceData struct {
...
 
 type ModuleCall struct {
-	Source string           `json:"source"`
-	Module ModuleCallModule `json:"module"`
-}
-
-type ModuleCallModule struct {
-	Resources []ResourceData `json:"resources"`
+	Source string       `json:"source"`
+	Module ModuleConfig `json:"module"`
```
vdmgolub, line 504: Should we call it `ModuleConfig`?
```suggestion
	ModuleConfig ModuleConfig `json:"module"`
```

There are too many "module" words thrown around. :) I guess types help in the editor, but it requires extra attention when reading in GH diff. How do you feel about it?

internal/providers/terraform/testdata/hcl_provider_test/renders_multiple_count_resources_correctly/expected.json
```diff
@@ -0,0 +1,371 @@
+{
+  "format_version": "1.0",
+  "terraform_version": "1.1.0",
+  "planned_values": {
+    "root_module": {
+      "resources": [
```
vdmgolub, line 6: Do we have test coverage for resources without counts? It seems like all new tests use counts, but should we check how other resources behave with new logic?

internal/hcl/block.go
```diff
@@ -545,6 +599,10 @@ func (b *Block) LocalName() string {
 // Would have its FullName as module.web_app.aws_instance.t3_standard
 // FullName is what Terraform uses in its JSON output file.
 func (b *Block) FullName() string {
+	if b == nil {
```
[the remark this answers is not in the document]
hugorut: no it wouldn't, you can call methods on a nil interface, example here: https://go.dev/play/p/sg2pjOxOQdW

internal/hcl/evaluator.go
```diff
@@ -76,31 +73,30 @@ func NewEvaluator(
...
+		blockBuilder:   blockBuilder,
 	}
 }
 
 // Run builds the Evaluator Context using all the provided Blocks. It will build up the Context to hold
 // variable and reference information so that this can be used by Attribute evaluation. Run will also
 // parse and build up and child modules that are referenced in the Blocks and runs child Evaluator on
-// these Modules.
-func (e *Evaluator) Run() ([]*Module, error) {
+// these Module.
```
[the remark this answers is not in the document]
hugorut: I don't think I actually meant to alter this file, it seems my IDE auto altered comments. I'll revert

internal/hcl/evaluator.go
```diff
@@ -184,17 +177,24 @@ func (e *Evaluator) evaluateModules() {
...
 		moduleEvaluator := NewEvaluator(
-			e.projectRootPath,
-			module.Path,
+			Module{
+				Name:       module.Definition.FullName(),
+				Source:     module.Module.Source,
+				Blocks:     module.Module.Blocks,
+				RootPath:   e.module.RootPath,
+				ModulePath: module.Path,
+				Modules:    nil,
```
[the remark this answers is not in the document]
hugorut: you won't have to have any additional checks, see: https://go.dev/play/p/Lb3sRQfE_C2 as an example

internal/hcl/modules/registry.go
```diff
@@ -46,7 +46,7 @@ func (r *RegistryLoader) lookupModule(moduleAddr string, versionConstraints stri
 		return nil, err
 	}
 
-	// Modules are in the format (registry)/namspace/module/target
+	// Module are in the format (registry)/namspace/module/target
```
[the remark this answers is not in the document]
hugorut: see prior comment about unintentional reformatting by IDE

internal/hcl/modules/registry.go
```diff
@@ -190,7 +190,7 @@ func findLatestMatchingVersion(versions []string, constraints string) (string, e
 // This does not mean that the module address is a registry module, it could still be a remote module.
 // To work that out we need to try looking up the module using the `lookupModule` function
 func normalizeRegistrySource(moduleAddr string) (string, error) {
-	// Modules are in the format (registry)/namspace/module/target
+	// Module are in the format (registry)/namspace/module/target
```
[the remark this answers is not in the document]
hugorut: see prior comment about unintentional reformatting by IDE

internal/providers/terraform/hcl_provider.go
```diff
@@ -77,182 +81,259 @@ func NewHCLProvider(ctx *config.ProjectContext, provider *PlanJSONProvider) (HCL
...
+			ProviderConfigKey: block.ModuleName() + ":" + block.Provider(),
+			Expressions:       blockToReferences(block),
+			CountExpression:   countReferences(block),
+		}
+	} else {
+		configuration = ResourceData{
+			Address:           stripCount(block.FullName()),
+			Mode:              "managed",
+			Type:              block.TypeLabel(),
+			Name:              stripCount(stripCount(block.NameLabel())),
```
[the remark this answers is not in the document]
hugorut: no we don't - typo!

internal/hcl/block.go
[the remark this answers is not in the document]
vdmgolub: TIL, but confused 😅. Thanks for the example.

internal/hcl/evaluator.go
hugorut: this should be fine, a nil slice is not the same as a nil struct see: https://go.dev/play/p/Rq1T8rdoV3j

internal/hcl/evaluator.go
hugorut: yup agreed

internal/providers/terraform/hcl_provider.go
hugorut: Yeah sounds good

internal/providers/terraform/testdata/hcl_provider_test/renders_multiple_count_resources_correctly/expected.json
hugorut: yup, I'll add test coverage here, but also remember all the other resource Golden file tests cover the HCL functionality. So don't fret too much that it's not covered by this specific JSON file case.

vdmgolub (approved):
👍 🚀

PR #1551, opened 2022-04-08
feat: Add support for SNS Topic notification costs

### LINKED ISSUE #1393: Add sms support for `aws_sns_topic_subscription`
[](https://aws.amazon.com/sns/sms-pricing/)Currently infracost only supports `http|https` message delivery for `aws_sns_topic_subscription`. Lots of infracost users use sns to deliver sms messages to their end users. Infracost should figure out a way to support this use case.

## Pricing

AWS SMS pricing information can be found here: https://aws.amazon.com/sns/sms-pricing/ .  For the `aws_sns_topic_subscription` use case, Infracost would only be interested in outbound costs.  

For outbound sms prices varies by destination country. At face value this seems easy to do, however there are a few things that make implementing SMS costs problematic.

### Per country billing variance

The price and components for sending SMS messages varies between countries. For example see below:

**US**
  
Screenshot 2022-02-23 at 11 53 54

**Spain**

Screenshot 2022-02-23 at 12 00 40

**India**

Screenshot 2022-02-23 at 11 54 01

AWS also states that pricing can also vary between carrier in country (although from our investigation none of the inter country carriers have a different price). 

The lack of consistency in pricing means that offering Infracost users all encompassing configuration options for sms pricing could be quite complex for a user to configure. Below I outline a few examples.

### Complex usage parameters

We could modify the `monthly_requests` usage param to be a more complex array type which would allow users to specify sms destination info. Take the following examples.

```bash
aws_sns_topic.my_sns_topic:
  monthly_requests: 
    - destination: "us" # The destination of the sms requests, must be a valid region/country.
      sent: 1000000 # Monthly requests which are sent to US based numbers.
      sms_identity: "10dlc" # SMS Origination Identities, valid options are 10dlc, longcode, shortcode, tollfree.
      sms_type: "transactional" # The type of sms sent, can be one of transactional or promotional
      # carrier: "some carrier" can be used to specify the destination carrier	
```

This approach would need us to modify the pricing api to incorporate custom scrapers for outbound sms pricing, found here: [https://s3.amazonaws.com/aws-messaging-pricing-information/TextMessageOutbound/prices.json](https://s3.amazonaws.com/aws-messaging-pricing-information/TextMessageOutbound/prices.json) . 

**Pros**

- The most flexible way to specify outbound sms costs
- Automatically adjusts to aws pricing changes

**Cons**

- Requires complex validation for destination rules
- Difficult for users to understand

### Static usage cost

Another approach would be to let users to specify a static cost for the `monthly_request` sns usage param. This would enable users to provide their own estimate for outbound sms costs. Usage could look like such:

```bash
aws_sns_topic.my_sns_topic:
    monthly_requests: 1000
    sms_req_cost: 0.0058
```

**Pros**

- Simple to implement
- Easy to understand

**Cons**

- Requires users to keep up to date with monthly pricing
- Does not change per country/region so users need to provide an average - therefore a worse estimate of costs incurred

---

A third approach could lie somewhere between both approaches. This could mean offering more complex usage options but removing option to specify additional usage params like `sms_type` or `carrier`. In these cases Infracost could assume a set scenario, however we would need to communicate this with the users. For example 

*infracost only supports transactional sms sending costs* 

or 

*infracost  does not support inter carrier sending, we assume a single cost for a country*

tim775:
Fixed #1393

- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/pull/200)

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -431,7 +431,13 @@ resource_usage:
 
   aws_sns_topic.my_sns_topic:
     monthly_requests: 1000000 # Monthly requests to SNS.
-    request_size_kb: 64       # Size of requests to SNS, billed in 64KB chunks. So 1M requests at 128KB uses 2M requests.
+    request_size_kb: 64 # Size of requests to SNS, billed in 64KB chunks. So 1M requests at 128KB uses 2M requests.
+    http_subscriptions: 100 # Number of HTTP/HTTPS subscriptions
+    email_subscriptions: 100 # Number of Email/Email-JSON subscriptions
+    kinesis_subscriptions: 0 # Number of Kinesis Firehose subscriptions
+    mobile_push_subscriptions: 0 # Number of Mobile Push subscriptions
+    macos_subscriptions: 0 # Number of MacOS subscriptions
+    sms_subscriptions: 1 # Number of SMS subscriptions
 
   aws_sns_topic_subscription.my_topic_subscription:
     monthly_requests: 1000000 # Monthly requests to SNS.
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -441,8 +441,8 @@ resource_usage:
     sms_notification_price: 0.008 # Average price for each SMS notification
 
   aws_sns_topic_subscription.my_topic_subscription:
-    monthly_requests: 1000000 # Monthly requests to SNS.
-    request_size_kb: 64       # Size of requests to SNS, billed in 64KB chunks. So 1M requests at 128KB uses 2M requests.
+    monthly_requests: 1000000 # (DEPRECATED use aws_sns_topic.monthly_requests instead) Monthly requests to SNS.
+    request_size_kb: 64       # (DEPRECATED use aws_sns_topic.request_size_kb instead) Size of requests to SNS, billed in 64KB chunks. So 1M requests at 128KB uses 2M requests.
 
   aws_sqs_queue.my_queue:
     monthly_requests: 1000000 # Monthly requests to SQS.
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -438,6 +438,7 @@ resource_usage:
     mobile_push_subscriptions: 0 # Number of Mobile Push subscriptions
     macos_subscriptions: 0 # Number of MacOS subscriptions
     sms_subscriptions: 1 # Number of SMS subscriptions
+    sms_notification_price: 0.008 # Average price for each SMS notification
 
   aws_sns_topic_subscription.my_topic_subscription:
     monthly_requests: 1000000 # Monthly requests to SNS.
--- a/internal/resources/aws/sns_topic.go
+++ b/internal/resources/aws/sns_topic.go
@@ -40,17 +40,17 @@ var SNSTopicUsageSchema = []*schema.UsageItem{
 // without having any IaAC)
 func (r *SNSTopic) CostComponents() []*schema.CostComponent {
 	return []*schema.CostComponent{
-		r.APIRequestsCostComponent(nil),
-		r.HTTPNotificationsCostComponent(nil, nil),
-		r.EmailNotificationsCostComponent(nil, nil),
-		r.KinesisNotificationsCostComponent(nil, nil),
-		r.MobilePushNotificationsCostComponent(nil, nil),
-		r.MacOSNotificationsCostComponent(nil, nil),
-		r.SMSNotificationsCostComponent(nil, nil, nil),
+		r.apiRequestsCostComponent(nil),
+		r.httpNotificationsCostComponent(nil, nil),
+		r.emailNotificationsCostComponent(nil, nil),
+		r.kinesisNotificationsCostComponent(nil, nil),
+		r.mobilePushNotificationsCostComponent(nil, nil),
+		r.macOSNotificationsCostComponent(nil, nil),
+		r.smsNotificationsCostComponent(nil, nil, nil),
 	}
 }
 
-func (r *SNSTopic) APIRequestsCostComponent(requests *int64) *schema.CostComponent {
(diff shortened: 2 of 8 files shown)


infracost-usage-example.yml
```diff
@@ -431,11 +431,18 @@ resource_usage:
...
     monthly_requests: 1000000 # Monthly requests to SNS.
-    request_size_kb: 64       # Size of requests to SNS, billed in 64KB chunks. So 1M requests at 128KB uses 2M requests.
+    request_size_kb: 64 # Size of requests to SNS, billed in 64KB chunks. So 1M requests at 128KB uses 2M requests.
+    http_subscriptions: 100 # Number of HTTP/HTTPS subscriptions
+    email_subscriptions: 100 # Number of Email/Email-JSON subscriptions
+    kinesis_subscriptions: 0 # Number of Kinesis Firehose subscriptions
+    mobile_push_subscriptions: 0 # Number of Mobile Push subscriptions
+    macos_subscriptions: 0 # Number of MacOS subscriptions
+    sms_subscriptions: 1 # Number of SMS subscriptions
+    sms_notification_price: 0.008 # Average price for each SMS notification
```
hugorut, line 441: No biggie, but I believe the comments need to be lined up with longest line right? so:

```
    sms_subscriptions: 1          # Number of SMS subscriptions
    sms_notification_price: 0.008 # Average price for each SMS notification
```

internal/resources/aws/sns_topic.go
```diff
@@ -1,62 +1,230 @@
...
+			Service:       strPtr("AmazonSNS"),
+			ProductFamily: strPtr("API Request"),
+		},
+		PriceFilter: &schema.PriceFilter{
+			StartUsageAmount: strPtr("1000000"),
+		},
+	}
+}
+
+func (r *SNSTopic) HTTPNotificationsCostComponent(subscriptions, requests *int64) *schema.CostComponent {
```
hugorut, line 79: are these notification cost component methods exported for a reason? If not let's alter them to be un-exported.

internal/resources/aws/sns_topic.go
```diff
@@ -1,62 +1,230 @@
...
+	{Key: "mobile_push_subscriptions", ValueType: schema.Int64, DefaultValue: 0},
+	{Key: "macos_subscriptions", ValueType: schema.Int64, DefaultValue: 0},
+	{Key: "sms_subscriptions", ValueType: schema.Int64, DefaultValue: 0},
+	{Key: "sms_notification_price", ValueType: schema.Float64, DefaultValue: 0.0075},
+}
+
+// This is an experiment to see if using an explicit structure to define the cost components
+// can enable anything interesting (e.g. list what cost components could apply to a resource
+// without having any IaAC)
+func (r *SNSTopic) CostComponents() []*schema.CostComponent {
```
hugorut, line 41: I know this is marked as an experiment, but are we using this anywhere right now? It doesn't look like it, so I'd just remove it, otherwise it's just dead code.

internal/schema/cost_component.go
```diff
@@ -55,6 +56,14 @@ func (c *CostComponent) PriceHash() string {
...
 }
 
+func (c *CostComponent) SetCustomPrice(price *decimal.Decimal) {
+	c.customPrice = price
+}
+
+func (c *CostComponent) CustomPrice() *decimal.Decimal {
+	return c.customPrice
+}
+
```
hugorut, line 66: Personally, I don't see the point of these setters and getters. I would prefer just an exported `CustomPrice` field.

internal/resources/aws/sns_topic.go
```diff
@@ -1,62 +1,230 @@
...
+	}
+
+	components := []*schema.CostComponent{
+		r.APIRequestsCostComponent(requests),
+		r.HTTPNotificationsCostComponent(r.HTTPSubscriptions, requests),
+		r.EmailNotificationsCostComponent(r.EmailSubscriptions, requests),
+		r.KinesisNotificationsCostComponent(r.KinesisSubscriptions, requests),
+		r.MobilePushNotificationsCostComponent(r.MobilePushSubscriptions, requests),
+		r.MacOSNotificationsCostComponent(r.MacOSSubscriptions, requests),
+		r.SMSNotificationsCostComponent(r.SMSSubscriptions, requests, r.SMSNotificationPrice),
```
hugorut, line 180: I'm not sure why we are passing in struct properties into struct methods. Could we not just do:

```suggestion
		r.HTTPNotificationsCostComponent(requests),
		r.EmailNotificationsCostComponent(requests),
		r.KinesisNotificationsCostComponent(requests),
		r.MobilePushNotificationsCostComponent(requests),
		r.MacOSNotificationsCostComponent(requests),
		r.SMSNotificationsCostComponent(requests),
```

and then the struct methods would fetch the relevant property, e.g:

```go

func (r *SNSTopic) EmailNotificationsCostComponent(requests *int64) *schema.CostComponent {
	return r.notificationsCostComponent(
		"Email/Email-JSON notifications (over 1k)",
		"100k notifications",
		100000,
		"DeliveryAttempts-SMTP",
		1000,
		r.EmailSubscriptions,
		requests,
	)
}
```

hugorut (commented):
A few tweaks, but looks good

internal/resources/aws/sns_topic.go
tim775: Fair, I'll comment it out so it's clear.

internal/schema/cost_component.go
tim775: I sort of agree they are pointless and it seems like a personal aesthetic preference either way, but I think it is worthwhile to be consistent with Price/PriceHash defined above.  If it's important to clean those up I guess we could have a tech debt task.

internal/resources/aws/sns_topic.go
tim775: I considered that, but I prefer the clarity of passing them in explicitly.

hugorut (approved):

PR #1552, opened 2022-04-08
feat: Add SNS FIFO Topic cost estimates.

### LINKED ISSUE #1393: Add sms support for `aws_sns_topic_subscription`

tim775:
Fixes #1393

This is branched from `feat/add_sns_topic_subscription_costs` so if that hasn't been merged yet review the diff to eliminate the noise: https://github.com/infracost/infracost/compare/feat/add_sns_topic_subscription_costs...feat/add_sns_topic_fifo_costs

- [ ] Compared test case output to cloud cost calculator
- [ ] Created a PR to update "Supported Resources" in the [docs](<LINK TO DOCS PULL REQUEST>)


internal/providers/terraform/aws/sns_topic.go
```diff
@@ -7,17 +7,28 @@ import (
...
+			Address:       d.Address,
+			Region:        d.Get("region").String(),
+			Subscriptions: int64(len(d.References("aws_sns_topic_subscription.topic_arn"))),
+		}
 
-	r.PopulateUsage(u)
-	return r.BuildResource()
+		r.PopulateUsage(u)
+		return r.BuildResource()
+	} else {
```
hugorut, line 26: no need for an else here as we return early on line `25`

internal/resources/aws/sns_topic.go
```diff
@@ -222,3 +222,158 @@ func (r *SNSTopic) notificationsCostComponent(name, unit string, multiplier int6
 		},
 	}
 }
+
+type SNSFIFOTopic struct {
```
hugorut, line 226: Doc blocks please

internal/resources/aws/sns_topic.go
```diff
@@ -222,3 +222,158 @@ func (r *SNSTopic) notificationsCostComponent(name, unit string, multiplier int6
...
+	var requestSizeGB *float64
+	if r.RequestSizeKB != nil {
+		requestSizeGB = floatPtr(*r.RequestSizeKB / 1000000.0)
+	}
+
+	components := []*schema.CostComponent{
+		r.PublishAPIRequestsCostComponent(r.MonthlyRequests),
+		r.PublishAPIPayloadCostComponent(r.MonthlyRequests, requestSizeGB),
+		r.NotificationsCostComponent(r.Subscriptions, r.MonthlyRequests),
+		r.NotificationPayloadCostComponent(r.Subscriptions, r.MonthlyRequests, requestSizeGB),
```
hugorut, line 371: No need to pass into struct properties into a struct method. Let's remove and just use the struct properties directly in the method.

internal/resources/aws/sns_topic.go
```diff
@@ -222,3 +222,158 @@ func (r *SNSTopic) notificationsCostComponent(name, unit string, multiplier int6
...
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "usagetype", Value: strPtr("F-DA-SQS")},
+			},
+		},
+		PriceFilter: &schema.PriceFilter{
+			StartUsageAmount: strPtr("0"),
+		},
+	}
+}
+func (r *SNSFIFOTopic) NotificationPayloadCostComponent(subscriptions int64, requests *int64, requestSizeGB *float64) *schema.CostComponent {
```
hugorut, line 329:
```suggestion
func (r *SNSFIFOTopic) notificationPayloadCostComponent(subscriptions int64, requests *int64, requestSizeGB *float64) *schema.CostComponent {
```

internal/resources/aws/sns_topic.go
```diff
@@ -222,3 +222,158 @@ func (r *SNSTopic) notificationsCostComponent(name, unit string, multiplier int6
...
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "usagetype", Value: strPtr("F-Ingress-Tier1")},
+			},
+		},
+		PriceFilter: &schema.PriceFilter{
+			StartUsageAmount: strPtr("0"),
+		},
+	}
+}
+func (r *SNSFIFOTopic) NotificationsCostComponent(subscriptions int64, requests *int64) *schema.CostComponent {
```
hugorut, line 302:
```suggestion
}

func (r *SNSFIFOTopic) notificationsCostComponent(subscriptions int64, requests *int64) *schema.CostComponent {
```

internal/resources/aws/sns_topic.go
```diff
@@ -222,3 +222,158 @@ func (r *SNSTopic) notificationsCostComponent(name, unit string, multiplier int6
...
+				{Key: "usagetype", Value: strPtr("F-Request-Tier1")},
+			},
+		},
+		PriceFilter: &schema.PriceFilter{
+			StartUsageAmount: strPtr("0"),
+		},
+	}
+}
+
+func (r *SNSFIFOTopic) PublishAPIPayloadCostComponent(requests *int64, requestSizeGB *float64) *schema.CostComponent {
```
hugorut, line 277:
```suggestion
func (r *SNSFIFOTopic) publishAPIPayloadCostComponent(requests *int64, requestSizeGB *float64) *schema.CostComponent {
```

internal/resources/aws/sns_topic.go
```diff
@@ -222,3 +222,158 @@ func (r *SNSTopic) notificationsCostComponent(name, unit string, multiplier int6
...
+func (r *SNSFIFOTopic) CostComponents() []*schema.CostComponent {
+	return []*schema.CostComponent{
+		r.PublishAPIRequestsCostComponent(nil),
+		r.PublishAPIPayloadCostComponent(nil, nil),
+		r.NotificationsCostComponent(0, nil),
+		r.NotificationPayloadCostComponent(0, nil, nil),
+	}
+}
+
+func (r *SNSFIFOTopic) PublishAPIRequestsCostComponent(requests *int64) *schema.CostComponent {
```
hugorut, line 251:
```suggestion
func (r *SNSFIFOTopic) publishAPIRequestsCostComponent(requests *int64) *schema.CostComponent {
```

internal/resources/aws/sns_topic.go
```diff
@@ -222,3 +222,158 @@ func (r *SNSTopic) notificationsCostComponent(name, unit string, multiplier int6
...
+	return &schema.CostComponent{
+		Name:            "FIFO Publish API payload",
+		Unit:            "GB",
+		UnitMultiplier:  decimal.NewFromInt(1),
+		MonthlyQuantity: q,
+		ProductFilter: &schema.ProductFilter{
+			VendorName:    strPtr("aws"),
+			Region:        strPtr(r.Region),
+			Service:       strPtr("AmazonSNS"),
+			ProductFamily: strPtr(""),
```
hugorut, line 292: Is it necessary to have this product family blank for the lookup?

hugorut (commented):
minor tweaks, logic is solid

internal/resources/aws/sns_topic.go
tim775: That is what's in the db.

hugorut (approved):

PR #1553, opened 2022-04-11
chore(azure): Fix tests for CosmosDB related resource

vdmgolub:
The attribute is case-sensitive now. Fixed after provider's v3 release here:
https://github.com/hashicorp/terraform-provider-azurerm/pull/16152.


hugorut (approved):
😓

PR #1565, opened 2022-04-14
ci(fig): update autocomplete spec automatically

### LINKED ISSUE #1118: Shell completion improvements
- Add Fig.io integration: https://fig.io/docs/guides/autocomplete-for-teams
- Could `$SHELL` be used to automatically choose the shell when not specified? If not, then running `infracost completion` should show the help text as well as the existing error message. In the same way that `infracost diff` does.

fedeci:
Closes #1118

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -92,3 +92,25 @@ jobs:
       AMPLIFY_APP_ID: ${{ secrets.AMPLIFY_APP_ID }}
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+  update-fig-autocomplete-spec:
+    name: Update Fig autocomplete spec
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Get the new version
+        id: version
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yml
```diff
@@ -92,3 +92,25 @@ jobs:
...
+        id: version
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+      - name: Generate the spec
+        run: |
+          make build
+          ./build/infracost fig-autocomplete > fig-spec.ts
+      - name: Create Autocomplete PR
+        uses: withfig/publish-to-fig-autocomplete@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
aliscott, line 111: @fedeci what's the token used for? Since this is a third-party action, I think we should scope the permissions of the token to the minimum required by the action (https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/).

fedeci: It is used to: 
- create a fork of the autocomplete repo;
- create a branch with the updated spec on the forked autocomplete repo;
- create the PR against the autocomplete repo;

Repo scope is sufficient.

aliscott: Thanks 🙏. I'm going to update our default to be `contents: read` which should work with this (see https://github.com/infracost/infracost/pull/1566).

fedeci: 👍🏻

aliscott (approved):
LGTM, thanks @fedeci

fedeci:
Great! Ping me if it happens to fail!

PR #1568, opened 2022-04-15
fix(hcl): evalute modules in order of reference

hugorut:
Fixed issue where modules that used outputs from a module that appeared after them in the file structure received null values. 

For example:

```
module "sub" {
  source = "./modules/sub"
  some_var = module.this.output
}

module "this" {
  source  = "./modules/test"
  enabled = false
}
```

`module.sub` would receive a `null` value for `some_var`. This caused discrepancies in HCL cost estimates as resources that depended on these module inputs were provided with incorrect values.

This problem was caused by the fact we evaluated module blocks by order of occurrence in the HCL file.

This PR resolves this issue by sorting module blocks by reference so that we evaluate blocks whose outputs are used by other modules first.

### DIFF
--- a/cmd/infracost/cmd_test.go
+++ b/cmd/infracost/cmd_test.go
@@ -133,8 +133,8 @@ func goldenFileCommandTest(t *testing.T, testName string, args []string, testOpt
 		_, err := os.Stat(hclFilePath)
 		if err == nil {
 			testutil.AssertGoldenFile(t, hclFilePath, actual)
+			return
 		}
-		return
 	}
 
 	testutil.AssertGoldenFile(t, goldenFilePath, actual)
--- a/internal/hcl/evaluator.go
+++ b/internal/hcl/evaluator.go
@@ -521,7 +521,8 @@ func (e *Evaluator) loadModule(b *Block) (*ModuleCall, error) {
 func (e *Evaluator) loadModules() []*ModuleCall {
 	var moduleDefinitions []*ModuleCall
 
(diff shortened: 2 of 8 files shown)


cmd/infracost/cmd_test.go
```diff
@@ -133,8 +133,8 @@ func goldenFileCommandTest(t *testing.T, testName string, args []string, testOpt
 		_, err := os.Stat(hclFilePath)
 		if err == nil {
 			testutil.AssertGoldenFile(t, hclFilePath, actual)
+			return
```
hugorut, line 136: moved this here to return only if we find a `hcl` specific file. Having a return in the prior place was a bug as it was ignoring golden files in some HCL tests

internal/hcl/evaluator.go
```diff
@@ -521,7 +521,8 @@ func (e *Evaluator) loadModule(b *Block) (*ModuleCall, error) {
 func (e *Evaluator) loadModules() []*ModuleCall {
 	var moduleDefinitions []*ModuleCall
 
-	expanded := e.expandBlocks(e.module.Blocks.OfType("module"))
+	// @todo if a module uses a count that depends on a module output, then the block expansion might be incorrect.
```
aliscott, line 524:
```suggestion
	// TODO: if a module uses a count that depends on a module output, then the block expansion might be incorrect.
```
Just to be consistent with our other TODOs.

aliscott (approved):
LGTM

PR #1573, opened 2022-04-19
feat(google): Add usage parameter for Compute Address/Global Address

### LINKED ISSUE #1572: Set IP address type for `google_compute_address` resource
Based on our community Slack discussion:

> Regarding to issue https://github.com/infracost/infracost/issues/421, I would like to know the usage file format of google_compute_address/google_compute_global_address because it’s not quite clear in the doc. I want to use just one type of the IP address, how can I specify it in usage file.

Currently the resource doesn't support usage params and shows all cost components with prices only:
```
├─ IP address (if used by standard VM)                                                    730  hours                  $2.92 
├─ IP address (if used by preemptible VM)                                                 730  hours                  $1.46 
└─ IP address (if unused)                                                                 730  hours                  $0.00 
```

We can add a new usage param `address_type` with values like `standard_vm`/`preemptible_vm`/`unused`:
- If the flag is set show only the corresponding cost component (without `(if...)` part).
- Otherwise show all cost components as before.

### Note for community
Please 👍 if you'd like to upvote this or receive updates.

vdmgolub:
Resolves #1572.


aliscott (approved):
LGTM

aliscott (approved):
LGTM

PR #1580, opened 2022-04-22
chore(aws): Load EBS block devices from LT for Instance resource

vdmgolub:


internal/providers/terraform/aws/instance.go
```diff
@@ -87,27 +103,60 @@ func NewInstance(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
...
-			IOPS:    data.Get("iops").Int(),
+			Type:    volumeType,
+			Size:    volumeSize,
+			IOPS:    iops,
 		}
 
-		if data.Get("volume_size").Type != gjson.Null {
-			ebsBlockDevice.Size = intPtr(data.Get("volume_size").Int())
-		}
+		delete(ltEBSBlockDevices, deviceName)
```
vdmgolub, line 147: @hugorut FYI, replaced nils with `delete`.

PR #1583, opened 2022-04-26
fix: compare-to tweaks

hugorut:
several minor fixes and tweaks to `--compare-to` flag:

* fixes panic error
* protects `--compare-to` flag with invalid output formats
* if project exists in prior state but not in current, show 100% deletion
* tweak help text


aliscott (approved):
LGTM

PR #1618, opened 2022-05-02
feat: add project-type flag and default to HCL parsing

aliscott:


cmd/infracost/breakdown_test.go
```diff
@@ -262,6 +262,8 @@ func TestBreakdownInitFlagsError(t *testing.T) {
 			"breakdown",
 			"--path",
 			path.Join("./testdata", testutil.CalcGoldenFileTestdataDirName()),
+			"--project-type",
+			"terraform_cli",
```
hugorut, line 266: do we want to force `terraform_cli` here or should there be `hcl` coverage?

cmd/infracost/diff.go
```diff
@@ -64,8 +66,8 @@ func checkDiffConfig(cfg *config.Config) error {
 			return errors.New("terraform_use_state cannot be used with `infracost diff` as the Terraform state only contains the current state")
 		}
 
-		if projectConfig.TerraformParseHCL && cfg.CompareTo == "" {
-			return errors.New("Use `infracost diff --path /code --terraform-parse-hcl --compare-to infracost-previous-run.json` to generate a diff")
+		if projectConfig.ProjectType == "terragrunt_dir" && cfg.CompareTo == "" {
```
hugorut, line 69: should we also include `terraform_dir` case here? Or is that already covered?

cmd/infracost/testdata/diff_with_config_file_compare_to/prior.json
```diff
@@ -6,7 +6,7 @@
       "name": "infracost/infracost/cmd/infracost/testdata/diff_with_config_file_compare_to/dev",
       "metadata": {
         "path": "dev",
-        "type": "terraform_dir",
+        "type": "terraform_cli",
```
hugorut, line 9: I think all the compare-to JSON files with this change can just be changed back to `terraform_dir`. It seems like this might have been a find and replace issue

internal/hcl/parser.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 // This sets a global logger for this package, which is a bit of a hack. In the future we should use a context for this.
-var log = logrus.StandardLogger().WithField("parser", "terraform_hcl")
+var log = logrus.StandardLogger().WithField("parser", "terragrunt_dir")
```
hugorut, line 24: should this not remain the same?

internal/providers/detect.go
```diff
@@ -70,67 +71,74 @@ func Detect(ctx *config.ProjectContext, includePastResources bool) (schema.Provi
...
-		return &ValidationError{
-			err: "CloudFormation projects are not yet supported by the parse HCL option.",
-		}
-	}
-
-	if isTerraformPlanJSON(path) {
+	if ctx.ProjectConfig.TerraformPlanFlags != "" {
 		return &ValidationError{
-			err: "Path type cannot be a Terraform plan JSON file when using the parse HCL option\n\nSet --path to a Terraform directory.",
+			warn: "terraform-plan-flags is deprecated with Terraform HCL parsing, use --terraform-var or --terraform-var-file instead.",
```
hugorut, line 129: maybe this should say "not supported" over "deprecated" as HCL doesn't use `terraform-plan-flags`

internal/providers/detect.go
```diff
@@ -70,67 +71,74 @@ func Detect(ctx *config.ProjectContext, includePastResources bool) (schema.Provi
...
 		}
 	}
 
-	if isTerraformStateJSON(path) {
+	if ctx.ProjectConfig.TerraformInitFlags != "" {
 		return &ValidationError{
-			err: "Path type cannot be a Terraform state file when using the parse HCL option\n\nSet --path to a Terraform directory.",
+			warn: "terraform-init-flags is deprecated with Terraform HCL parsing.",
```
hugorut, line 135: same as above

hugorut (commented):
small tweaks and a few questions. Looks good though.

internal/providers/detect.go
aliscott: I'm not sure. Technically it is supported at the moment, since we'll parse out the `-var` and `-var-file` flags from it.

hugorut: Ah, in that case deprecated works

cmd/infracost/breakdown_test.go
aliscott: Done, and added a separate golden file with the `terraform_dir` error that shows here.

internal/providers/detect.go
aliscott: I've updated this one since it's not supported.

cmd/infracost/diff.go
aliscott: Cheers, looks like this actually had a bug, so have fixed it and added a test.

hugorut: 👍

hugorut:
@aliscott is this good for another review?

cmd/infracost/cmd_test.go
```diff
@@ -43,20 +47,29 @@ func DefaultOptions() *GoldenFileOptions {
...
-			goldenFileCommandTest(t, testName, args, testOptions, false, ctxOptions...)
+			tfCLI := make([]string, len(args)+2)
+			copy(tfCLI, args)
+			tfCLI[len(args)] = "--project-type"
+			tfCLI[len(args)+1] = "terraform_cli"
+			goldenFileCommandTest(t, testName, tfCLI, testOptions, false, ctxOptions...)
+		})
+	}
+
+	if testOptions != nil && (testOptions.RunTerragruntCLI || testOptions.OnlyRunTerragruntCLI) {
```
aliscott, line 66: This isn't super nice, I thought about having a `ProjectTypes` option to pass to the tests but didn't want to refactor a lot of the tests just now.

hugorut: yeah agreed, it's not great, let's leave it out of the scope of this PR

aliscott:
> @aliscott is this good for another review?
Ready now

hugorut (approved):
LGTM

PR #1620, opened 2022-05-03
fix: update compare-to flag to only be used by diff

hugorut:
* removes `--compare-to` from `breakdown` and `output` to reduce complexity
* adds `--format` to `diff` cmd making this the main entry point for `--compare-to`


cmd/infracost/diff.go
```diff
@@ -45,14 +45,14 @@ func diffCmd(ctx *config.RunContext) *cobra.Command {
...
 
-			ctx.Config.Format = "diff"
-
 			return runMain(cmd, ctx)
 		},
 	}
 
 	addRunFlags(cmd)
 
+	cmd.Flags().String("compare-to", "", "Path to Infracost JSON file to compare against, cannot be used with table and html formats")
```
aliscott, line 54:
```suggestion
	cmd.Flags().String("compare-to", "", "Path to Infracost JSON file to compare against")
```
Am I right in saying the other part of the message is no longer needed?

aliscott (approved):

cmd/infracost/diff.go
hugorut: 👍

PR #1622, opened 2022-05-04
fix: remove spinner if workspace does not use remote execution

hugorut:
only new up a spinner if the execution mode is remote

### DIFF
--- a/internal/hcl/remote_variables_loader.go
+++ b/internal/hcl/remote_variables_loader.go
@@ -123,12 +123,6 @@ func (r *RemoteVariablesLoader) Load(blocks Blocks) (map[string]cty.Value, error
 		}
 	}
 
-	var spinner *ui.Spinner
-	if r.newSpinner != nil {
-		spinner = r.newSpinner(spinnerMsg)
-		defer spinner.Success()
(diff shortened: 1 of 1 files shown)


internal/hcl/remote_variables_loader.go
```diff
@@ -123,40 +123,34 @@ func (r *RemoteVariablesLoader) Load(blocks Blocks) (map[string]cty.Value, error
...
-		return vars, errors.New("unable to parse Workspace response")
+		log.Warnf("malformed Terraform API response using workspace: %s organization: %s error: %s", config.workspace, config.organization, err)
+		return vars, nil
 	}
 
 	if workspaceResponse.Data.Attributes.ExecutionMode != "remote" {
-		if spinner != nil {
-			spinner.Fail()
-		}
+		log.Debugf("Terraform workspace %s does not use remote execution continuing with local execution", config.workspace)
```
aliscott, line 144:
```suggestion
		log.Debugf("Terraform workspace %s does not use remote execution, skipping downloading remote variables", config.workspace)
```

aliscott (approved):

PR #1624, opened 2022-05-05
feat: add infracost json support for diff

hugorut:
Adds the ability for users to compare to Infracost runs using `diff`. `diff` now accepts an Infracost JSON file as a valid `--path` type.


cmd/infracost/run.go
```diff
@@ -226,36 +153,23 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
...
-		b, err = output.ToJSON(r, opts)
-	case "html":
-		b, err = output.ToHTML(r, opts)
-	case "diff":
-		b, err = output.ToDiff(r, opts)
-	default:
-		b, err = output.ToTable(r, opts)
-	}
-
+	b, err := output.MarshalOutput(format, r, output.Options{
```
aliscott, line 162:
```suggestion
	b, err := output.FormatOutput(format, r, output.Options{
```
Being picky here, but only because marshal to me means serializing (i.e. converting to a machine format).

aliscott (approved):
LGTM

cmd/infracost/run.go
hugorut: 👍

PR #1630, opened 2022-05-08
feat: add --terraform-force-cli flag

aliscott:
This does the following:
1. Defaults to using the HCL parsing if the path is a directory
2. Adds a hidden `--terraform-force-cli` flag to get the old functionality
3. Hides and deprecates the `--terraform-plan-flags`, `--terraform-init-flags` and `--terraform-use-state` flags.


cmd/infracost/diff.go
```diff
@@ -127,12 +127,14 @@ func checkDiffConfig(cfg *config.Config) error {
...
-			projectType = providers.DetectProjectType(projectConfig.Path)
-		}
+		projectType := providers.DetectProjectType(projectConfig.Path, projectConfig.TerraformForceCLI)
 		if (projectType == "terraform_dir" || projectType == "terragrunt_dir") && cfg.CompareTo == "" {
-			return errors.New("Use `infracost diff --path /code --compare-to infracost-previous-run.json` to generate a diff")
+			msg := fmt.Sprintf(`To show a diff:
+  1. Generate a cost estimate baseline: %s
+  2. Make a Terraform code change
+  3. Generate a cost estimate diff: %s`,
+				"`infracost breakdown --path /code --format json --out-file infracost-base.json`", "`infracost diff --path /code --compare-to infracost-base.json`")
```
tim775, line 136: Since we have the cmd here already, it might be nice if we show the path they already used in the example instead of `--path /code`.  (This is something Ali and I were talking about for "empty diffs")

Maybe we only do that if they provided a single path to avoid weird corner cases.  Something like:
```go
example_path := "/code"
if paths, _ := cmd.Flags().GetStringArray("path"); len(paths) == 1 {
  example_path = paths[0]
}
...
msg := fmt.Sprintf(`To show a diff: .... `infracost breakdown --path %s --format ...`, ... example_path);
```

tim775 (approved):

cmd/infracost/diff.go
aliscott: Good idea. I've updated it to use the path passed by the projectConfig.

PR #1631, opened 2022-05-08
feat: enable hcl multi-project autodetect

hugorut:
Enables hcl multi-project autodetection. This allows users to run `infracost breakdown` at the repo level and get an automatic project output, no config file required. e.g:

With a project structure as such:

```
foo/
   dev/
   prod/
```

running `infracost breakdown --path foo/` would return something similar to:

```
Project:  foo/dev

 Name                                                   Monthly Qty  Unit   Monthly Cost 

 aws_instance.web_app                                                                    
 ├─ Instance usage (Linux/UNIX, on-demand, m5.4xlarge)          730  hours       $560.64 
 ├─ root_block_device                                                                    
 │  └─ Storage (general purpose SSD, gp2)                        50  GB            $5.00 
 └─ ebs_block_device[0]                                                                  
    ├─ Storage (provisioned IOPS SSD, io1)                    1,000  GB          $125.00 
    └─ Provisioned IOPS                                         800  IOPS         $52.00 

 Project total                                                                   $742.64 

Project: foo/prod

 Name                                                   Monthly Qty  Unit   Monthly Cost 

 aws_instance.web_app                                                                    
 ├─ Instance usage (Linux/UNIX, on-demand, m5.8xlarge)          730  hours     $1,121.28 
 ├─ root_block_device                                                                    
 │  └─ Storage (general purpose SSD, gp2)                        50  GB            $5.00 
 └─ ebs_block_device[0]                                                                  
    ├─ Storage (provisioned IOPS SSD, io1)                    1,000  GB          $125.00 
    └─ Provisioned IOPS                                         800  IOPS         $52.00 

 Project total                                                                 $1,303.28 

 OVERALL TOTAL                                                                 $2,045.92 
```

## Why

1. Simplifies running `infracost` in a multi-project setting. This means it's super easy for users to migrate from CLI to CI.
2. This autodetection functionality is useful for editor extensions (vscode, e.t.c) that normally run in the context of a repo, not a specific project.

---

This PR also adds file metadata to the infracost JSON output. e.g:

```
"resources": [
  {
    "name": "aws_instance.web_app",
    "metadata": {
      "infracost_filename": "examples/terraform/main.tf"
    },
    "hourlyCost": "1.017315068493150679",
    "monthlyCost": "742.64",
    "costComponents": [
        ...
    ],
  }
]
```

This helps editor extensions locate the resources that are in the current file the user is editing. 

---

@alikhajeh1 / @aliscott - I've added this as a PR to the `v0.10` branch, but I'm not sure if we want to block the `v0.10` release on this. I think it might be better to just get `v0.10` out. However, this might be a nice bonus feature that could make upgrading to `v0.10` more appealing.


aliscott:
@hugorut heads up, I had to rebase this again.

hugorut:
Thanks @aliscott

internal/output/output.go
```diff
@@ -288,9 +288,16 @@ func outputResource(r *schema.Resource) Resource {
 		subresources = append(subresources, outputResource(s))
 	}
 
+	metadata := make(map[string]string)
+	if r.InfracostMetadata != nil {
+		for k, v := range r.InfracostMetadata {
+			metadata["infracost_"+k] = v.String()
```
aliscott, line 294: If I'm understanding this correctly this is just to store data in the "plan JSON-style" output we then pass the the plan JSON parser? What do we see being in there longer term? I'm assuming things like line numbers as well.

internal/schema/resource.go
```diff
@@ -25,6 +26,7 @@ type Resource struct {
 	UsageSchema       []*UsageItem
 	EstimateUsage     EstimateFunc
 	EstimationSummary map[string]bool
+	InfracostMetadata map[string]gjson.Result
```
aliscott, line 29: Does this still need to be called `InfracostMetadata` at this point? Should it be called `Metadata` since that maps to the `output.Resource` struct?

internal/output/output.go
aliscott: Also do we need to prefix these with `infracost_`? Since this goes into the Infracost JSON format then it seems redundant to have that prefix.

aliscott (commented):
LGTM

internal/schema/resource.go
hugorut: Yeah agreed

internal/output/output.go
hugorut: Yeah I think line numbers are probably the next thing we add. Re the Infracost prefix, yeah I agree it's not needed. Not really sure why it was there.

aliscott:
@alikhajeh1 I guess 3 is unrelated to this PR correct? We'd want that in v0.10 if we include this or not?

alikhajeh1:
@aliscott tru dat, we'd want that in v0.10

aliscott:
@alikhajeh1 I've fixed 3 in my branch (https://github.com/infracost/infracost/pull/1630/commits/9753daa59c786195d265b0770f333a885f03184e)

hugorut:
Ok I've fixed up the changes requested by @aliscott and changed project detection to a depth of 5.

However, this highlighted an issue: The HCL provider detects nested Terraform projects by finding folders that have valid Terraform configuration files (.tf/.tf.json), but also makes sure that these files contain a `provider` block. I did this as I wanted a way to ignore shared modules that might be used across a mono repo. This is exactly the setup we have for the Infracost infra repo (with prod and dev folders and a shared modules folder).

Terraform docs state that provider blocks should only be [added at the root module level](https://www.terraform.io/language/providers/configuration#provider-configuration-1) so I thought this method would be relatively safe for detecting valid projects.

Yet, I ran `breakdown` with the depth change on the `infra` project for Infracost and it now thinks that `infra/modules/base` is a project. That's because it has a `provider "kubernetes"` block. If we're to believe Terraform docs this is actually invalid and this should be configured at the root level, but Terraform doesn't seem to error with this configuration.

So my question is should we proceed with this method of "asserting" Terraform projects (1. valid tf files, 2. valid provider blocks) or is there a better way? Thoughts @aliscott / @alikhajeh1 

*p.s the above only applied to nested projects, if a user points Infracost to a top level directory that contains Terraform files we don't validate the provider block. This is done so that users can see cost snapshots of things like third party modules*

aliscott:
> So my question is should we proceed with this method of "asserting" Terraform projects (1. valid tf files, 2. valid provider blocks) or is there a better way? Thoughts @aliscott / @alikhajeh1

Other options off the top of my head - still thinking of others...

2. Exclude any directory called `modules`. This seems to be a fairly well-used path. We could even have an `--exclude-paths` flag that defaults to `**/modules`.

3. Check the `.terraform/modules/modules.json` file of each directory and exclude any local paths that appear in there. This will only work if the project has been init'd.

hugorut:
@aliscot Yeah I was initially thinking of option 2 with something like the `--exclude-paths` flag. I believe anything option that involves initing should be avoided, this could have a real speed penalty.

I think option 5 could work and should be pretty easy. i.e right now we just do a shallow parse of the files to return the `provider` blocks, which could modify this to also return the `module` blocks and fetch the `src` attribute. This would build an exclusion list on which we filter the projects.

@tim775 yup nice suggestion, we actually use this library in the module downloading functionality. However, it might be easier and quicker to just do the shallow parse as it'll probably be quicker because we ignore blocks we don't need. Plus we're already doing it this way, so it's less code to change.

aliscott:
> I believe anything option that involves initing should be avoided, this could have a real speed penalty.

Yes if we assume it has been init'd this will only work locally, and if we try initing ourselves then we'll require the terraform binary. So either way these options aren't suitable.

hugorut:
@aliscott I've implemented **option 5**. This branch is technically good to go, but I'll hand it off to you to make any tweaks and additions that you see fit. I don't want to slow down the release because I'm on holiday. I'm assigning this to you now.

PR #1637, opened 2022-05-10
feat: Add auth login command

vdmgolub:


tim775 (approved):

alikhajeh1:
Can we please make an integration branch (from master) for the new auth stuff and merge this into that? That way we can keep tweaking via PRs to the integration branch, but not mess with a potential v0.10.2 release

PR #1639, opened 2022-05-10
feat(google): Show only one cost component for Google Compute Address

### LINKED ISSUE #1619: google_compute_address removal and cost calculation
Whenever we remove a public compute address from a VM in GCP, infracost shows this:

```
- google_compute_address.my-instance
  -$11.68

    - IP address (if used by standard VM)
      -$2.92

    - IP address (if used by preemptible VM)
      -$1.46

    - IP address (if unused)
      -$7.30
```

Then it sums all these 3 costs and says our total cost will decrease by $11.68, which does not make much sense.

The only cost decrease we are expecting is the $2.92 for a in-use IP address and nothing else.

How can we improve infracost's understanding in this case?

vdmgolub:
Fixes #1619.

Showing 3 cost components simultaneously for IP address is not correct,
if we can't determine which price to use, we should have a default case.

The usage param is removed as redundant.


internal/providers/terraform/google/testdata/compute_address_test/compute_address_test.golden
```diff
@@ -1,39 +1,47 @@
...
+ └─ IP address (preemptible VM)                                    730  hours                    $1.46 
+                                                                                                       
+ google_compute_address.standard_static                                                                
+ └─ IP address (standard VM)                                       730  hours                    $2.92 
+                                                                                                       
+ google_compute_address.standard_with_usage                                                            
+ └─ IP address (standard VM)                                       730  hours                    $2.92 
+                                                                                                       
+ google_compute_address.static                                                                         
+ └─ IP address (unused)                                 Monthly cost depends on usage: $0.01 per hours
```
tim775, line 20: It looks like this is a GCE_ENDPOINT that isn't attached to an instance.  Doesn't that mean it should be charged?  Why do we have the "Monthly cost depends on usage..." message?

internal/providers/terraform/google/testdata/compute_address_test/compute_address_test.golden
```diff
@@ -1,39 +1,47 @@
...
- OVERALL TOTAL                                                                $49.64 
+ Name                                                      Monthly Qty  Unit              Monthly Cost 
+                                                                                                       
+ google_compute_address.invalid_usage                                                                  
+ └─ IP address (unused)                                 Monthly cost depends on usage: $0.01 per hours 
+                                                                                                       
+ google_compute_address.preemptible_static                                                             
+ └─ IP address (preemptible VM)                                    730  hours                    $1.46 
+                                                                                                       
+ google_compute_address.preemptible_with_usage
```
tim775, line 10: Can we get rid of the usage parameter now since we can detect the purpose and attachment status?  What is the case where we still need to support the usage param?

internal/providers/terraform/google/testdata/compute_address_test/compute_address_test.golden
vdmgolub: The missing usage param takes priority. Without it, it should show the cost. 👍

internal/providers/terraform/google/testdata/compute_address_test/compute_address_test.golden
vdmgolub: I kept it for backward-compatibility, but thinking about it again, I think you're right, it brings more confusion. I'll remove it 👍

tim775 (approved):
Looks good!

PR #1645, opened 2022-05-11
feat: Add support for google_compute_per_instance_config

### LINKED ISSUE #1644: Add support for google_compute_per_instance_config
When using a managed instance group (the `google_compute_instance_group_manager` resource) in GCP, you can add stateful instances (the `google_compute_per_isntance_config` resource).

This resource does not incur cost, but it will increase the `target_size` of the group manager (which incurs cost), and this is not taken into account by `infracost` today.

golgeek:
Closes #1644.

I was not entirely sure on how to implement this as `google_compute_per_instance_config` is a free resource that will increase the cost of `google_compute_instance_group_manager`.

I added a reversed reference from the `google_compute_instance_group_manager` resource to the `google_compute_per_instance_config` to increase the `target_size` and compute the actual cost, and did not add any cost for the `google_compute_per_instance_config` resource.

### DIFF
--- a/internal/providers/terraform/google/compute_instance_group_manager.go
+++ b/internal/providers/terraform/google/compute_instance_group_manager.go
@@ -10,7 +10,10 @@ func getComputeInstanceGroupManagerRegistryItem() *schema.RegistryItem {
 		Name:                "google_compute_instance_group_manager",
 		RFunc:               newComputeInstanceGroupManager,
 		Notes:               []string{"Multiple versions are not supported."},
-		ReferenceAttributes: []string{"version.0.instance_template"},
+		ReferenceAttributes: []string{"version.0.instance_template", "google_compute_per_instance_config.instance_group_manager"},
+		CustomRefIDFunc: func(d *schema.ResourceData) []string {
+			return []string{d.Get("self_link").String(), d.Get("name").String()}
+		},
 	}
 }
 
@@ -24,6 +27,9 @@ func newComputeInstanceGroupManager(d *schema.ResourceData, u *schema.UsageData)
 	targetSize := int64(1)
 	if d.Get("target_size").Exists() {
 		targetSize = d.Get("target_size").Int()
+	}
+	if len(d.References("google_compute_per_instance_config.instance_group_manager")) > 0 {
+		targetSize += int64(len(d.References("google_compute_per_instance_config.instance_group_manager")))
 	}
 
 	var machineType string
--- a/internal/providers/terraform/google/compute_per_instance_config.go
+++ b/internal/providers/terraform/google/compute_per_instance_config.go
@@ -1,0 +1,14 @@
+package google
+
+import (
+	"github.com/infracost/infracost/internal/schema"
+)
+
+func getComputePerInstanceConfigRegistryItem() *schema.RegistryItem {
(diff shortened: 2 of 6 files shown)


internal/providers/terraform/google/compute_per_instance_config.go
```diff
@@ -0,0 +1,25 @@
...
+
+func newComputePerInstanceConfig(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+
+	r := &google.ComputePerInstanceConfig{
+		Address: d.Address,
+	}
+
+	r.PopulateUsage(u)
+
+	return r.BuildResource()
```
tim775, line 24: Since this resource has no cost component you can make it free like this: 
```suggestion
	return &schema.RegistryItem{
		Name:                "google_compute_per_instance_config",
		NoPrice: true,
		ReferenceAttributes: []string{"instance_group_manager"},
		Notes: []string{"Free resource."},
	}
```

Once that is done you can also remove internal/resources/google/compute_per_instance_config.go

internal/providers/terraform/google/compute_instance_group_manager.go
```diff
@@ -10,7 +10,10 @@ func getComputeInstanceGroupManagerRegistryItem() *schema.RegistryItem {
 		Name:                "google_compute_instance_group_manager",
 		RFunc:               newComputeInstanceGroupManager,
 		Notes:               []string{"Multiple versions are not supported."},
-		ReferenceAttributes: []string{"version.0.instance_template"},
+		ReferenceAttributes: []string{"version.0.instance_template", "google_compute_per_instance_config.instance_group_manager"},
+		CustomRefIDFunc: func(d *schema.ResourceData) []string {
+			return []string{d.Get("self_link").String(), d.Get("name").String()}
```
tim775, line 15: Nicely done adding the reference for `self_link`.  Am I correct in thinking we should be automatically adding this for every GCE resource (similar to how we automally add `arn` as a reference for AWS resources)?

tim775 (changes requested):
This is great, thanks!  I added one suggestion for a cleaner way to implement the free resource but other than that it looks good.

internal/providers/terraform/google/compute_instance_group_manager.go
golgeek: > Nicely done adding the reference for `self_link`. Am I correct in thinking we should be automatically adding this for every GCE resource (similar to how we automally add `arn` as a reference for AWS resources)?

Yes, all google resources have a `self_link` attached (which usually looks like the `id` field), and it could be great to auto add it.

In this specific case, the `name` field is required because that's how the `per_instance_config` is linked to the `group_manager`, but it's kinda exceptional and links usually use `id` or `self_link`.

golgeek:
> This is great, thanks! I added one suggestion for a cleaner way to implement the free resource but other than that it looks good.

Thanks for your review, I just pushed your suggested changes!

internal/providers/terraform/google/compute_instance_group_manager.go
tim775: 👍  I created a new issue for that.

Issue #1575, opened 2022-04-20
Support AWS RDS Reserved Instances


raserma:
AWS EC2 RI is already supported by using the Infracost usage file. Could we extend the support for RDS Reserved Instances as well? Thanks! 🙌 

---

### Infracost details

Add Reserved Instances usage params for [aws_rds_cluster_instance](https://github.com/infracost/infracost/blob/master/internal/resources/aws/rds_cluster_instance.go) and [aws_db_instance](https://github.com/infracost/infracost/blob/master/internal/resources/aws/db_instance.go) resources:

- `reserved_instance_term` string
- `reserved_instance_payment_option` string

And use them when calculating Instance usage cost component. You can use [AWS instance resource](https://github.com/infracost/infracost/blob/master/internal/resources/aws/instance.go#L37-L39) as example.

### Note for community
Please 👍 if you'd like to upvote this or receive updates.
Add a comment if you'd like to [contribute](https://github.com/infracost/infracost/blob/master/contributing/add_new_resource_guide.md#add-usage-based-cost-components).

vdmgolub:
@raserma Hello! Are you referring to [aws_rds_cluster_instance](https://registry.terraform.io/providers/aaronfeng/aws/latest/docs/resources/rds_cluster_instance) resource? I just checked our cloud pricing API and indeed, it contains prices for RI and it seems like a good idea.

We can follow the same logic AWS Instance does here with usage based params: https://github.com/infracost/infracost/blob/master/internal/resources/aws/instance.go#L37-L39.

Please let me know if you’d like to [contribute](https://github.com/infracost/infracost/blob/master/contributing/add_new_resource_guide.md#add-usage-based-cost-components), otherwise I’ll open it up for the whole community.

vdmgolub:
Sounds good! I'll update your first comment with some details to keep them at the top. :) Thank you for creating the issue! We'll notify here when it's worked on and merged.

carmeloriolo:
Hello! I'd like to help with this issue!
Can I have it assigned?
Thanks :)

aliscott:
Thanks @carmeloriolo I've assigned this to you. Please reach out if we can help in any way 🙏

aliscott:
Thanks @carmeloriolo, tagging @vdmgolub to give this a look when he gets a chance.

vdmgolub:
[...]

PR #1653, opened 2022-05-13
feat(aws): added support for reservation in db_instance

carmeloriolo:
Implements #1575 

This PR adds Reserved Instances usage params:
```
ReservedInstanceTerm                         *string `infracost_usage:"reserved_instance_term"`
ReservedInstancePaymentOption                *string `infracost_usage:"reserved_instance_payment_option"`
```

for the resources [db_instance.go](https://github.com/infracost/infracost/pull/1653/files#diff-733dbca11c73a2081f9713ed7cfeee44637bcb13d69e57a8c3f8d241c73bfe28) and [rds_cluster_instance.go](https://github.com/infracost/infracost/pull/1653/files#diff-7a64634ce99019460b5ba69e647775216f6d39eb5a324691656bebdead1f262a). 

* Updated the compute of the cost component for both classes.
* Updated test cases and usage files.
* Moved some common logics in [util.go](https://github.com/infracost/infracost/pull/1653/files#diff-1ff680d3552a9984ec4c896cb493bc60d11dec5341e7acd066dd81569590c928) to avoid code duplication in the case of reserved resources:
    * `reservedTermsMapping`
    * `reservedPaymentOptionMapping`
    * `validatereservedinstanceparams(reservedInstanceTerm, reservedInstancePaymentOption string) (bool, string)`
* Added an util function named [sliceOfKeysFromMap](https://github.com/infracost/infracost/blob/86c3dff55bbd5689ce5bc7110e00c82f7031897f/internal/resources/aws/util.go#L92) in util.go to get a slice with all the keys of a map (This function use generics feature  introduced in Go 1.18). This has been used in `validatereservedinstanceparams` to get slices of keys for `reservedTermsMapping` and `reservedPaymentOptionMapping`
* Updated also instance.go (even if out of the scope of this PR) to use the previous utilities about reservation.


vdmgolub:
@carmeloriolo Thank you for fixing the function name! Just a heads-up that I'm reviewing and testing your PR today, will let you know how it goes :)

internal/resources/aws/util.go
```diff
@@ -662,3 +670,26 @@ var InstanceTypeToVCPU = map[string]int64{
...
+	validTerms := sliceOfKeysFromMap(reservedTermsMapping)
+	if !stringInSlice(validTerms, reservedInstanceTerm) {
+		return false, fmt.Sprintf("Invalid reserved_instance_term, ignoring reserved options. Expected: 1_year, 3_year. Got: %s", reservedInstanceTerm)
+	}
+	validOptions := sliceOfKeysFromMap(reservedPaymentOptionMapping)
+	if !stringInSlice(validOptions, reservedInstancePaymentOption) {
+		return false, fmt.Sprintf("Invalid reserved_instance_payment_option, ignoring reserved options. Expected: no_upfront, partial_upfront, all_upfront. Got: %s", reservedInstancePaymentOption)
+	}
+	return true, ""
+}
```
vdmgolub, line 695: After poking around AWS pricing calculator I found another distinction from EC2 instance: `3_year` term doesn't have `no_upfront` value. If we allow to pass it, we'll end up with a warning "No products found". I think we can improve the validation here. I still think reusing it for EC2 instance is a great idea, so what do you think if we organize the code under a struct, so it is more seamless? Here's an example:
```go
type reservedInstanceResolver struct {
	term          string
	paymentOption string
	dbInstance    bool
}

func (r reservedInstanceResolver) Term() string {
	return reservedTermsMapping[r.term]
}

func (r reservedInstanceResolver) PaymentOption() string {
	return reservedPaymentOptionMapping[r.paymentOption]
}

func (r reservedInstanceResolver) validate() (bool, string) {
	validTerms := sliceOfKeysFromMap(reservedTermsMapping)
	if !stringInSlice(validTerms, r.term) {
		return false, fmt.Sprintf("Invalid reserved_instance_term, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validTerms, ", "), r.term)
	}
	validOptions := sliceOfKeysFromMap(reservedPaymentOptionMapping)

	if r.term == "3_year" && r.dbInstance {
		validOptions = []string{"partial_upfront", "all_upfront"}
	}

	if !stringInSlice(validOptions, r.paymentOption) {
		return false, fmt.Sprintf("Invalid reserved_instance_payment_option, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validOptions, ", "), r.paymentOption)
	}
	return true, ""
}

# EC2 instance
resolver := reservedInstanceResolver{
	term:          strVal(r.ReservedInstanceTerm),
	paymentOption: strVal(r.ReservedInstancePaymentOption),
}

# DB/RDS instance
resolver := reservedInstanceResolver{
	term:          strVal(r.ReservedInstanceTerm),
	paymentOption: strVal(r.ReservedInstancePaymentOption),
	dbInstance:    true,
}

# And then use it like
valid, err := resolver.validate()

# And hide the mapping maps with methods:
TermLength:         strPtr(resolver.Term()),
TermPurchaseOption: strPtr(resolver.PaymentOption()),
```

internal/resources/aws/instance.go
```diff
@@ -196,17 +196,7 @@ func (a *Instance) validateReserveInstanceParams() (bool, string) {
...
-		return false, fmt.Sprintf("Invalid reserved_instance_term, ignoring reserved options. Expected: 1_year, 3_year. Got: %s", strVal(a.ReservedInstanceTerm))
-	}
-
-	validOptions := []string{"no_upfront", "partial_upfront", "all_upfront"}
-	if !stringInSlice(validOptions, strVal(a.ReservedInstancePaymentOption)) {
-		return false, fmt.Sprintf("Invalid reserved_instance_payment_option, ignoring reserved options. Expected: no_upfront, partial_upfront, all_upfront. Got: %s", strVal(a.ReservedInstancePaymentOption))
-	}
-
-	return true, ""
+	return validateReservedInstanceParams(strVal(a.ReservedInstanceTerm), strVal(a.ReservedInstancePaymentOption))
```
vdmgolub, line 199: On lines 203-212 we could also reuse the new mappings.

vdmgolub (commented):
Great job! Thank you for the refactoring! :) All looks good, just one comment about `3_year` term and `no_upfront` option.

 I've also pushed a tiny change to add examples and missing UsageSchema items.

internal/resources/aws/util.go
carmeloriolo: Thank you @vdmgolub for the review! 

My bad I missed the UsageSchema items for RDSClusterInstance! Thank you for fixing it :)   

I also noticed too that 3_year term doesn't have no_upfront. In fact while updating the tests I got that warning but I thought this was not a problem because also terraform won't generate any plan in that case. So I considered it as an "impossibile" input.  But of course I agree that validation can be improved to avoid noisy warnings for that case. 

Great idea creating a new struct to handle and embed everything regarding reservation. This can be useful for all reservation cases, for example another case is the one related to OpenSearch reserved instances (I was thinking about opening an issue to implement this too). So we may can consider some other kind of mapping instead of several booleans like `dbInstance    bool` to differentiate all the services? Should this struct stay in util.go or we can evaluate to create a new file for it?

vdmgolub: No worries at all, team work 🤜 🤛 :) When making similar changes in several places it's easy to overlook things, talking from experience 😅.

You are correct, Terraform would complain and it is a good safety net. However with our new HCL parsing option using `--terraform-parse-hcl` flag (which, by the way, will become a default in the upcoming [v0.10](https://github.com/infracost/infracost/releases/tag/v0.10.0-beta.1) release) Terraform won't watch our backs anymore and we'll have to rely on our own validations. :)

I think we could keep it in utils for now to move this PR forward. If you consider to implement the OpenSearch changes, I think it will be a great opportunity to refactor it and decide what works better for that resource. How does it sound to you?

carmeloriolo: Cool I did not know about `--terraform-parse-hcl` feature!

You are right, just keep it simple for the moment to move the this PR forward. I pushed the reservedInstanceResolver with the new validation in the latest commit!

vdmgolub (approved):
Great job! 🚀 Thank you! :)

PR #1660, opened 2022-05-13
refactor: rework reference mapping

### LINKED ISSUE #1652: GCE resource references should work automatically with `self_link`
From [this discussion](https://github.com/infracost/infracost/pull/1645#discussion_r872457081) in #1645, all GCE resources contain a `self_link`.  We should support these `self_link` references, similar to [how we support `arn` references for AWS resources](https://github.com/infracost/infracost/blob/bcf2c70062a51acc104ef849100461de846335f5/internal/providers/terraform/parser.go#L415-L451).

golgeek:
This PR closes #1652.

I tried to (at least partially) rework `parser.go` move the provider-specific logic out of this file and into the specific packages.

I feel that the `internal/providers/terraform/*` packages could deserve to be implemented as interfaces (with GetResources(), GetFreeResources(), GetUsageOnlyResources() on top of the funcs I added in the `internal/providers/terraform/*/[provider].go` files), though it means refactoring lots of stuff in `parser.go` and `registry.go` and I'm not sure this PR is the right one for that (+ I feel a bit lazy tonight...).

Let me know what you think!

### DIFF
--- a/internal/providers/terraform/aws/aws.go
+++ b/internal/providers/terraform/aws/aws.go
@@ -1,0 +1,84 @@
+package aws
+
+import (
+	"strings"
+
+	"github.com/infracost/infracost/internal/schema"
+	log "github.com/sirupsen/logrus"
+	"github.com/tidwall/gjson"
+)
+
+var DefaultProviderRegion = "us-east-1"
+var arnAttributeMap = map[string]string{
+	"aws_cloudwatch_dashboard":     "dashboard_arn",
+	"aws_db_snapshot":              "db_snapshot_arn",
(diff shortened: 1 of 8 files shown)


aliscott:
Thanks @golgeek 🙏! Initial glance through looks good but I should get a chance to give this more of look through in the next couple of days.

aliscott (approved):
Thanks for cleaning this up @golgeek! I've manually tested this with a few resources (both before apply and after apply) and all seems great.

golgeek:
Thanks for taking the time to review this @aliscott!

PR #1666, opened 2022-05-18
enhance: add a better error message for errors originating from the Terragrunt libs

aliscott:
If someone runs into an error when running HCL parsing with Terragrunt, we weren't clarifying that the errors were from the Terragrunt libs. This updates that and points them to generate a plan JSON for just now.

Example:
Screenshot 2022-05-18 at 12 04 34


alikhajeh1:
Thanks! LGTM

alikhajeh1 (approved):

PR #1673, opened 2022-05-19
enhance: Add breakdown path hint

tim775:


cmd/infracost/testdata/output_format_azure_repos_comment/output_format_azure_repos_comment.golden
```diff
@@ -219,7 +219,7 @@ Percent: +100%
 ──────────────────────────────────
 
 The following projects have no cost estimate changes: infracost/infracost/cmd/infracost/testdata/terraform_v0.14_nochange_plan.json
-Run infracost breakdown to see their breakdown.
+Run infracost breakdown --path /path/to/code to see their breakdown.
```
vdmgolub, line 222: I would assume users would try to copy the command, but it might not be obvious what to copy. If we put it in the end that might be clearer. What do you think?
```suggestion
Run the following command to see their breakdown: infracost breakdown --path=/path/to/code
```

internal/output/diff.go
```diff
@@ -80,7 +80,7 @@ func ToDiff(out Root, opts Options) ([]byte, error) {
 	if len(noDiffProjects) > 0 {
 		s += "──────────────────────────────────\n"
 		s += fmt.Sprintf("\nThe following projects have no cost estimate changes: %s", strings.Join(noDiffProjects, ", "))
-		s += fmt.Sprintf("\nRun %s to see their breakdown.", ui.PrimaryString("infracost breakdown"))
+		s += fmt.Sprintf("\nRun %s to see their breakdown.", ui.PrimaryString("infracost breakdown --path /path/to/code"))
```
vdmgolub, line 83: Let's encourage using the equals character for flags :)
```suggestion
		s += fmt.Sprintf("\nRun %s to see their breakdown.", ui.PrimaryString("infracost breakdown --path=/path/to/code"))
```

vdmgolub (approved):
Looks good. 👍 Minor comments on formatting :)

Issue #1670, opened 2022-05-18
Add Reservation support for Elasticache


carmeloriolo:
The following resources still lack the support for Reserved instances:

- elasticache_cluster.go 
- elasticache_replication_group.go

**Some details**

**Standard reserved node cache offerings**
```
reserved_instance_term string // values: ["1_year", "3_year"]
reserved_instance_payment_option string // values: ["no_upfront", "partial_upfront", "all_upfront"]
```
**Legacy reserved node cache offerings**
Previous generation nodes
```
reserved_instance_term string // values: ["1_year", "3_year"]
reserved_instance_payment_option string // values: ["heavy_utilization", "medium_utilization", "light_utilization"]

```
The mapping between CacheNodeType and OfferingType can be retrieved via api or cli
https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/reserved-nodes-offerings.html#reserved-nodes-offerings-cli
```
aws elasticache describe-reserved-cache-nodes-offerings --output json | jq -r '[.ReservedCacheNodesOfferings[] | (.OfferingType |= ascii_downcase) |  (.OfferingType |= sub(" ";"_")) | "\"" + .CacheNodeType + "\": \"" + .OfferingType + "\""] | sort | .[]'
```

PR #1677, opened 2022-05-22
feat(aws): add elasticache support for reserved instances

carmeloriolo:
Implements #1670 

This PR:
- Adds Reservation suppor for AWS Elasticache
- Refactors `reservedInstanceResolver struct` as `reservationResolver interface` with an implementation for each involved service (_ec2ReservationResolver, rdsReservationResolver, elasticacheReservationResolver_)


hugorut:
Hi @carmeloriolo, thanks for the PR. I'll take a peek at this tomorrow morning (GMT) when I get a chance.

hugorut (changes requested):
Thanks @carmeloriolo this is great. I really appreciate the contribution and thanks for waiting on my review. Sorry I didn't get to it yesterday, it was a bit manic with the release.

---

Ok, so I'm not 100% sold on this resolver interface. I think it's probably a bit much to have a struct that has three methods:

```go
type reservationResolver interface {
	Term() string
	PaymentOption() string
	Validate() (bool, string)
}
```

When in reality all we do is call `Validate` and use that to create a `schema.PriceFilter`. That seems like a code smell to me. It's just a bit too complex for what is needed.

In reality, we want an interface with a single method:

```go
type reservationResolver interface {
	PriceFilter() (*schema.PriceFilter, error)
}
```

The rds resolver implementation that would look like such:

```go
type rdsReservationResolver struct {
	term          string
	paymentOption string
}

func (r rdsReservationResolver) PriceFilter() (*schema.PriceFilter, error) {
	termLength := reservedTermsMapping[r.term]
	purchaseOption := reservedPaymentOptionMapping[r.paymentOption]

	validTerms := sliceOfKeysFromMap(reservedTermsMapping)
	if !stringInSlice(validTerms, r.term) {
		return nil, fmt.Errorf("invalid reserved_instance_term, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validTerms, ", "), r.term)
	}

	validOptions := sliceOfKeysFromMap(reservedPaymentOptionMapping)
	if r.term == "3_year" {
		validOptions = []string{"partial_upfront", "all_upfront"}
	}

	if !stringInSlice(validOptions, r.paymentOption) {
		return nil, fmt.Errorf("invalid reserved_instance_payment_option, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validOptions, ", "), r.paymentOption)
	}

	return &schema.PriceFilter{
		PurchaseOption:     strPtr("reserved"),
		StartUsageAmount:   strPtr("0"),
		TermLength:         strPtr(termLength),
		TermPurchaseOption: strPtr(purchaseOption),
	}, nil
}
```

Then at the call site, we'd do something like:

```go
resolver := rdsReservationResolver{term: strVal(r.ReservedInstanceTerm), paymentOption: strVal(r.ReservedInstancePaymentOption)}
resolverFilter, err := resolver.PriceFilter()
if err != nil {
    log.Warnf(err)
} else {
    purchaseOptionLabel = "reserved"
    priceFilter = resolverFilter
}
```

in reality we could also make these just pure functions instead of structs, as they just do one thing. e.g:

```go
func rdsReservationPriceFilter(term, paymentOption string) (*schema.PriceFilter, error) {
	termLength := reservedTermsMapping[term]
	purchaseOption := reservedPaymentOptionMapping[paymentOption]

	validTerms := sliceOfKeysFromMap(reservedTermsMapping)
	if !stringInSlice(validTerms, term) {
		return nil, fmt.Errorf("invalid reserved_instance_term, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validTerms, ", "), term)
	}

	validOptions := sliceOfKeysFromMap(reservedPaymentOptionMapping)
	if term == "3_year" {
		validOptions = []string{"partial_upfront", "all_upfront"}
	}

	if !stringInSlice(validOptions, paymentOption) {
		return nil, fmt.Errorf("invalid reserved_instance_payment_option, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validOptions, ", "), paymentOption)
	}

	return &schema.PriceFilter{
		PurchaseOption:     strPtr("reserved"),
		StartUsageAmount:   strPtr("0"),
		TermLength:         strPtr(termLength),
		TermPurchaseOption: strPtr(purchaseOption),
	}, nil
}
```

I would probably favour the functional approach, but I'll leave it up to you to decide.

Further to this I don't think we need to stick the elasticacheResolver into the `util` file, as it's only used in one location. I'd keep it in the `elasticache_cluster` file, close to the actual callsite.

carmeloriolo:
Thanks @hugorut for the review! I like so much the solution with `PriceFilter() (*schema.PriceFilter, error)` as the only interface method, it makes everything much cleaner. I pushed the changes in the last commit. 

- `elasticacheReservationResolver` has been moved in `elasticache_cluster.go`
- `ec2ReservationResolver` in `instance.go` near where they are used.
- `rdsReservationResolver` for the moment remains in `util.go` because it is invoked both by `db_instance.go` and `rds_cluster_instance.go`.
- The `reservationResolver` interface and the several mappings remain also in `util.go`

internal/resources/aws/elasticache_cluster.go
```diff
@@ -104,3 +121,56 @@ func (r *ElastiCacheCluster) backupStorageCostComponent() *schema.CostComponent
 		},
 	}
 }
+
+// ElastiCache implementation of reservationResolver
```
hugorut, line 125: The convention for `godoc` is to have the name of the type as the first word in the comment. E.g. `elasticacheReservationResolver`. Read more here: https://go.dev/blog/godoc. Can we update all the comments on these types to make sure they're in line with this? Plus could we actually make these a bit more useful comments? I don't see a comment of `implementation of x type` as very useful for other engineers. Maybe remove comments here and just have a more in depth comment over each of the  ` PriceFilter` methods, explaining why the mapping for terms e.t.c.

internal/resources/aws/elasticache_cluster.go
```diff
@@ -104,3 +121,56 @@ func (r *ElastiCacheCluster) backupStorageCostComponent() *schema.CostComponent
...
+	cacheNodeType string
+}
+
+func newElasticacheReservationResolver(term, paymentOption, cacheNodeType string) reservationResolver {
+	return &elasticacheReservationResolver{
+		term:          term,
+		paymentOption: paymentOption,
+		cacheNodeType: cacheNodeType,
+	}
+}
```
hugorut, line 138: I don't see the value of these constructor functions. They just add noise. If you're not doing any additional functionality apart from just mapping terms to the fields, then let's remove it. If you're doing this constructor to return the interface, I would suggest against it. The general GO foo is ["accept interfaces return structs"](https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8).

internal/resources/aws/elasticache_cluster.go
```diff
@@ -104,3 +121,56 @@ func (r *ElastiCacheCluster) backupStorageCostComponent() *schema.CostComponent
...
+
+func (r elasticacheReservationResolver) IsElasticacheReservedNodeLegacyOffering() bool {
```
hugorut, line 140:
```suggestion
func (r elasticacheReservationResolver) isElasticacheReservedNodeLegacyOffering() bool {
```

as it's not exported outside this package

internal/resources/aws/elasticache_cluster.go
```diff
@@ -104,3 +121,56 @@ func (r *ElastiCacheCluster) backupStorageCostComponent() *schema.CostComponent
...
+		return nil, fmt.Errorf("Invalid reserved_instance_term, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validTerms, ", "), r.term)
+	}
+	validOptions := append(sliceOfKeysFromMap(reservedPaymentOptionMapping), sliceOfKeysFromMap(elasticacheReservedNodeCacheLegacyOfferings)...)
+
+	if !stringInSlice(validOptions, r.paymentOption) {
+		return nil, fmt.Errorf("Invalid reserved_instance_payment_option, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validOptions, ", "), r.paymentOption)
+	}
+	nodeType := strings.Split(r.cacheNodeType, ".")[1] // Get node type from cache node type. cache.m3.large -> m3
+	if stringInSlice(elasticacheReservedNodeLegacyTypes, nodeType) {
+		log.Warnf("No products found is possible for legacy nodes %s if provided payment option is not supported by the region.", strings.Join(elasticacheReservedNodeLegacyTypes, ", "))
```
hugorut, line 168: Is there any way to detect this with the information we have? I'd rather have an explicit validation and return an error.

internal/resources/aws/instance.go
```diff
@@ -154,19 +154,18 @@ func (a *Instance) computeCostComponent() *schema.CostComponent {
...
-		}
-		valid, err := a.validateReserveInstanceParams(resolver)
-		if err != "" {
-			log.Warnf(err)
-		}
-		if valid {
-			purchaseOptionLabel = "reserved"
-			return a.reservedInstanceCostComponent(osLabel, osFilterVal, purchaseOptionLabel, resolver)
+		resolver := newEc2ReservationResolver(strVal(a.ReservedInstanceTerm), strVal(a.ReservedInstancePaymentOption), strVal(a.ReservedInstanceType))
+		priceFilter, err = resolver.PriceFilter()
```
hugorut, line 164: we actually only want to explicitly set the `priceFilter` if `err` is not nil. Because if it isn't then `priceFilter` will be set to nil.

internal/resources/aws/instance.go
```diff
@@ -312,3 +271,41 @@ func (a *Instance) cpuCreditCostComponent(instanceFamily string) *schema.CostCom
...
+	termOfferingClass string
+}
+
+func newEc2ReservationResolver(term, paymentOption, termOfferingClass string) reservationResolver {
+	return &ec2ReservationResolver{
+		term:              term,
+		paymentOption:     paymentOption,
+		termOfferingClass: termOfferingClass,
+	}
+}
```
hugorut, line 288: See other comments about godoc and constructors

internal/resources/aws/rds_cluster_instance.go
```diff
@@ -44,25 +44,14 @@ func (r *RDSClusterInstance) BuildResource() *schema.Resource {
...
-		if valid {
-			purchaseOptionLabel = "reserved"
-			priceFilter = &schema.PriceFilter{
-				PurchaseOption:     strPtr("reserved"),
-				StartUsageAmount:   strPtr("0"),
-				TermLength:         strPtr(resolver.Term()),
-				TermPurchaseOption: strPtr(resolver.PaymentOption()),
-			}
+		resolver := newRdsReservationResolver(strVal(r.ReservedInstanceTerm), strVal(r.ReservedInstancePaymentOption))
+		priceFilter, err = resolver.PriceFilter()
```
hugorut, line 50: we actually only want to explicitly set the priceFilter if err is not nil. Because if it isn't then priceFilter will be set to nil.

internal/resources/aws/util.go
```diff
@@ -682,33 +687,45 @@ var reservedPaymentOptionMapping = map[string]string{
...
 }
 
-func (r reservedInstanceResolver) PaymentOption() string {
-	return reservedPaymentOptionMapping[r.paymentOption]
+func newRdsReservationResolver(term, paymentOption string) reservationResolver {
+	return &rdsReservationResolver{
+		term:          term,
+		paymentOption: paymentOption,
+	}
 }
```
hugorut, line 709: see other comments about godoc and constructors

hugorut (changes requested):
Looking much nicer, few minor tweaks

internal/resources/aws/elasticache_cluster.go
carmeloriolo: Nice article! :) I removed the interface and so the not necessary extra complexity!

internal/resources/aws/elasticache_cluster.go
carmeloriolo: Updated in the last commit 👍

internal/resources/aws/instance.go
carmeloriolo: Sorry copy and paste refuse! Fixed!

internal/resources/aws/instance.go
carmeloriolo: Removed constructor

internal/resources/aws/rds_cluster_instance.go
carmeloriolo: Sorry copy and paste refuse! Thanks for the review!

internal/resources/aws/util.go
carmeloriolo: Removed constructor

internal/resources/aws/elasticache_cluster.go
carmeloriolo: Unfortunately, actually without a dynamic validation, that I guess it is out of scope, this is not possible. 
The problem is that for a specific legacy instance, the available values may differ between regions. 
For example, `cache.m1.small` in  most regions supports `heavy_utilization, medium_utilization, light_utilization`, while in `sa-east-1` only supports `heavy_utilization`. Or `cache.m3.large` only supports `heavy_utilization` in all regions. So the solution I agreed with @vdmgolub was to warn the user about this possible behaviour in the case of reserved legacy instance.
I  described this corner case in the new comment for PriceFilter in elasticache.

internal/resources/aws/elasticache_cluster.go
carmeloriolo: Let me know if them are clear enough 🙏

internal/resources/aws/instance.go
hugorut: hmm this still looks the same for me, needs to be edited for all cases where `PriceFilter` is used

hugorut (changes requested):
Nice! Just one thing that needs changing

internal/resources/aws/instance.go
carmeloriolo: Ok now I understood it better now. But now that Validation is embedded in PriceFilter how should we set priceFilter when validation fails? 
Something like this sounds good? Or we want to embed all the properties even if those fails?
```
	if r.ReservedInstanceTerm != nil {
		purchaseOptionLabel = "reserved"
		resolver := &elasticacheReservationResolver{
			term:          strVal(r.ReservedInstanceTerm),
			paymentOption: strVal(r.ReservedInstancePaymentOption),
			cacheNodeType: r.NodeType,
		}
		reservedFilter, err := resolver.PriceFilter()
		if err != nil {
			log.Warnf(err.Error())
			priceFilter = &schema.PriceFilter{
				PurchaseOption: strPtr(purchaseOptionLabel),
			}
		} else {
			priceFilter = reservedFilter
		}
	}
```

hugorut: would the "reserved" price label still be applicable if the validation failed?

carmeloriolo: Reserved is still applicable but a warning about multiple prices found will be raised
```
"level=warning msg="Multiple prices found for aws_elasticache_cluster.redis_reserved_1yr_no_upfront ElastiCache (reserved, cache.m6g.12xlarge), using the first price"" 
```
Before of this also a validation error will be raised, as error returned from PriceFilter(), describing why validation is failing.

hugorut: Is

```go
	priceFilter := &schema.PriceFilter{
		PurchaseOption: strPtr(a.PurchaseOption),
	}
	
	if r.ReservedInstanceTerm != nil {
		purchaseOptionLabel = "reserved"
		...
		
		reservedFilter, err := resolver.PriceFilter()
		if err != nil {
			log.Warnf(err.Error())
		} else {
			priceFilter = reservedFilter
		}
	}
```

not what you want or am I missing something? Prior to this change, we did not set the `PurchaseOption` to be "reserved" so I'm unsure why you're doing that now?

carmeloriolo: Prior to this change if `err` was not `nil` then `priceFilter` was `nil` as you mentioned. 
Adding
 ```
 ...
  } else {
	  priceFilter = reservedFilter
  }
```
If `err` is not `nil` then priceFilter will remain as set before, something like:
- https://github.com/carmeloriolo/infracost/blob/feat/elasticache-reserved-instances/internal/resources/aws/db_instance.go#L143
- https://github.com/carmeloriolo/infracost/blob/feat/elasticache-reserved-instances/internal/resources/aws/rds_cluster_instance.go#L43

So if this is not set to 'reserved' it will have an 'on_demand' value

Actually this property PurchaseOption (https://github.com/carmeloriolo/infracost/blob/feat/elasticache-reserved-instances/internal/resources/aws/instance.go#L23), is present only in instance.go

Comment (hugorut) [reply internal/resources/aws/instance.go:168]: ok yeah that was were I was getting confused, I was looking at the `db_instance` code which didn't reflect the price filter as `reserved`. Yup, what you're suggesting for `instance.go` makes sense. Alternatively, you could return a default from the `PriceFilter` method on error (for instance only) which would be something like:

```go
def :=  &schema.PriceFilter{
	PurchaseOption: strPtr("reserved"),
}

...

return def, fmt.Errorf("...")
```

PR #1679, opened 2022-05-23
chore: fallback to the workspace defined by TF env for remote var loading

### LINKED ISSUE #1659: Use TF_ env vars for TFC
When `TF_WORKSPACE`, `TF_CLOUD_ORGANIZATION` or `TF_CLOUD_HOSTNAME` are set, we should use them as fallbacks if `INFRACOST_TERRAFORM_WORKSPACE`, `INFRACOST_TERRAFORM_CLOUD_TOKEN`​ and `INFRACOST_TERRAFORM_CLOUD_HOST​` are not set.

As part of this issue, the precedence order mentioned in [this](https://www.infracost.io/docs/features/environment_variables/) and [this](https://www.infracost.io/docs/features/config_file/#precedence) page should be updated to:
1. CLI flags (run infracost --help to see them)
2. INFRACOST_ environment variables
3. [Config file](https://www.infracost.io/docs/features/config_file/) 
4. TF_ environment variables

hugorut:
fixes #1659

### DIFF
--- a/internal/hcl/remote_variables_loader.go
+++ b/internal/hcl/remote_variables_loader.go
@@ -3,6 +3,7 @@ package hcl
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"sort"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


aliscott (approved):
I think this is good and fixes the original issue so we should go with it.

I'm assuming this doesn't support the case where the user reference the workspace in their HCL? e.g.
```hcl
count = terraform.workspace == "prod" ? 3 : 1
```
If so, I think that's fine, just want to check.

PR #1681, opened 2022-05-23
fix: parse outputs from Terragrunt dependencies

hugorut:
fixes: https://github.com/infracost/infracost/issues/1641

Changes the Terragrunt HCL parser to build `outputs` for each Terragrunt project run. These outputs are then used by Terragrunt projects that make use of `dependency` blocks. How this works:

1. TerragruntHCLParser has the Terragrunt stack output set on it every run. This contains tree information about which Terragrunt projects there are, and which of those have dependencies.
2. `runTerragrunt` method now evaluates the blocks for the project rather than just returning directory information. This allows us to fetch the `output` blocks.
3. `runTerragrunt` uses `ParseConfigFile` and explicitly passes a `cty.Value` of the `outputs` from any dependency projects used. If none are used we have a mock `cty.Value` which bypasses the Terragrunt call to Terraform.
4. I've changed the HCLProvider to expose a `Modules` method which returns the raw HCL blocks instead of the JSON, so that we can use it in the Terragrunt provider

### DIFF
--- a/cmd/infracost/breakdown_test.go
+++ b/cmd/infracost/breakdown_test.go
@@ -227,6 +227,10 @@ func TestBreakdownTerragruntHCLMulti(t *testing.T) {
 	GoldenFileCommandTest(t, testutil.CalcGoldenFileTestdataDirName(), []string{"breakdown", "--path", "../../examples/terragrunt"}, nil)
 }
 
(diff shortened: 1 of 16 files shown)


aliscott (approved):
LGTM

PR #1697, opened 2022-05-31
fix: module inputs incorrectly evaluated

hugorut:
Fixes various issues with nested modules being evaluated in HCL

1. Modules using an input variable from another module output would have occasionally got `nil` values. This happened if outputs were abstracted behind other Terraform blocks (e.g. `locals`). I've now changed this so that the `evaluateModule` step will re-evaluate if the input variables change. This is consistent with the rest of the evaluator functionality, which runs the evaluation process in a loop until the evaluation context remains the same.
2. If users define input variables as strings but they are actually other types (bool, number) then this sometimes causes havoc with expression evaluation. Certain conditional expressions require the same types to be evaluated correctly. I've now changed `variable` evaluation so that if a variable `type` is defined, we try and convert the `cty.Type` to the one specified. This is only needed for simple types
3. On module load the loader was getting the first module that matched the specified `source`. This is problematic if there were two modules using the same source and different versions. I've changed it so that we check the `key` of the module as well as the `source`.

### DIFF
--- a/internal/hcl/attribute.go
+++ b/internal/hcl/attribute.go
@@ -50,6 +50,10 @@ func (attr *Attribute) IsIterable() bool {
 // the Attribute Context. This returns a cty.Value with the values filled from any variables or references
 // that the Context carries.
 func (attr *Attribute) Value() (ctyVal cty.Value) {
+	if attr == nil {
+		return cty.NilVal
+	}
+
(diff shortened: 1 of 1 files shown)


vdmgolub (approved):
Great job! 🤩 💪

Issue #1682, opened 2022-05-24
Skip paths when running breakdown/diff


alikhajeh1:
Before Infracost v0.10.0, we had an `INFRACOST_TERRAGRUNT_FLAGS` environment variable that could be used by Terragrunt users to pass in [Terragrunt CLI flags](https://terragrunt.gruntwork.io/docs/reference/cli-options/#cli-options) such as `--terragrunt-exclude-dir dev`.

As of [v0.10.0](https://www.infracost.io/docs/guides/v0.10_migration/), we parse HCL code by default, meaning there is no dependency on the Terraform or Terragrunt CLIs and thus that env variable is not used. However, users still need a way to  skip directories where Infracost fails on, so Infracost commands can run. One case of this is Terragrunt projects where some modules have issues.

Proposal:
Add a new flag to `infracost breakdown` and `infracost diff`, sub-directories that match the pattern should also be skipped.
```
--skip-path strings   Paths of directories to skip, glob patterns need quotes
```
Example: 
```
infracost breakdown --path . --skip-path "./gke/*/monitoring/wavefront"  --skip-path "./gke/*/applications/rbac"
```

hugorut:
This is now live in the [v0.10.1](https://github.com/infracost/infracost/releases/tag/v0.10.1) release

PR #1702, opened 2022-06-01
feat: add skip-path flag to ignore directories

hugorut:
closes https://github.com/infracost/infracost/issues/1682

Adds a `--skip-path` flag that enables users to ignore set directories. Supports glob and filepatterns.


aliscott:
Should we name this `exclude-path`? Terragrunt and Snyk both use `exclude` and if we ever want the opposite we can have `include-path` in the future.

vdmgolub (approved):

hugorut:
@aliscott sure thing

Issue #1713, opened 2022-06-01
The --exclude-path flag doesn't work with Terraform mono repos


alikhajeh1:
The new `--exclude-path` flag introduced in v0.10.1 works fine with Terragrunt (which is why we added it) but has issues with Terraform mono repos.

`--exclude-path` is a strings array but when multiple paths are passed in, only the last one is taken into account:

```
$ ls my-mono-repo

dev           docker        infracost.yml k8s           modules       prod

$ cd my-mono-repo
$ infracost breakdown --path . --exclude-path=dev --exclude-path=prod

# Output includes dev, which should have been excluded.
```

---
Note for community
Please 👍 if you'd like to upvote this or receive updates.

PR #1714, opened 2022-06-01
fix: issues with multiple exclude-path flags being used

hugorut:
Fixes: https://github.com/infracost/infracost/issues/1713#issue-1256882215

* Solves issue with `--exclude-path` being used with multiple directories. Before we were assigning variables to a function in a loop. This meant that the excluded patterns were always being used as the last value in the loop. I've reworked the logic to build all matches into a single evaluation step.
* Reworks internals to parse all directories before filtering so that we don't show module pricing. Before we were doing the ignored directory check at a lower level. This meant that we weren't properly filtering out modules that were used in skipped directories. We now do it at the top level once all directories have been evaluated.

### DIFF
--- a/internal/hcl/parser.go
+++ b/internal/hcl/parser.go
@@ -170,10 +170,10 @@ type Parser struct {
 // in the given initialPath and returns a Parser for each directory it locates a Terraform project within. If
 // the initialPath contains Terraform files at the top level parsers will be len 1.
 func LoadParsers(initialPath string, excludePaths []string, options ...Option) ([]*Parser, error) {
-	pl := newProjectLocator(excludePaths...)
+	pl := &projectLocator{moduleCalls: make(map[string]struct{}), excludedDirs: excludePaths}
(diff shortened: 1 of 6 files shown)


internal/hcl/parser.go
```diff
@@ -602,14 +628,3 @@ func (p *projectLocator) walkPaths(fullPath string, level int) []string {
 
 	return dirs
 }
-
-func (p *projectLocator) isSkipped(info os.DirEntry) bool {
-	for _, isSkipped := range p.excludedDirs {
-		if isSkipped(info.Name()) {
-			log.Debugf("skipping directory %s as it is marked as exluded by --exclude-path", info.Name())
```
vdmgolub, line 609: Do you want to keep this debug logging on line 536?

vdmgolub (approved):

internal/hcl/parser.go
hugorut: 👍

PR #1720, opened 2022-06-02
fix: Add internal env var for Infracost Dashboard

vdmgolub:


tim775 (approved):

vdmgolub:
@tim775 These two env vars related to Dashboard are for internal use only, we don't list them in the docs.

Issue #1730, opened 2022-06-07
Warning "Failed to load module err"


Joerg-L:
We observe in the logs of infracost fowllowing message

`level=warning parser=terraform_hcl msg="Failed to load module err: missing module with source 'registry.terraform.io/git::git@ssh.dev.azure.com:v3/<<AZURE_DEVOPS_ORG>>/<<AZURE_DEVOPS_PROJECT>>/<<AZURE_DEVOPS_REPO>> git::git@ssh.dev.azure.com:v3/<<AZURE_DEVOPS_ORG>>/<<AZURE_DEVOPS_PROJECT>>/<<AZURE_DEVOPS_REPO>>' -  try to 'terraform init' first"`

We have already run `terraform init` in the step before with follwoing outcome

```
Initializing modules...
Downloading git::ssh://git@ssh.dev.azure.com/v3/<<AZURE_DEVOPS_ORG>>/<<AZURE_DEVOPS_PROJECT>>/<<AZURE_DEVOPS_REPO>> for spoke...
- spoke in .terraform/modules/spoke
Downloading git::ssh://git@ssh.dev.azure.com/v3/<<AZURE_DEVOPS_ORG>>/<<AZURE_DEVOPS_PROJECT>>/<<AZURE_DEVOPS_REPO>> for spoke.peering_spoke_with_hub...
- spoke.peering_spoke_with_hub in .terraform/modules/spoke.peering_spoke_with_hub
Downloading git::ssh://git@ssh.dev.azure.com/v3/<<AZURE_DEVOPS_ORG>>/<<AZURE_DEVOPS_PROJECT>>/<<AZURE_DEVOPS_REPO>> for spoke.snet...
- spoke.snet in .terraform/modules/spoke.snet
Downloading git::ssh://git@ssh.dev.azure.com/v3/<<AZURE_DEVOPS_ORG>>/<<AZURE_DEVOPS_PROJECT>>/<<AZURE_DEVOPS_REPO>> for spoke.vnet...
- spoke.vnet in .terraform/modules/spoke.vnet

Initializing the backend...
```

so the message in the log seems to be wrong.

We are running infracost with following command

```
- script: infracost breakdown --path . --show-skipped --out-file=/tmp/infracost.json --format json   
  displayName: Run Infracost Output to file
  workingDirectory: '$(configroot)'
  env:
    INFRACOST_CURRENCY: EUR
```

aliscott:
[...]

PR #1731, opened 2022-06-07
fix: match modules based on key which should always be unique

aliscott:
Fixes https://github.com/infracost/infracost/issues/1730


hugorut (approved):
LGTM - have we tested this on something like: https://github.com/cloudposse/terraform-aws-eks-workers/tree/master/examples/complete to make sure there's no issue with the key?

aliscott:
What case do you think we should be testing? That repo shows me $0 for both master and this branch.

aliscott:
> > What case do you think we should be testing? That repo shows me $0 for both master and this branch.
> 
> @aliscott can we print out the paths it uses for the modules - make sure there consistent on master and this branch

Done and they show the same, just slightly different order which I think is unrelated/not an issue.

PR #1736, opened 2022-06-09
fix: Allow CORS for callback request from Dashboard UI

vdmgolub:


hugorut (approved):
I think this looks good. However, I'd prefer if we just set the required headers on the response writer rather than using an imported package. It seems pretty trivial.

PR #1740, opened 2022-06-10
chore: hide auth cmd and show register

hugorut:


aliscott (approved):
Looks good. Some of the golden file cmd tests might need updated.

hugorut:
@aliscott ah yeee

PR #1745, opened 2022-06-13
feat: add configure option for disable_version_check

gtirloni:


alikhajeh1:
@gtirloni I'm curious to know why you need this? We already have [an env](https://www.infracost.io/docs/features/environment_variables/#infracost_skip_update_check), which some CI/CD users use. But on the terminal or dev machines, where `infracost configure` is used, we want users to see the upgrade notice as we release new versions regularly.

gtirloni:
Personal preference, mostly. I don't want to be reminded of a new version every 3-5 days. It's practically a permanent notification on my machine.

gtirloni:
Setting `INFRACOST_SKIP_UPDATE_CHECK` satisfies my use case. Closing.

alikhajeh1:
@gtirloni that makes sense, so it’s like spam just now :)
We have a story to make all env vars supported in `infracost configure` so if you still want this, maybe tweak the PR to make it work with the env var, so users can do this: `infracost configure set skip_update_check true`. Not sure if @aliscott has any other thoughts on this - maybe reducing the frequency of how often we check for updates

alikhajeh1:
Cool - @gtirloni up to you if you want to re-open the PR and to make this `infracost configure set skip_update_check true` :)

alikhajeh1:
> Personal preference, mostly. I don't want to be reminded of a new version every 3-5 days. It's practically a permanent notification on my machine.

@gtirloni what do you think is a reasonable time frame for the CLI to check for updates? Once every 2 weeks? (that's usually how often we release new versions but sometimes we do it more for bug fixes)

PR #1750, opened 2022-06-14
chore: Fix linter issues

vdmgolub:


internal/hcl/funcs/marks.go
```diff
@@ -12,7 +12,7 @@ import (
 type valueMark string
 
 func (m valueMark) GoString() string {
-	return "marks." + strings.Title(string(m))
+	return "marks." + cases.Title(language.English).String(string(m))
```
vdmgolub, line 15: > SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punct
uation properly. Use golang.org/x/text/cases instead. (staticcheck)

internal/hcl/block.go
```diff
@@ -110,15 +110,11 @@ func (blocks Blocks) ModuleBlocks() Blocks {
...
-		toSort[i] = block
-	}
+	copy(toSort, justModules)
 
 	sort.Sort(toSort)
 
-	for i, block := range toSort {
-		justModules[i] = block
-	}
+	copy(justModules, toSort)
```
vdmgolub, line 117: > S1001: should use copy() instead of a loop (gosimple)

internal/providers/terraform/tftest/tftest.go
```diff
@@ -410,7 +410,7 @@ func createTerraformProject(t *testing.T, tfProject TerraformProject) string {
 		err := copyInitCacheToPath(initCache, tmpDir)
 		require.NoError(t, err)
 	} else {
-		t.Log(fmt.Sprintf("Couldn't copy terraform init cache from %s", initCache))
+		t.Logf("Couldn't copy terraform init cache from %s", initCache)
```
vdmgolub, line 413: > S1038: should use t.Logf(...) instead of t.Log(fmt.Sprintf(...)) (gosimple)

hugorut (approved):

PR #1755, opened 2022-06-15
feat: Don't run docker process as root

pedrocarrico:
### Objective

This PR makes the docker process run as a non root user.

### Background

Some state and cache files are created when running the infracost docker process and I noticed that on my CI/CD server those files became owned by root causing some issues when trying to clean them up afterwards on subsequent pipeline runs.

As best practice we try to run docker processes with a non-privileged user in order to prevent any security issues and I took this opportunity to tackle it on infracost.

### Running

There is a slight difference when running the docker process with these changes and I didn't find where I could document it.
Previous to these changes we did:
```
docker run -e INFRACOST_API_KEY -v "$PWD":/root --rm infracost/infracost:ci-<version> breakdown --path . --format json --out-file infracost.json
```

With these changes:
```
docker run -e INFRACOST_API_KEY -v "$PWD":/infracost --rm infracost/infracost:ci-<version> breakdown --path . --format json --out-file infracost.json
```

Notice how we bind mount the code directory, we bind mount the code directory to `/infracost` instead of `/root`.

Let me know your thoughts and if there's anything else I could help with.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,4 +65,12 @@ COPY --from=builder /usr/bin/terraform* /usr/bin/
 COPY --from=builder /usr/bin/terragrunt /usr/bin/
 COPY --from=builder /app/build/infracost /usr/bin/
 
+RUN addgroup --gid 1000 infracost; \
+  adduser infracost --uid 1000 -G infracost -H -D -s /sbin/nologin; \
+  mkdir -p /infracost; \
+  chown infracost:infracost /infracost
+
+USER infracost
+WORKDIR /infracost
+
 ENTRYPOINT [ "infracost" ]
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -43,4 +43,12 @@ ENV INFRACOST_CI_IMAGE=true
 ENV INFRACOST_SKIP_UPDATE_CHECK='true'
 ENV INFRACOST_LOG_LEVEL=info
 
+RUN addgroup --gid 1000 infracost; \
+  adduser infracost --uid 1000 -G infracost -H -D -s /sbin/nologin; \
+  mkdir -p /infracost; \
+  chown infracost:infracost /infracost
+
+USER infracost
+WORKDIR /infracost
+
 ENTRYPOINT ["infracost"]


pedrocarrico:
@aliscott I don't think it will be a breaking change, but existing users must bind mount to another directory as we cannot (should not) use the previous `/root` directory anymore.

aliscott:
> @aliscott I don't think it will be a breaking change

From what I understand from this if an existing user is currently running this in their CI pipeline:
```
docker run -e INFRACOST_API_KEY -v "$PWD":/root --rm infracost/infracost:ci-latest breakdown --path /root --format json --out-file infracost.json
```

When we release this change they will start seeing the following until they update to use the `/infracost` path:

```
Error: No valid Terraform files found at the given path, try a different directory
```

However, if they are running as per our docker example in the [Getting started guide](https://www.infracost.io/docs/). It will continue to work since they are pointing `--path` to an absolute path.

I think this is a great change, I just want to make sure we release it with minimal affect to existing users - people really don't like when we break their CI pipelines 😓.

cc @alikhajeh1 if he has opinions. One option is to start a v0.11 branch and put this in there so it gets released as part of v0.11, where we will have a migration guide for users.

aliscott:
> I'm probably missing something but I can't find where in our docs or CI/CD integration repos where we do "$PWD":/root.

Correct, we don't have that in our Getting Started guide. For anyone who followed that when setting up their CI workflow it will still work. The problem is if people didn't follow that and are using `"$PWD":/root` then their CI workflows will break.

alikhajeh1:
Thanks @aliscott - in that case either v0.11 seems safest so there is explicit action needed by users. ppl who use `latest` are fine living on the edge and getting cut by it 🤣

pedrocarrico:
@aliscott @alikhajeh1 I might have missed that `-v $PWD/:/code/` in the Get Started when I setup my pipelines and went straight to the Dockerfile code to see what `WORKDIR` the docker process was using.

> Thanks @aliscott - in that case either v0.11 seems safest so there is explicit action needed by users. ppl who use latest are fine living on the edge and getting cut by it 🤣

Totally agree!

aliscott:
I've created a v0.11 branch for anything we want to release into v0.11 and updated the base of this PR so we can merge it into that.

aliscott (commented):
Should we use gid and uid `1001`? I'm not familiar with the best practice for this with Docker, but when I looked into this for the [cloud-pricing-api](https://github.com/infracost/cloud-pricing-api) we chose `1001` because `1000` is the first non-system user which can sometimes be an administrator user.

pedrocarrico:
@aliscott The common practice I've seen being done in Dockerfiles is to run it as the first gid and uid available on the image.

In the case of [cloud-pricing-api](https://github.com/infracost/cloud-pricing-api) I see that you guys are using the node alpine as a base image and as such the gid and uid `1000` was already taken for the `node` user.

aliscott:
> @aliscott The common practice I've seen being done in Dockerfiles is to run it as the first gid and uid available on the image.

👍 makes sense thank you!

aliscott (approved):

PR #1762, opened 2022-06-20
fix: git metadata cleanup after demo

hugorut:
* commitTimestamp to readable string
* remove trailing lines from commit message
* use GitLab pipeline id over job id


internal/vcs/metadata.go
```diff
@@ -202,8 +203,8 @@ func commitToMetadata(commit *object.Commit) Commit {
 		SHA:         commit.Hash.String(),
 		AuthorName:  commit.Author.Name,
 		AuthorEmail: commit.Author.Email,
-		Timestamp:   commit.Author.When.Unix(),
-		Message:     commit.Message,
+		Time:        commit.Author.When,
+		Message:     strings.TrimRight(commit.Message, "\n"),
```
vdmgolub, line 207: We probably should remove trailing spaces too. The message can be multiline, so something like

```git
feat: Add test feature

This test feature is awesome!      


```

could be nicely formatted.

vdmgolub (approved):
👍

internal/vcs/metadata.go
hugorut: updated to  `strings.TrimRight(commit.Message, "\n\r ")` https://go.dev/play/p/wjsUkT2REzz

Issue #1728, opened 2022-06-07
get_env() doesn't work when used with Infracost


prajjwaldimri:
The terragrunt.hcl file in this case contains an after_hook which uses `get_env()` to retrieve environment variables. 
`terragrunt plan` is able to access the environment variables however when I run `infracost breakdown` it gives me the following error:

```
Error in function call; Call to function "get_env" failed: EnvVarNotFound: Required environment variable XXXXXXXX - not found., and 1 other diagnostic(s)
```

aliscott:
Thanks for reporting @prajjwaldimri 🙏. As a workaround just now you can you add the `--terraform-force-cli` flag to any `infracost breakdown` and `infracost diff` commands you run. This flag triggers the old behavior where HCL-parsing was not used, and instead the `terragrunt` CLI was invoked internally.

See https://github.com/infracost/infracost/issues/1695 for more information. We will investigate this more so we can get a fix out in a patch version.

hugorut:
@prajjwaldimri a fix for this is in master and will be released later this week in `v0.10.4`. I'll let you know when it's live.

hugorut:
Hey @prajjwaldimri `get_env` support is now live in the [`v0.10.5`](https://github.com/infracost/infracost/releases) version of Infracost. Thanks for bringing this to our attention.

juanr2001:
@prajjwaldimri my lord, it's still happening, can you help me:

I am still getting the error get I do

```
export TF_VAR_env="git"
terraform init
```

```
# terragrunt.hcl

locals {
  test = get_env("env") # Error:  Required environment variable env - not found.
}
```

hugorut:
hi @juanr2001, are you getting that error when you run `infracost`? From your example you seem just to be running `terraform`?

juanr2001:
@hugorut 
It was my fault, I had to reference it like this, and it worked:
```
get_env("TF_VAR_env")
```

PR #1764, opened 2022-06-21
fix: set environment vars on Terragrunt run options

hugorut:
Fixes https://github.com/infracost/infracost/issues/1728

Inits `TerragruntHCLProvider` with sanitised env vars and sets them on the `TerragruntOptions` so they are available for the `get_env` function.


cmd/infracost/breakdown_test.go
```diff
@@ -261,6 +262,17 @@ func TestBreakdownTerragruntHCLDepsOutput(t *testing.T) {
 	GoldenFileCommandTest(t, testutil.CalcGoldenFileTestdataDirName(), []string{"breakdown", "--path", path.Join("./testdata", testutil.CalcGoldenFileTestdataDirName())}, nil)
 }
 
+func TestBreakdownTerragruntGetEnv(t *testing.T) {
+	os.Setenv("CUSTOM_OS_VAR", "test")
```
tim775, line 266: You can also set these as test options in the last param to `GoldenFileCommandTest`:
```
&GoldenFileOptions{
    Env: map[string]string{
        "CUSTOM_OS_VAR": "test",
    },
}
		```

tim775 (approved):

Issue #1667, opened 2022-05-18
HCL parsing does not work with modules that have a `source` in a private Terraform registry


aliscott:
HCL parsing does not work if the Terraform code contains a module with a source in a private Terraform registry. Private git modules work ok.

Example this code:
```hcl
provider "aws" {
  region                      = "us-east-1"
  skip_credentials_validation = true
  skip_requesting_account_id  = true
  access_key                  = "mock_access_key"
  secret_key                  = "mock_secret_key"
}

module "ec2_cluster" {
  source  = "app.terraform.io/infracost/ec2-instance/aws"
  version = "~> 2.0"

  name           = "my-cluster"
  instance_count = 5

  ami                    = "ami-ebd02392"
  instance_type          = "t2.micro"
  key_name               = "user1"
  monitoring             = true
  vpc_security_group_ids = ["sg-12345678"]
  subnet_id              = "subnet-eddcdzz4"
}
```

Shows this error:
```
Error: Error loading Terraform modules: error downloading 'file:///examples/terraform/.infracost/terraform_modules/ec2_cluster/app.terraform.io/infracost/ec2-instance/aws': source path error: stat /examples/terraform/.infracost/terraform_modules/ec2_cluster/app.terraform.io/infracost/ec2-instance/aws: no such file or directory
```

There's two issues which are causing this:
1. Our code assumes that the module API is always at `<host>/v1/modules` but this can be different for different registries. We should use the [terraform-svchost](https://github.com/hashicorp/terraform-svchost/tree/master/disco) package to discover the correct service URLs
2. Our code does not pass through any credentials when pulling remote registry modules.

I've created a [work in progress branch here](https://github.com/infracost/infracost/compare/fix/hcl-private-registry-modules?expand=1).

EppO:
if you want feedback about this feature, I can test your changes with a GitLab Terraform registry which uses "<host>/terraform/<project name>/azurerm" module address

alikhajeh1:
@EppO thanks! Just to double check, you're using a private TF module hosted on [GitLab Terraform Registry](https://docs.gitlab.com/ee/user/packages/terraform_module_registry/)? I've not used that before, does GitLab support public modules too? Or would those just be public git repos?

EppO:
I guess that depends if your repo is public or private. I would assume it supports both

EppO:
[...]

PR #1781, opened 2022-06-27
fix: resolve private Terraform registry modules

hugorut:
Closes https://github.com/infracost/infracost/issues/1667

* Uses https://github.com/hashicorp/terraform-svchost discovery mechanism to try and resolve the real location for the module
* Adds credential support based off of `disco` mechanism
* Adds private module tests and reworks CI pipeline to include Terraform Cloud tokens for test

### DIFF
--- a/internal/providers/terraform/cloud.go
+++ b/internal/providers/terraform/cloud.go
@@ -1,24 +1,15 @@
 package terraform
 
 import (
-	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
-	"os"
-	"path"
-	"path/filepath"
-	"runtime"
 
-	"github.com/hashicorp/hcl2/gohcl"
-	"github.com/hashicorp/hcl2/hclparse"
-	"github.com/mitchellh/go-homedir"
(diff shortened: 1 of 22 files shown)


hugorut:
@aliscott just an update from this morning. I tested this branch with @jgrumboe and it worked great for their setup 🎉

internal/providers/terraform/cloud.go
```diff
@@ -38,135 +29,10 @@ func cloudAPI(host string, path string, token string) ([]byte, error) {
 	defer resp.Body.Close()
 
 	if resp.StatusCode == 401 {
-		return []byte{}, ErrInvalidCloudToken
+		return []byte{}, modules.ErrInvalidCloudToken
```
aliscott, line 32: Seems odd we are referencing the `hcl/modules` package from the `cloud` and `dir_provider`. I assume the errors and functions were moved there to avoid circular dependencies? Not sure if it's worth introducing a separate package for these.

aliscott (approved):
LGTM

PR #1782, opened 2022-06-27
fix: introduce keyed mutex to wait for clone to finish before access

hugorut:
Changes `vcs` package to include a keyed mutex so that go routines wait for cloned repos to be available. Fixes issue where commit ref is not available in `config-file` project run.


aliscott (approved):
LGTM

PR #1785, opened 2022-06-28
enhance: Improve comment error output when the provider fails

tim775:


hugorut (approved):

internal/comment/handler.go
```diff
@@ -258,9 +259,21 @@ func (h *CommentHandler) deleteComments(ctx context.Context, comments []Comment)
...
+// newPlatformError wraps a platform error with multi-line formatting and a link to the docs
+func (h *CommentHandler) newPlatformError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	return errors.Errorf("%s\n%v\n\n%s",
+		"The pull request comment was generated successfully but could not be posted:",
+		err,
+		"See https://infracost.io/docs/troubleshooting/#5-posting-comments for help.")
```
hugorut, line 278: could we use `fmt.Errorf` instead, please?

PR #1787, opened 2022-06-29
fix: loading remote variables

### LINKED ISSUE #1786: Remote variables should also be loaded for `agent` execution mode
Currently, infracost skips the remote variable loading if the execution mode is `remote`.
https://github.com/infracost/infracost/blob/e1db77e3a61e605e4a49974460a8432fe42cd264/internal/hcl/remote_variables_loader.go#L147

In our case we use the `agent` execution mode, which is more or less the same as `remote` but on private runners.
So, I would like to see infracost fetching the remote variables of the TFC workspace in that case.

jgrumboe:
Closes #1786

Remote variables should also be fetched if the TFC workspace is set to `agent` execution mode.
I thought it's just better to flip the if-statement and check if `local` execution mode is set and then skip fetching the remote variables.

### DIFF
--- a/internal/hcl/remote_variables_loader.go
+++ b/internal/hcl/remote_variables_loader.go
@@ -144,8 +144,8 @@ func (r *RemoteVariablesLoader) Load(blocks Blocks) (map[string]cty.Value, error
 		return vars, nil
 	}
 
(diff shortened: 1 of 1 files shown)


hugorut (approved):
seems fine to me

PR #1794, opened 2022-06-30
fix: AWS DMS resource throwing nil pointer if `replication_instance_c…

aliscott:
…lass` is in the wrong format


hugorut (approved):
good catch

PR #1808, opened 2022-07-03
feat: structured logging

hugorut:
A WIP branch with a first pass at refactoring some logs to be structured. This is branch is designed to facilitate upcoming discussions on structured logging.

## Changes

1. Changes the base log output to be `JSON`
2. Adds contextual file information to all log output
3. Passes logger down the stack, appending useful metadata to help debug
4. Adds a `debug-report` flag which can generate a log output file which is formatted correctly as a pretty printed `JSON` array. Making it easier for users to generate and upload to issues. Also means it's easier for us to search log reports, using something like jq, e.g:
```
jq '.[] | select(.block_name == "aws_instance.web_app")' infracost-debug-report.json
```

All of the above changes mean that a log line like this:

```go
log.Debug("fetching attribute value")
```

goes from:

```
time="2022-07-03T09:50:18+01:00" level=debug parser=terraform_hcl msg="fetching attribute value"
```

to

```json
{
  "attribute_name": "ami",
  "block_name": "aws_instance.web_app",
  "file": "/Users/hugorut/code/infracost/infracost/internal/hcl/attribute.go:67",
  "func": "github.com/infracost/infracost/internal/hcl.(*Attribute).Value",
  "level": "debug",
  "msg": "fetching attribute value",
  "parent_provider": "terragrunt_dir",
  "project_name": "",
  "project_path": "examples/terragrunt/.infracost/.terragrunt-cache/swegQ5ifuM3-jiDXRFa8bvohfsM/-7-kNid-o4yAQ5mYeGWKMRBIGIA/modules/example",
  "provider": "terraform_dir",
  "routine": 16,
  "time": "2022-07-03T09:43:01+01:00"
},
```

---

Note I've chosen to implement the structured logging with the `logrus` package we already use, as it's the quickest way to success for this POC. However, [logrus](https://github.com/sirupsen/logrus) is in maintenance mode, and I would strongly suggest that in the future migrate to something like [zap](https://github.com/uber-go/zap), which is not only faster but actively maintained.


aliscott:
> I would strongly suggest that in the future migrate to something like [zap](https://github.com/uber-go/zap), which is not only faster but actively maintained

Yeah +1 on this. When I tried [refactoring this a while ago](https://github.com/infracost/infracost/tree/refactor/logging) before I used zerolog (I think because I preferred its API).

tim775 (approved):

aliscott (approved):
BOOOOM!

Issue #1556, opened 2022-04-11
Calculate cost for on-demand instances which are not always on


DeltaWhy:
My company uses a lot of EC2 instances which only run during business hours and are stopped otherwise. Would it be possible to add an 'hours per month' or 'percent usage per month' field to the usage file to account for this use case?

For instance, we might estimate costs based on 180 hours per month rather than 730 for the instances, but the disk subresources are still charged at the full monthly rate (since we're stopping rather than terminating the instances).

alikhajeh1:
@DeltaWhy great idea - I've heard this a few times from users so thanks for creating the issue! If you'd like to [contribute](https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md), this seems like a good first issue:
1. Add `monthly_hours` to the [example usage file](https://github.com/infracost/infracost/blob/master/infracost-usage-example.yml#L202) (we have this for some of the other resources too).
2. Add the above param to [this](https://github.com/infracost/infracost/blob/master/internal/resources/aws/instance.go#L18)
3. change [this](https://github.com/infracost/infracost/blob/master/internal/resources/aws/instance.go#L172) to `MonthlyQuantity` and use the param. The param should default to 730 so the existing behavior continues to work if users don't define the usage param (I suppose we should also validate the param is >0 and <= 730).
4. Review the other `HourlyQuantity` mentions in the instance.go file above and see if they should use the same param or, if they should remain as 730 (e.g. do users pay for the Inference accelerator when the instance is stopped?) - if needed we can also help with this by reviewing the AWS EC2 pricing page.

Semi-related to the above change could be to also update the Azure and Google instance resources so instances are consistent across the providers. 

Unrelated to this GH issue, this feature'll enable us to get feedback from AWS RDS and other resource users to see if the same thing is needed for those resources.

alikhajeh1:
@hugorut plz close this after v0.10.8 is live as it looks like it's being released as part of that.

PR #1812, opened 2022-07-04
feat: add monthly_hrs usage field for aws/azure/gcp VMs

jessecureton:
Updates the following resource types to support a `monthly_hrs` usage flag:

- `aws_instance`
- `azurerm_virtual_machine`
- `azurerm_linux_virtual_machine`
- `azurerm_windows_virtual_machine`
- `google_compute_instance`

This addresses #1556.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,12 @@ test_cmd:
 test_update_cmd:
 	INFRACOST_LOG_LEVEL=warn go test -timeout 30m $(LD_FLAGS) ./cmd/infracost $(or $(ARGS), -update -v -cover)
 
+test_usage:
+	INFRACOST_LOG_LEVEL=warn go test -timeout 30m $(LD_FLAGS) ./internal/usage $(or $(ARGS), -v -cover)
+
+test_update_usage:
+	INFRACOST_LOG_LEVEL=warn go test -timeout 30m $(LD_FLAGS) ./internal/usage $(or $(ARGS), -update -v -cover)
+
 # Run AWS resource tests
 test_aws:
 	INFRACOST_LOG_LEVEL=warn go test -timeout 30m $(LD_FLAGS) ./internal/providers/terraform/aws $(or $(ARGS), -v -cover)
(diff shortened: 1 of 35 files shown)


jessecureton:
At its core this is a pretty straightforward PR, but the myriad golden file updates have made it appear larger than it is.

Happy to break this up into a separate PR each for AWS / Azure / GCP resources, if the maintainers prefer. Looking forward to getting feedback and getting this landed!

aliscott:
Thanks for the contribution @jessecureton 🙏! One of our team will look at this in the next couple of days.

jessecureton:
Sounds good! I took a look at the failing test case in the CI checks and it seems unrelated to this change. I did notice the failing test case (`TestBreakdownWithPrivateTerraformRegistryModule`) does not run locally, and instead prints the following message.

```
breakdown_test.go:369: Skipping because INFRACOST_TERRAFORM_CLOUD_TOKEN is not set and external contributors won't have this.
```

Locally running `make test_all` passes all tests as expected, so hazarding a guess this seems like it might be something related to CI secrets when running in my forked repo?

aliscott:
@jessecureton you're correct. I pushed a fix to skip this test in https://github.com/infracost/infracost/pull/1809, but I think there might still be issues here I haven't had a chance to investigate today.

jessecureton:
>     1. I think we should we use `monthly_hrs` to be consistent with the other `*_hrs` usage keys.

I used `monthly_hours` to match the same field name already in use in several other resources - namely `azurerm_synapse_spark_pool`, `aws_glue_job`, and `aws_glue_crawler`. That said, I'm happy to make this change if you still prefer the `_hrs` syntax, just wanted to point out the mixed convention already in use.

>     2. Should it be a float value for the use-case if a user only runs the instance for a few minutes in the month?

This seems reasonable! I will make this change alongside the above once I hear back re: the preferred syntax

jessecureton:
I've updated this to include both the requested changes for float values and the switch to monthly_hrs! Thanks for the link to the policy doc for naming, I had missed that initially!

aliscott (approved):
Thanks @jessecureton, this looks great! I'll rebase and run the tests, to make sure they all pass.

PR #1826, opened 2022-07-06
enhance: display correct error if module detected as registry module …

aliscott:
…and fails

Before we were falling back to treating the module as a remote module if there were any issues looking it up from the registry. This changes that to differentiate between if it's not a registry module or it is a registry module but there was an error looking it up.

Before this change we show the following error if a private module fails with a 401:
```
Error: Error loading Terraform modules: Failed to download remote module ec2_cluster: error downloading 'file:///testdata/breakdown_with_private_terraform_registry_module/.infracost/terraform_modules/ec2_cluster/app.terraform.io/infracost/ec2-instance/aws': source path error: stat /testdata/breakdown_with_private_terraform_registry_module/.infracost/terraform_modules/ec2_cluster/app.terraform.io/infracost/ec2-instance/aws: no such file or directory
```

After this change, we show the actual error:
```
Error: Error loading Terraform modules: error looking up registry module ec2_cluster: Module versions endpoint returned status code 401
```

### DIFF
--- a/internal/hcl/modules/cache.go
+++ b/internal/hcl/modules/cache.go
@@ -67,7 +67,7 @@ func (c *Cache) lookupModule(key string, moduleCall *tfconfig.ModuleCall) (*Mani
 		return checkVersion(moduleCall, manifestModule)
 	}
 
-	url, err := c.disco.ModuleLocation(moduleCall.Source)
+	url, _, err := c.disco.ModuleLocation(moduleCall.Source)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


aliscott:
I'm not able to get any further with the dependabot issues without this at the moment - adding this commit to them or rebasing them against the branch doesn't help.

@hugorut there's probably a better way to structure this so open to suggestions - I started off with just trying not to change the existing structure too much.

hugorut (approved):

Issue #1773, opened 2022-06-22
Support JSON files as Terraform var-file


pacorreia:
Team,

Please consider supporting JSON files as Terraform input files.

Terraform already supports that as per documenation [here](https://www.terraform.io/language/values/variables#variable-definitions-tfvars-files)

This would allow me to start using Infracost on top of source code instead of relying on Terraform JSON plans

alikhajeh1:
@pacorreia great! @hugorut and @aliscott just FYI on this, I don't think it requires a CLI interface change so hopefully it's a fairly straightforward change?

pacorreia:
count me on to help test it 👌

pacorreia:
Just want to share that found this tool written in GO that can do HCL  to JSON a vice-versa, thought it might help:

https://github.com/gorpher/hcl2json

hugorut:
hey @pacorreia, this should be a simple fix we just don't have it on the roadmap atm. Let me bring it up at our grooming meeting on Thursday and hopefully we can get it into the next release.

pacorreia:
That would be very cool indeed 😊

A terça, 5/07/2022, 17:47, Hugo Rut ***@***.***> escreveu:

hugorut:
@pacorreia this has now been fixed and is merged into master awaiting the next release. I'll ping here to let you know when it's live.

pacorreia:
@hugorut will be glad to test asap and give a shout out 😊

pacorreia:
Just saw it, already testing it, great work not only with this, but with other features!

PR #1831, opened 2022-07-07
fix: support input JSON input variable files

hugorut:
closes https://github.com/infracost/infracost/issues/1773


trueface00:
- []()

Issue #1835, opened 2022-07-08
Panic in Azure CosmoDB resource


aliscott:
Stack trace:
```
"runtime error: index out of range [0] with length 0
goroutine 65 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.18.3/x64/src/runtime/debug/stack.go:24 +0x65
main.(*parallelRunner).run.func1.1()
	/home/runner/work/infracost/infracost/cmd/infracost/run.go:304 +0x46
panic({0x1f9d580, 0xc0015ad218})
	/opt/hostedtoolcache/go/1.18.3/x64/src/runtime/panic.go:838 +0x207
github.com/infracost/infracost/internal/providers/terraform/azure.cosmosDBCostComponents(0xc000bbdef0, 0xc0019d7b88?, 0xc000bbdd90)
	/home/runner/work/infracost/infracost/internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go:68 +0x14b7
github.com/infracost/infracost/internal/providers/terraform/azure.NewAzureRMCosmosdb(0xc000bbdef0, 0xc001100b70?)
	/home/runner/work/infracost/infracost/internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go:38 +0x376
github.com/infracost/infracost/internal/providers/terraform.(*Parser).createResource(0xc000c5b920, 0xc000bbdef0, 0x0)
	/home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:57 +0x12b
github.com/infracost/infracost/internal/providers/terraform.(*Parser).parseJSONResources(_, _, {_, _, _}, _, {0x5, {0xc002f42000, 0x69a47b}, {0x0, ...}, ...}, ...)
	/home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:105 +0x5ef
github.com/infracost/infracost/internal/providers/terraform.(*Parser).parseJSON(0xc000c5b920, {0xc0037a2000, 0x69a47b, 0x82c581}, 0xc002012000?)
	/home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:145 +0x648
github.com/infracost/infracost/internal/providers/terraform.(*HCLProvider).parseResources(0xc000688000, {0x7fff817b9a9e?, 0x30?}, {0xc0037a2000, 0x69a47b, 0x82c581}, 0x7fc2fc35f208?)
[...]

PR #1838, opened 2022-07-08
fix: blank geo_locations causing panic errors

hugorut:
closes https://github.com/infracost/infracost/issues/1835

### DIFF
--- a/internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go
+++ b/internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go
@@ -47,6 +47,22 @@ func cosmosDBCostComponents(d *schema.ResourceData, u *schema.UsageData, account
 	region := lookupRegion(account, []string{"account_name", "resource_group_name"})
 	geoLocations := account.Get("geo_location").Array()
 
+	// The geo_location attribute is a required attribute however it can be an empty list because of
+	// expressions evaluating as nil, e.g. using a data block. If the geoLocations variable is empty
+	// we set it as a sane default which is using the location from the parent region.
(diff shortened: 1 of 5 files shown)


internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go
```diff
@@ -47,6 +47,22 @@ func cosmosDBCostComponents(d *schema.ResourceData, u *schema.UsageData, account
...
 
+	// The geo_location attribute is a required attribute however it can be an empty list because of
+	// expressions evaluating as nil, e.g. using a data block. If the geoLocations variable is empty
+	// we set it as a sane default which is using the location from the parent region.
+	if len(geoLocations) == 0 {
+		log.WithFields(log.Fields{
+			"resource": d.Address,
+		}).Debugf("empty set of geo_location attributes provided using fallback region %s", region)
+
+		geoLocations = []gjson.Result{
```
aliscott, line 58: Seems odd we're creating a gjson, but I guess it will involve more refactoring at the moment to pass the location directly to the functions we are calling here?

aliscott (approved):

internal/providers/terraform/azure/cosmosdb_cassandra_keyspace.go
hugorut: yeah the `geo_locations` are used by a ton of downstream functions that expect an array of `gjson.Result` it's jankey, but I didn't think it warranted a ton of refactoring

aliscott: 👍

PR #1840, opened 2022-07-09
deps: upgrade to use go1.18 and alpine3.16 and github action dependen…

chenrui333:
…cies

For github action dependency updates:
These github action upgrades are mainly relates to the nodejsv16 runtime update.

The upstream dependency notes would generally go like this:
> Node 12 has an end of life on April 30, 2022.
> This PR updates the default runtime to node16, rather then node12.
> This is supported on all Actions Runners v2.285.0 or later.

go@1.18 is already used to build [homebrew formula](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/infracost.rb#L18).

Signed-off-by: Rui Chen <rui@chenrui.dev>


aliscott:
@chenrui333 this looks good. Any chance you could rebase this against master and repush? - that should fix the build failures

aliscott:
@chenrui333 I've merged master into this to see if it fixes the build issues.

aliscott (approved):

PR #1850, opened 2022-07-11
feat(Azure): Add IoT Hub resources

### LINKED ISSUE #1616: Add support for azure iot hub devices
Add estimation support for `azurerm_iothub` resources, e.g. from here: https://azure.microsoft.com/en-us/pricing/details/iot-hub/

Each SKU has a defined limit of messages per day per hub unit - so estimating this resource based solely on hub unit capacity should be sufficient - this is what the Azure pricing calculator does.

---

### Cost details

Paid resources:

- `azurerm_iothub`

  | Name | Units | Usage param | Notes |
  | ---- | ----- | ----------- | ----- |
  | `IoT hub` | `units` |  | `MonthlyQuantity`, the price is multiplied by the number of units from `sku.capacity`. `sku.name` defines the tier. If value is `F1` the resource is free |

- `azurerm_iothub_dps`

  | Name | Units | Usage param | Notes |
  | ---- | ----- | ----------- | ----- |
  | `Device provisioning` | `1k operations` | `monthly_operations`  | `MonthlyQuantity`, the price is multiplied by the number of units from `sku.capacity`. NB: the CPAPI price is for 1k operations.  |

CPAPI query fo reference: `select * from products where "vendorName" = 'azure' and service = 'IoT Hub' and region = 'westus'`. `S1 Operations` meterName defines price for `azurerm_iothub_dps` cost component.

Free resources:

- `azurerm_iothub_certificate`
- `azurerm_iothub_consumer_group`
- `azurerm_iothub_dps_certificate`
- `azurerm_iothub_dps_shared_access_policy`
- `azurerm_iothub_shared_access_policy`

---

### Note for community
Please 👍 if you'd like to upvote this or receive updates.
Add a comment if you'd like to [contribute](https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md).

sawyerw-mastery:
## Objective:

Add support for Azure IoT Hub resources. Fixes #1616.

## Pricing details:

https://azure.microsoft.com/en-us/services/iot-hub/
https://azure.microsoft.com/en-us/pricing/details/iot-hub/

## Status:

- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/pull/253)

## Issues:

None

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -964,6 +964,9 @@ resource_usage:
   azurerm_firewall.my_firewall:
     monthly_data_processed_gb: 100000 # Monthly data processed by the firewall in GB.
 
+  azurerm_iothub_dps.my_iothub_dps:
+    monthly_operations: 10000
+
   azurerm_linux_virtual_machine.my_linux_vm:
     os_disk:
       monthly_disk_operations: 2000000 # Number of disk operations (writes, reads, deletes) using a unit size of 256KiB.
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -965,7 +965,7 @@ resource_usage:
     monthly_data_processed_gb: 100000 # Monthly data processed by the firewall in GB.
 
   azurerm_iothub_dps.my_iothub_dps:
-    monthly_operations: 10000
+    monthly_operations: 10000 # Monthly number of device provisioning operations
 
   azurerm_linux_virtual_machine.my_linux_vm:
     os_disk:
--- a/internal/providers/terraform/azure/iothub.go
+++ b/internal/providers/terraform/azure/iothub.go
@@ -1,0 +1,38 @@
+package azure
+
+import (
+	"github.com/infracost/infracost/internal/resources/azure"
+	"github.com/infracost/infracost/internal/schema"
+)
+
+func getIoTHubRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_iothub",
+		RFunc: newIoTHub,
+	}
+}
+
+func getIoTHubDPSRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:  "azurerm_iothub_dps",
+		RFunc: newIoTHub,
+	}
+}
+
+func newIoTHub(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := lookupRegion(d, []string{})
+
+	sku := d.Get("sku.0.name").String()
+	count := d.Get("sku.0.capacity").Int()
+
+	r := &azure.IoTHub{
+		Address:  d.Address,
+		Region:   region,
+		Sku:      sku,
+		Capacity: count,
+	}
+
+	r.PopulateUsage(u)
+
+	return r.BuildResource()
+}
--- a/internal/providers/terraform/azure/iothub.go
+++ b/internal/providers/terraform/azure/iothub.go
@@ -3,8 +3,6 @@ package azure
 import (
 	"github.com/infracost/infracost/internal/resources/azure"
 	"github.com/infracost/infracost/internal/schema"
-	log "github.com/sirupsen/logrus"
-	"github.com/tidwall/gjson"
 )
 
 func getIoTHubRegistryItem() *schema.RegistryItem {
@@ -32,7 +30,6 @@ func newIoTHub(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
 		Region:   region,
 		Sku:      sku,
 		Capacity: capacity,
-		DPS:      false,
 	}
 
 	r.PopulateUsage(u)
@@ -44,22 +41,14 @@ func newIoTHubDPS(d *schema.ResourceData, u *schema.UsageData) *schema.Resource 
 	region := lookupRegion(d, []string{})
 
 	sku := d.Get("sku.0.name").String()
-	capacity := d.Get("sku.0.capacity").Int()
 
-	if u != nil && u.Get("monthly_operations").Type != gjson.Null {
-		r := &azure.IoTHub{
-			Address:  d.Address,
-			Region:   region,
-			Sku:      sku,
-			Capacity: capacity,
-			DPS:      true,
-		}
-
-		r.PopulateUsage(u)
-
-		return r.BuildResource()
+	r := &azure.IoTHubDPS{
+		Address: d.Address,
+		Region:  region,
+		Sku:     sku,
 	}
 
-	log.Warnf("Skipping resource %s. Could not find a way to get its cost components from the resource or usage file.", d.Address)
-	return nil
+	r.PopulateUsage(u)
+
+	return r.BuildResource()
 }
--- a/internal/providers/terraform/azure/iothub.go
+++ b/internal/providers/terraform/azure/iothub.go
@@ -3,6 +3,8 @@ package azure
 import (
 	"github.com/infracost/infracost/internal/resources/azure"
 	"github.com/infracost/infracost/internal/schema"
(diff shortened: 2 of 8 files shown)


internal/providers/terraform/azure/iothub.go
```diff
@@ -0,0 +1,65 @@
...
+	return r.BuildResource()
+}
+
+func newIoTHubDPS(d *schema.ResourceData, u *schema.UsageData) *schema.Resource {
+	region := lookupRegion(d, []string{})
+
+	sku := d.Get("sku.0.name").String()
+	capacity := d.Get("sku.0.capacity").Int()
+
+	if u != nil && u.Get("monthly_operations").Type != gjson.Null {
```
vdmgolub, line 49: It seems like incorporating two TF resources into one Infracost resource involves adding a bunch of conditionals. What do you think if we split them into independent resources with different files. This way this if is not needed. And `DPS` attribute won't be needed.

internal/providers/terraform/azure/iothub.go
```diff
@@ -0,0 +1,65 @@
...
+			Capacity: capacity,
+			DPS:      true,
+		}
+
+		r.PopulateUsage(u)
+
+		return r.BuildResource()
+	}
+
+	log.Warnf("Skipping resource %s. Could not find a way to get its cost components from the resource or usage file.", d.Address)
```
vdmgolub, line 63: If `monthly_operations` is not set we can still show the price of the cost component. So I guess we can always build the resource. It would look like this:
```
 aws_lambda_function.hello_world
 ├─ Requests                                    Monthly cost depends on usage: $0.20 per 1M requests
 └─ Duration                                    Monthly cost depends on usage: $0.0000166667 per GB-seconds
```

I would also add a test resource with missing usage param to test such output.

internal/resources/azure/iothub.go
```diff
@@ -0,0 +1,104 @@
...
+
+	Operations *int64 `infracost_usage:"monthly_operations"`
+}
+
+func (r *IoTHub) PopulateUsage(u *schema.UsageData) {
+	resources.PopulateArgsWithUsage(r, u)
+}
+
+var OperationsUsageSchema = []*schema.UsageItem{
+	{Key: "monthly_operations", DefaultValue: 0, ValueType: schema.Float64},
```
vdmgolub, line 30: I think it can be an integer value, as we can't have fractional amount of operations.
```suggestion
	{Key: "monthly_operations", DefaultValue: 0, ValueType: schema.Int64},
```

internal/resources/azure/iothub.go
```diff
@@ -0,0 +1,104 @@
...
+func (r *IoTHub) iotHubDPSCostComponent() []*schema.CostComponent {
+	var quantity *decimal.Decimal
+	itemsPerCost := 1000
+
+	value := decimal.NewFromInt(*r.Operations)
+	quantity = decimalPtr(value.Div(decimal.NewFromInt(int64(itemsPerCost))))
+
+	costComponents := []*schema.CostComponent{
+		{
+			Name:            fmt.Sprintf("Instance (on-demand, %s)", r.Sku),
```
vdmgolub, line 63: Per issue's description this component is about device provisioning operations, I guess this name works better than "instance":
```suggestion
			Name:            "Device provisioning",
```

internal/resources/azure/iothub.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	return costComponents
+}
+
+func (r *IoTHub) iotHubCostComponent() []*schema.CostComponent {
+	costComponents := []*schema.CostComponent{
+		{
+			Name:            fmt.Sprintf("Instance (on-demand, %s)", r.Sku),
+			Unit:            "month",
```
vdmgolub, line 87: It seems like this cost component should describe "hub units", not instances:
```suggestion
			Name:            "IoT Hub",
			Unit:            "units",
```

internal/providers/terraform/azure/testdata/iothub_test/iothub_test.tf
```diff
@@ -0,0 +1,41 @@
...
+}
+
+# Add example resources for Iothub below
+resource "azurerm_iothub" "example" {
+  name                = "Example-IoTHub"
+  resource_group_name = azurerm_resource_group.example.name
+  location            = azurerm_resource_group.example.location
+
+  sku {
+    name     = "S1"
```
vdmgolub, line 18: We probably should also handle the case with `F1` free tier when the resource has no price and is skipped. Here's an example from another resource:
```go
	if r.SKU == skuFree {
		return &schema.Resource{
			Name:        r.Address,
			IsSkipped:   true,
			NoPrice:     true,
			UsageSchema: LogAnalyticsWorkspaceUsageSchema,
		}
	}
```
https://github.com/infracost/infracost/blob/master/internal/resources/azure/log_analytics_workspace.go#L131-L138

internal/resources/azure/iothub.go
```diff
@@ -0,0 +1,104 @@
...
+			UnitMultiplier:  decimal.NewFromInt(1),
+			MonthlyQuantity: decimalPtr(decimal.NewFromInt(1)),
```
vdmgolub, line 89: It seems like `r.Capacity` can be the `MonthlyQuantity` here. This way we don't need to use `schema.MultiplyQuantities`. What do you think?

internal/providers/terraform/azure/testdata/iothub_test/iothub_test.tf
```diff
@@ -0,0 +1,41 @@
...
+
+resource "azurerm_iothub_dps" "example" {
+  name                = "example"
+  resource_group_name = azurerm_resource_group.example.name
+  location            = azurerm_resource_group.example.location
+  allocation_policy   = "Hashed"
+
+  sku {
+    name     = "S1"
+    capacity = 1
```
vdmgolub, line 35: I saw your question about DPS' capacity. I couldn't find anything explicit about this too, but we probably still should multiply the number of operations by this capacity. My guess would be that capacity is related to IoT Hub units that are created along with this resource, because `linked_hubs` block is optional.

infracost-usage-example.yml
```diff
@@ -964,6 +964,9 @@ resource_usage:
   azurerm_firewall.my_firewall:
     monthly_data_processed_gb: 100000 # Monthly data processed by the firewall in GB.
 
+  azurerm_iothub_dps.my_iothub_dps:
+    monthly_operations: 10000
```
vdmgolub, line 968: We usually try to put a comment near the usage param value as documentation, so maybe we could include something like this:
```suggestion
    monthly_operations: 10000 # Monthly number of device provisioning operations
```

vdmgolub (commented):
Great work, thank you! I've left couple comments, it seems that we could simplify things a bit.

internal/providers/terraform/azure/testdata/iothub_test/iothub_test.tf
sawyerw-mastery: https://docs.microsoft.com/en-us/azure/iot-dps/concepts-service?WT.mc_id=Portal-Microsoft_Azure_IotHub#linked-iot-hubs

I don't think capacity is tied to the linked_hubs in any way. They would be created as separate resources and then linked in.

vdmgolub: Sounds good. I guess we could mention in usage param comment that users should account the capacity in the `monthly_operations` value. At least they will be aware.

vdmgolub (approved):
Awesome work! Thank you @sawyerw-mastery! :) I've run the CI on Infracost repo and it was all green: https://github.com/infracost/infracost/actions/runs/2662251512. This will go out with the next release.

craigbryson:
Thanks for this @sawyerw-mastery

PR #1854, opened 2022-07-13
enhance: format stacktraces so we can match them when reported

aliscott:
* Make sure stack traces match when reported by not showing abs paths or function args
* Report the stacktrace up as a separate property

### DIFF
--- a/internal/clierror/error.go
+++ b/internal/clierror/error.go
@@ -1,22 +1,152 @@
 package clierror
 
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	"github.com/maruel/panicparse/v2/stack"
+	log "github.com/sirupsen/logrus"
+)
+
 // SanitizedError allows errors to be wrapped with a sanitized message for sending upstream
-type SanitizedError struct {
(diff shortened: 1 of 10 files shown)


internal/clierror/error.go
```diff
@@ -1,22 +1,152 @@
...
+	}
+
+	return string(sanitizedStack)
+}
+
+// processStack processes the raw stack trace into a format that can be reported upstream.
+// This strips out the absolute path information of the local machine and removes any function args
+// so that the reported value always equals the same string value as an equivalent stack trace.
+// Adapted from: https://pkg.go.dev/github.com/maruel/panicparse/v2/stack#example-package-Text
+func processStack(rawStack []byte) ([]byte, error) {
```
hugorut, line 80: can we add a defer panic catch in this function since this does a lot and it's designed to handle panics. We don't want to do all this work and then just not catch a panic at this level!

hugorut (approved):

internal/clierror/error.go
aliscott: Discussed in person. We could change this so it just sends the raw stack trace if we catch a panic here, but decided it wasn't worth doing just now.

PR #1861, opened 2022-07-14
fix: parser index panic

aliscott:
This was happening in the VS Code extension when a resource name is blank. Although this is technically invalid Terraform, we can't guarantee the user always has valid Terraform while writing the code so we shouldn't fail on this.

### DIFF
--- a/internal/providers/terraform/parser.go
+++ b/internal/providers/terraform/parser.go
@@ -385,10 +385,6 @@ func parseRegion(providerConf gjson.Result, vars gjson.Result, providerKey strin
 		}
 	}
 
-	if strings.Contains(region, "mock") {
-		return ""
-	}
-
 	return region
 }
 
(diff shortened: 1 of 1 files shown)


internal/providers/terraform/parser.go
```diff
@@ -385,10 +385,6 @@ func parseRegion(providerConf gjson.Result, vars gjson.Result, providerKey strin
 		}
 	}
 
-	if strings.Contains(region, "mock") {
-		return ""
-	}
-
```
hugorut: don't remove this please

internal/providers/terraform/parser.go
```diff
@@ -616,9 +612,11 @@ func addressResourcePart(addr string) string {
 
 	if len(p) >= 3 && p[len(p)-3] == "data" {
 		return strings.Join(p[len(p)-3:], ".")
+	} else if len(p) >= 2 {
+		return strings.Join(p[len(p)-2:], ".")
 	}
```
hugorut, line 621:
```suggestion
	}
	
	 if len(p) >= 2 {
		return strings.Join(p[len(p)-2:], ".")
	}
```

internal/providers/terraform/parser.go
```diff
@@ -630,7 +632,7 @@ func addressModulePart(addr string) string {
 
 	if len(ap) >= 3 && ap[len(ap)-3] == "data" {
 		mp = ap[:len(ap)-3]
-	} else {
+	} else if len(ap) >= 2 {
 		mp = ap[:len(ap)-2]
 	}
```
hugorut, line 637:
```suggestion
```suggestion
	}
	
	if len(ap) >= 2 {
		mp = ap[:len(ap)-2]
	}
```

hugorut (approved):
nice catch, minor formatting things.

PR #1868, opened 2022-07-18
fix(google): Update product filters for Google Storage Bucket resource

vdmgolub:
Google introduced new SKUs that have descriptions with 'Tagging' word,
the filter was catching two records. For example:
- "Archive Class A Operations" - existing record
- "Archive Tagging Class A Operations" - new one


hugorut (approved):
👍

PR #1877, opened 2022-07-19
refactor: change register command to point to auth login

hugorut:
* hides register command
* adds warning string when register is called and delegates to login command
* adds warning string for register help and shows login command help
* unhides auth command


vdmgolub (approved):
:rocket:

PR #1881, opened 2022-07-19
feat: Add --summary-only option to Bitbucket's output/comment commands

### LINKED ISSUE #1783: Long Bitbucket comments
## Problem

As shown in the following screenshot, Bitbucket does not support collapsible sections in pull request comments so the comment can become too big if the diff has more than a few changes.

Screen Shot 2022-06-27 at 3 07 47 PM

### 1. Ask Bitbucket to support collapsible sections

GitHub, GitLab and Azure Repos all support collapsible sections in pull request comments. If you're a Bitbucket user, please consider commenting on [this feature request](https://jira.atlassian.com/browse/BCLOUD-15185?focusedCommentId=2253944&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel) in Bitbucket so Infracost can do the same thing as GitHub/GitLab/Azure Repos.

### 2. Add `--summary-only` option to `infracost comment bitbucket`

This would result in the Infracost pull request comment to only have the top summary table. This is the simplest option but it means users don't have the diff, which is useful if the summary shows something that requires further investigation.

### 3. Similar to option 2 but with a share link

The [Infracost share link](https://www.infracost.io/docs/features/share_links/) feature could be used to generate a link to the diff, so there is always a summary table and a link to the details if `--summary-only` is used.

## Which option do you prefer?

Please add a comment with the option you prefer?.
Also 👍 if you'd like to upvote this or receive updates.

vdmgolub:
Fixes #1783.

Bitbucket does not support collapsible sections in pull request comments
so the comment can become too big if the diff has more than a few changes.

This would result in the Infracost pull request comment to only have the
top summary table. This is the simplest option but it means users don't
have the diff, which is useful if the summary shows something that requires
further investigation.


internal/output/templates.go
```diff
@@ -344,13 +344,16 @@ var CommentMarkdownWithHTMLTemplate = `
 </table>
 {{- end }}
 
+{{- if not .MarkdownOptions.OmitDetails }}
+
```
vdmgolub, line 348: Adding this to html+markdown template to keep all templates similar to each other as much as possible.

cmd/infracost/comment_bitbucket.go
```diff
@@ -137,6 +140,7 @@ func commentBitbucketCmd(ctx *config.RunContext) *cobra.Command {
 	cmd.Flags().Var(&prNumber, "pull-request", "Pull request number to post comment on")
 	cmd.Flags().String("repo", "", "Repository in format workspace/repo")
 	_ = cmd.MarkFlagRequired("repo")
+	cmd.Flags().Bool("summary-only", false, "Omit output details")
```
alikhajeh1, line 143:
```suggestion
	cmd.Flags().Bool("summary-only", false, "Omit diff details so only the summary table is output")
```

cmd/infracost/testdata/comment_bitbucket_summary_only/comment_bitbucket_summary_only.golden
```diff
@@ -0,0 +1,11 @@
...
+## Infracost estimate: **monthly cost will increase by $40.56 (+100%) ↑**
+
+| **Project** | **Previous** | **New** | **Diff** |
+| ----------- | -----------: | ------: | -------- |
+| infracost/infracost/cmd/infraco...data/terraform_v0.14_plan.json | $40.56 | $81.12 | +$40.56 (+100%) |
+
+This comment will be updated when the cost estimate changes.
+Is this comment useful? [Yes](https://dashboard.infracost.io/feedback/redirect?runId=&value=yes), [No](https://dashboard.infracost.io/feedback/redirect?runId=&value=no), [Other](https://dashboard.infracost.io/feedback/redirect?runId=&value=other)
+
+Comment not posted to Bitbucket (--dry-run was specified)
```
vdmgolub, line 11: @alikhajeh1 FYI, this is an example of the comment with `--summary-only`, it will also include policy checks output if they are set up, only details are excluded.

aliscott: Maybe we should call it `--exclude-cli-output` or something if it's just that that is hidden?

alikhajeh1: huh good idea @aliscott `--exclude-cli-output` is clearer than summary-only as @vdmgolub mentioned that policy checks are included too.

cmd/infracost/comment_bitbucket.go
```diff
@@ -137,6 +140,7 @@ func commentBitbucketCmd(ctx *config.RunContext) *cobra.Command {
 	cmd.Flags().Var(&prNumber, "pull-request", "Pull request number to post comment on")
 	cmd.Flags().String("repo", "", "Repository in format workspace/repo")
 	_ = cmd.MarkFlagRequired("repo")
+	cmd.Flags().Bool("summary-only", false, "Omit diff details so only the summary table is output")
```
vdmgolub, line 143: @alikhajeh1 Guess we could rephrase this description with the new flag `--exclude-cli-output`. Maybe something like `Do not include CLI output`? Not sure if negation works better than "exclude".

cmd/infracost/comment_bitbucket.go
```diff
@@ -137,6 +140,7 @@ func commentBitbucketCmd(ctx *config.RunContext) *cobra.Command {
 	cmd.Flags().Var(&prNumber, "pull-request", "Pull request number to post comment on")
 	cmd.Flags().String("repo", "", "Repository in format workspace/repo")
 	_ = cmd.MarkFlagRequired("repo")
+	cmd.Flags().Bool("exclude-cli-output", false, "Omit diff details so only the summary table is output")
```
alikhajeh1, line 143:
```suggestion
	cmd.Flags().Bool("exclude-cli-output", false, "Exclude CLI output so comment has just the summary table")
```

cmd/infracost/comment_bitbucket.go
alikhajeh1: good point, I added a suggestion.

hugorut (approved):
Nice

Issue #1574, opened 2022-04-20
Error on missing usage-file


alikhajeh1:
With v0.9.22 of the CLI, I took me a while to debug an issue as `infracost breakdown --path examples/terraform --usage-file BLAH.yaml` worked even though the file BLAH.yaml didn't exist. We should throw a nice error like we do for the config-file:
```
> infracost breakdown --config-file BLAH.yaml
Error: config file does not exist at BLAH.yaml
```

alikhajeh1:
@vdmgolub correct, the error should only be if: `--usage-file` is used AND the file does not exist AND `--sync-usage-file` is not being used.

aliscott:
Closing this since it is stale.

PR #1883, opened 2022-07-20
fix: error on missing usage file when sync-usage-file not in use

cjdalan:
Fixes: https://github.com/infracost/infracost/issues/1574

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -592,7 +592,7 @@ func (r *parallelRunner) generateUsageFile(ctx *config.ProjectContext, provider 
 		return errors.Wrap(err, "Error creating usage file")
 	}
 
-	usageFile, err = usage.LoadUsageFile(usageFilePath, true)
+	usageFile, err = usage.LoadUsageFile(usageFilePath, ctx.RunContext.Config.SyncUsageFile)
 	if err != nil {
 		return errors.Wrap(err, "Error loading usage file")
 	}
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -409,7 +409,7 @@ func (r *parallelRunner) runProjectConfig(ctx *config.ProjectContext) (*projectO
 
 	if ctx.ProjectConfig.UsageFile != "" {
 		var err error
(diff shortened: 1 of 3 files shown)


hugorut:
hi @cjdalan thanks for the PR. I agree it's a bit weird to fail silently here - let me check in with the team to see what was the original reasoning behind this. Also my initial thought is we might want to throw a `WARN` the user and not fail explicitly.

hugorut:
; I, I wasn't aware of https://github.com/infracost/infracost/issues/1574. Thanks for raising this PR. There are a few lint errors to fix; this looks like they might be originating from tests or other parts of the application where you haven't updated the caller to use the new function signature. If you have a crack at fixing those, I'll give this a once over, and we can merge it.

cjdalan:
i have some errors when running `make lint` unrelated to my changes:
```
golangci-lint run
WARN [linters context] bodyclose is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649. 
WARN [linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649. 
internal/hcl/attribute.go:617:5: SA4004: the surrounding loop is unconditionally terminated (staticcheck)
                                break
                                ^
internal/apiclient/auth.go:69:4: S1037: should use time.Sleep instead of elaborate way of sleeping (gosimple)
                        select {
                        ^
internal/apiclient/auth.go:68:3: S1000: should use for range instead of for { select {} } (gosimple)
                for {
                ^
make: *** [lint] Error 1
```

hugorut:
@cjdalan  hmm I'm not seeing that on my `lint` it might be because you have a different version of `golangci-lint` locally then what we use on ci. I'm seeing some test errors now. Could you resolve these? 

Quick tip, you can run the test suite locally using `make test` and then for a specific cloud vendor using a suffix, eg: `make test_aws`. You can read more about setting up the development environment in our contributing guide, [here](https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md#setting-up-the-development-environment).

alikhajeh1:
@cjdalan any update on this?

PR #1886, opened 2022-07-20
enhance: Mention organization name when uploading runs to Infracost Cloud

vdmgolub:


internal/apiclient/dashboard.go
```diff
@@ -126,6 +126,10 @@ func (c *DashboardAPIClient) AddRun(ctx *config.RunContext, out output.Root) (Ad
 			addRun(run: $run) {
 				id
 				shareUrl
+				organization {
+					id
+					name
+				}
```
vdmgolub, line 132: CI fails because production dashboard API doesn't have this field yet. Ran it locally with `INFRACOST_DASHBOARD_API_ENDPOINT` and it was green.

Not sure how to approach this and I question if we should do real requests to API. Yes, it's closer to the actual use, but API CI should cover this case already, we just need to test that the success/failure response is handled correctly which can be emulated with stubbed requests. Making real requests also pollutes production db as CI is run pretty frequently.

tim775 (approved):

internal/apiclient/dashboard.go
```diff
@@ -134,21 +138,28 @@ func (c *DashboardAPIClient) AddRun(ctx *config.RunContext, out output.Root) (Ad
...
-		response.RunID = results[0].Get("data.addRun.id").String()
-		response.ShareURL = results[0].Get("data.addRun.shareUrl").String()
+		cloudRun := results[0].Get("data.addRun")
+
+		orgName := cloudRun.Get("organization.name").String()
+		orgMsg := ""
+		if orgName != "" {
+			orgMsg = fmt.Sprintf("organization '%s' in ", orgName)
+		}
+		successMsg := fmt.Sprintf("Estimate uploaded to %sInfracost Cloud", orgMsg)
```
vdmgolub, line 153: @tim775 Sorry, found an edge case for the old API keys.

tim775 (approved):

PR #1892, opened 2022-07-22
chore: add friendly error message that tells user what to do

alikhajeh1:
@hugorut I'm not sure if there's a better way to handle this, feel free to push/take-over this branch, if you think it's needed also add a warning for "error reading Infracost JSON file"

Thanks to Benjamin Riou for the suggestion!
Screen Shot 2022-07-22 at 10 51 34 AM

### DIFF
--- a/internal/output/combined.go
+++ b/internal/output/combined.go
@@ -33,7 +33,7 @@ func Load(p string) (Root, error) {
 	var out Root
 	_, err := os.Stat(p)
 	if errors.Is(err, os.ErrNotExist) {
-		return out, errors.New("Infracost JSON file does not exist")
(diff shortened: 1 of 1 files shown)


hugorut (approved):
LGTM - sounds sensible

hugorut (changes requested):
actually, this is a wrapped error, so we can't do it like this - with the `%w` verb.

Let me fixup later today

hugorut (approved):
Played around with this a little locally and was thinking of a refactor. I left it as is for now because I think we need to introduce introduce a larger change around diagnostic errors vs user errors. I'm going to merge. Thanks

Issue #1874, opened 2022-07-19
`index out of range [1] with length 1` panic in for AWS CodeBuildProject


aliscott:
Version: v0.10.7

Stacktrace:
```
runtime error: index out of range [1] with length 1
goroutine 22 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:24 +0x68
main.(*parallelRunner).run.func1.1()
	github.com/infracost/infracost/cmd/infracost/run.go:304 +0x40
panic({0x10249bdc0, 0x14001bf1998})
	runtime/panic.go:838 +0x204
github.com/infracost/infracost/internal/resources/aws.(*CodeBuildProject).nameLabel(0x14000307400?)
	github.com/infracost/infracost/internal/resources/aws/codebuild_project.go:71 +0x1f0
github.com/infracost/infracost/internal/resources/aws.(*CodeBuildProject).BuildResource(0x14000307400)
	github.com/infracost/infracost/internal/resources/aws/codebuild_project.go:40 +0xa4
github.com/infracost/infracost/internal/providers/terraform/aws.NewCodeBuildProject(0x140009bbce0, 0x140033d80c0?)
	github.com/infracost/infracost/internal/providers/terraform/aws/codebuild_project.go:23 +0x38c
github.com/infracost/infracost/internal/providers/terraform.(*Parser).createResource(0x140009367c0, 0x140009bbce0, 0x0)
	github.com/infracost/infracost/internal/providers/terraform/parser.go:57 +0xe0
github.com/infracost/infracost/internal/providers/terraform.(*Parser).parseJSONResources(_, _, {_, _, _}, _, {_, {_, _}, {_, ...}, ...}, ...)
	github.com/infracost/infracost/internal/providers/terraform/parser.go:105 +0x404
github.com/infracost/infracost/internal/providers/terraform.(*Parser).parseJSON(0x140009367c0, {0x14002c1e000, 0x212e9, 0x41402}, 0x14000182d00?)
	github.com/infracost/infracost/internal/providers/terraform/parser.go:145 +0x3dc
github.com/infracost/infracost/internal/providers/terraform.(*HCLProvider).parseResources(0x14000637b00, {0x140000553e0?, 0x1006564c0?}, {0x14002c1e000, 0x212e9, 0x41402}, 0x1400090f938?)
	github.com/infracost/infracost/internal/providers/terraform/hcl_provider.go:181 +0x150
[...]

Issue #1891, opened 2022-07-22
Should Infracost validate users Terraform configuration


hugorut:
Infracost often runs on Terraform configuration that is incorrect or incomplete. For example, users can specify incorrect attribute values for Terraform blocks. Infracost has historically left validation of these values to Terraform and the `apply` step. This means that when users specify incorrect Terraform values, Infracost reports them as a `0` price resource.

This is often confusing for users as they believe that Infracost is providing incorrect estimates for their infrastructure. See https://github.com/infracost/infracost/issues/1890 as an example.

With the release of the Infracost [VSCode Extension](https://github.com/infracost/vscode-infracost), zero price outputs become even more confusing for users. As extension users are often prototyping or in a WIP state with their Infrastructure.

Infracost could provide validation for resources, highlighting that attribute values are incorrect and show users a list of acceptable values. This might provide immense benefit to users, as often infrastructure engineers spend their time looking at Terraform provider docs and cloud provider docs to try and figure out acceptable values before an `terraform apply`.

This would add additional complexity to Infracost and would have to be done in a way where we continually update acceptable values. Working as part of or on top of the pricing API and its values.

PR #1894, opened 2022-07-22
fix: `aws_codebuild_project` with invalid ComputeType causes index panic

hugorut:
Fixes https://github.com/infracost/infracost/issues/1874

Resolves issue where an environment.compute_type having an unexpected format caused a panic.
We now return a blank string for the name label if the compute_type is invalid.

### DIFF
--- a/internal/resources/aws/codebuild_project.go
+++ b/internal/resources/aws/codebuild_project.go
@@ -56,30 +56,26 @@ func (r *CodeBuildProject) BuildResource() *schema.Resource {
 }
 
 func (r *CodeBuildProject) nameLabel() string {
-	name := ""
 	switch r.EnvironmentType {
 	case "WINDOWS_SERVER_2019_CONTAINER":
-		name = "Windows ("
-		name += r.computeType()
+		return r.osWithComputeTypeLabel("Windows")
 	case "ARM_CONTAINER":
-		name = "Linux (arm1.large)"
+		return "Linux (arm1.large)"
 	case "LINUX_GPU_CONTAINER":
-		name = "Linux (gpu1.large)"
+		return "Linux (gpu1.large)"
 	default:
-		name = "Linux ("
-		name += r.computeType()
+		return r.osWithComputeTypeLabel("Linux")
 	}
-
-	return name
 }
 
-func (r *CodeBuildProject) computeType() string {
+func (r *CodeBuildProject) osWithComputeTypeLabel(os string) string {
 	pieces := strings.SplitAfter(r.ComputeType, "BUILD_")
 	if len(pieces) < 2 {
-		return ""
+		return os
 	}
 
-	return strings.Replace(strings.ToLower(pieces[1]), "_", ".", 1) + ")"
+	computeType := strings.Replace(strings.ToLower(pieces[1]), "_", ".", 1)
+	return fmt.Sprintf("%s (%s)", os, computeType)
 }
 
 func (r *CodeBuildProject) mapEnvironmentType() string {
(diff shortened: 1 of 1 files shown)


internal/resources/aws/codebuild_project.go
```diff
@@ -61,19 +60,28 @@ func (r *CodeBuildProject) nameLabel() string {
 	switch r.EnvironmentType {
 	case "WINDOWS_SERVER_2019_CONTAINER":
 		name = "Windows ("
-		name += strings.Replace(strings.ToLower(strings.SplitAfter(r.ComputeType, "BUILD_")[1]), "_", ".", 1) + ")"
+		name += r.computeType()
```
aliscott, line 63: This is going to return `Windows (` if `compute_type` is invalid. I'm not sure it's super important, because I'm guessing we don't find a price for it anyway, but it maybe we want something like:
```go
name = "Windows"
computeType := r.computeType()
if computeType {
  name += fmt.Sprintf(" (%s)", computeType)
}
```

aliscott (commented):
Do we want to return an `IsSkipped` resource for this case? I'm guessing we don't find any prices for it anyway, so doesn't seem worth sending a request to the pricing API.

internal/resources/aws/codebuild_project.go
hugorut: good catch, I'll update

hugorut:
>Do we want to return an IsSkipped resource for this case? I'm guessing we don't find any prices for it anyway, so doesn't seem worth sending a request to the pricing API.

Might be wise

hugorut:
Hmm, but when I do mark it as skipped, it adds to the skipped messaging as so:

```
        +∙ 1 is not supported yet, see https://infracost.io/requested-resources:
        +  ∙ 1 x aws_codebuild_project
```

I would find this confusing as a user, so I'm not going to mark it as skipped in this PR change. I think we should maybe have a section in the skip details which contains "problematic" resources that have been skipped.

I thought this is what the `IsSkipped` & `SkipMessage` combo would achieve, but it looks like I was wrong. I will add an issue for this to keep track of and prioritise if we want.

internal/resources/aws/codebuild_project.go
hugorut: resolved in https://github.com/infracost/infracost/pull/1894/commits/5dce8342b27e25f8e31609ae0f8103f8715f7a16

aliscott:
> I would find this confusing as a user, so I'm not going to mark it as skipped in this PR change.

Yeah I think this is what your issue here is also trying to solve? https://github.com/infracost/infracost/issues/1891.

We already skip resources currently if they use an invalid value for an attribute ([example here](https://github.com/infracost/infracost/blob/43a291c78c4d065154e946dd10361b93558998ce/internal/resources/azure/app_service_plan.go#L36-L42)). I don't think it's ideal since it doesn't distinguish between skipped because of invalid, or skipped because it's free. I think showing a $0 cost is probably more confusing for this case, since these are most likely occurring in VS Code so the $0 is visible, but the skipped message isn't.

hugorut:
> Yeah I think this is what your issue here is also trying to solve? https://github.com/infracost/infracost/issues/1891.

Something along those lines yep, not entirely sure how it would work, as the desired functionality/output would be considerably different based on VSCode vs CLI usage.

> We already skip resources currently if they use an invalid value for an attribute ([example here](https://github.com/infracost/infracost/blob/43a291c78c4d065154e946dd10361b93558998ce/internal/resources/azure/app_service_plan.go#L36-L42)). I don't think it's ideal since it doesn't distinguish between skipped because of invalid or skipped because it's free. I think showing a $0 cost is probably more confusing for this case since these are most likely occurring in VS Code so the $0 is visible but the skipped message isn't.

Good point, I'll update

aliscott (approved):

Issue #1873, opened 2022-07-19
`ForDisplay called on invalid Hostname` panic


aliscott:
Version: v0.10.7.

Error:
`ForDisplay called on invalid Hostname: idna: disallowed rune U+0040`

Stack trace:
No stack trace available, but looks like this is happening here: https://github.com/infracost/infracost/blob/9cc06e8e38334e0bd4ecd635b3c600b71f8326d0/internal/hcl/modules/credentials_source.go#L76

A potential fix might be to check `svchost.IsValid` when we parse out the module location here: https://github.com/infracost/infracost/blob/master/internal/hcl/modules/registry.go#L72.

aliscott:
This is now fixed in the [v0.10.9](https://github.com/infracost/infracost/releases/tag/v0.10.9) release.

PR #1895, opened 2022-07-22
fix: panic on invalid hostname used in credential discovery for private modules

hugorut:
Fixes: #1873

Resolves panic that originating from the `github.com/hashicorp/terraform-svchost` library because the host passed for
credential discovery was invalid and could not be converted back to a string.

### DIFF
--- a/internal/hcl/modules/registry.go
+++ b/internal/hcl/modules/registry.go
@@ -68,8 +68,12 @@ func (d Disco) ModuleLocation(source string) (RegistryURL, bool, error) {
 	}
 
 	host, namespace, moduleName, target := parts[0], parts[1], parts[2], parts[3]
+	hostname, err := svchost.ForComparison(host)
+	if err != nil {
+		return RegistryURL{}, false, fmt.Errorf("unable to use user provided module host %s as a Hostname for credential discovery %w", host, err)
+	}
 
-	serviceURL, err := d.disco.DiscoverServiceURL(svchost.Hostname(host), moduleServiceID)
+	serviceURL, err := d.disco.DiscoverServiceURL(hostname, moduleServiceID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/hcl/modules/registry.go
```diff
@@ -68,8 +68,12 @@ func (d Disco) ModuleLocation(source string) (RegistryURL, bool, error) {
 	}
 
 	host, namespace, moduleName, target := parts[0], parts[1], parts[2], parts[3]
+	hostname, err := svchost.ForComparison(host)
+	if err != nil {
+		return RegistryURL{}, false, fmt.Errorf("unable to use user provided module host %s as a Hostname for credential discovery %w", host, err)
```
aliscott, line 73:
```suggestion
		return RegistryURL{}, false, fmt.Errorf("unable to use user-provided module host %s as a Hostname for credential discovery: %w", host, err)
```

internal/hcl/modules/registry.go
```diff
@@ -90,7 +94,12 @@ func (d Disco) ModuleLocation(source string) (RegistryURL, bool, error) {
 }
 
 func (d Disco) DownloadLocation(moduleURL RegistryURL, version string) (string, error) {
-	serviceURL, err := d.disco.DiscoverServiceURL(svchost.Hostname(moduleURL.Host), moduleServiceID)
+	hostname, err := svchost.ForComparison(moduleURL.Host)
+	if err != nil {
+		return "", fmt.Errorf("unable to use module URL %s as a Hostname to dicsover service URL %w", moduleURL, err)
```
aliscott, line 99:
```suggestion
		return "", fmt.Errorf("unable to use module URL %s as a Hostname to discover service URL: %w", moduleURL, err)
```

aliscott (approved):

PR #1898, opened 2022-07-23
fix: fix HCL parsing module regex

aliscott:
* Fixes for modules that contain the word 'module' by matching only 'module.' at the beginning or '.module.' in the middle of the resource address.
* Fixes for_each and count support for submodules by making sure it only matches non [] characters inside [].

### DIFF
--- a/internal/hcl/evaluator.go
+++ b/internal/hcl/evaluator.go
@@ -27,8 +27,11 @@ import (
 )
 
 var (
-	errorNoVarValue     = errors.New("no value found")
-	modReplace          = regexp.MustCompile(`module\.|\[.*\]`)
+	errorNoVarValue = errors.New("no value found")
+	modReplace      = regexp.MustCompile(`^module\.`)
+	// Use a separate regex for nested modules since we want to replace this with '.'
+	nestedModReplace    = regexp.MustCompile(`\.module\.`)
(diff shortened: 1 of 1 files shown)


internal/hcl/evaluator.go
```diff
@@ -634,7 +637,11 @@ func (e *Evaluator) loadModule(b *Block) (*ModuleCall, error) {
 	if e.moduleMetadata != nil {
 		// if we have module metadata we can parse all the modules as they'll be cached locally!
 		for _, module := range e.moduleMetadata.Modules {
+			// Strip any "module." and "[*]" parts from the module name so it matches the manifest key format
 			key := modReplace.ReplaceAllString(b.FullName(), "")
+			key = nestedModReplace.ReplaceAllString(key, ".")
+			key = modArrayPartReplace.ReplaceAllString(key, "")
```
hugorut, line 643: Could we hoist this above the `range` as we don't need to do this for every loop (I know this was like this before, but it seems best to refactor now)

Then if we could add a `Debug` line on line `649` using the `key` and whatever `modulePath` has evaluated to. This might help tracking down problematic keys in the future.

aliscott: 👍

aliscott: Done

hugorut (approved):

Issue #1915, opened 2022-07-27
Missing cost for ECS Service when task definition is used from a module and family is used instead of ARN


anitakrueger:
I was able to reproduce this and made a sample in https://github.com/anitakrueger/infracost-test

It comes down to infracost not being able to make the link between the ecs service and the task definition when the task definition is in a module and the task_definition parameter is set to the family instead of the arn.
This does seem to work when the task definition and ecs service are in the same terraform. Then switching to the family, infracost can still make the link.
In our case we set the task definition as family because we want the latest revision in terraform

```
resource "aws_ecs_service" "service" {
  name                               = "infracost-test"
==> WORKS  task_definition                    = module.task.arn
  cluster                            = aws_ecs_cluster.ecs_cluster.arn
  platform_version                   = "LATEST"
  launch_type                        = "FARGATE"
  propagate_tags                     = "TASK_DEFINITION"
  desired_count                      = 1
  deployment_maximum_percent         = 100
  deployment_minimum_healthy_percent = 0
  network_configuration {
    subnets          = ["subnet-0f6c2ff31a59db1d3"]
    assign_public_ip = false
  }
}

resource "aws_ecs_service" "service-nocost" {
  name                               = "infracost-test"
==> BREAKS  task_definition                    = module.task.family
  cluster                            = aws_ecs_cluster.ecs_cluster.arn
  platform_version                   = "LATEST"
  launch_type                        = "FARGATE"
  propagate_tags                     = "TASK_DEFINITION"
  desired_count                      = 1
  deployment_maximum_percent         = 100
  deployment_minimum_healthy_percent = 0
  network_configuration {
    subnets          = ["subnet-0f6c2ff31a59db1d3"]
    assign_public_ip = false
  }
}
```

The cost output shows 2 estimated resources, but there is only cost for one of them:
```
infracost-test % infracost breakdown --path . --show-skipped        
Evaluating Terraform directory at .
  ✔ Downloading Terraform modules 
  ✔ Evaluating Terraform directory 
  ✔ Retrieving cloud prices to calculate costs 

Project: github.com/anitakrueger/infracost-test.git

 Name                     Monthly Qty  Unit  Monthly Cost 
                                                          
 aws_ecs_service.service                                  
 ├─ Per GB per hour               0.5  GB           $1.62 
 └─ Per vCPU per hour            0.25  CPU          $7.39 
                                                          
 OVERALL TOTAL                                      $9.01 
──────────────────────────────────
[...]

PR #1917, opened 2022-07-27
fix: fix references for `aws_ecs_service` to `aws_ecs_task_definition` with the `family` attribute

aliscott:
This uses the `CustomRefIDFunc` functionality to create the correct reference. Since `family` isn't really an ID, I added a check in the `aws_ecs_service` resource to make sure that the returned resource data type is an `aws_ecs_task_definition`.

I'm not sure if we have a current way of locking references to an expected target type and attribute. If not, it's probably something we want to explore in the future, or work out automatically based on the HCL.

This half fixes https://github.com/infracost/infracost/issues/1915. It works for cases where the user specifies the family:
```hcl
resource "aws_ecs_service" "my_service" {
  ...
  task_definition = "<family>"
  ...
}
```

but it doesn't work if the user also specifies a revision, i.e:
```hcl
resource "aws_ecs_service" "my_service" {
  ...
  task_definition = "<family>:<revision>"
  ...
}
```

@tim775 not sure if you have any ideas if that's possible/easy to fix just now?


internal/providers/terraform/aws/ecs_task_definition.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"github.com/infracost/infracost/internal/schema"
+)
+
+// This is a free resource but needs it's own custom registry item to specify the custom ID lookup function.
+func getECSTaskDefinitionRegistryItem() *schema.RegistryItem {
+	return &schema.RegistryItem{
+		Name:    "aws_ecs_task_definition",
+		NoPrice: true,
+		Notes:   []string{"Free resource 2."},
```
tim775, line 12:
```suggestion
		Notes:   []string{"Free resource."},
```

tim775 (approved):

internal/providers/terraform/aws/ecs_task_definition.go
tim775: > but it doesn't work if the user also specifies a revision, i.e:

I guess the issue is we don't have access to revision when parsing HCL because it's generated by aws.  The arn includes family + revision, so maybe the best thing we can do is add a note that to use an arn ref to get an exact version.  Or maybe even a warning if we detect the link includes the ":<revision>" part?

aliscott: > Suggested change 
Thanks!

> I guess the issue is we don't have access to revision...

Yeah I was thinking if they pass `family:revision` we should just use the task definition in the code that has that family, since most likely they are doing something like `family = "${aws_ecs_task_definition.task1.family}:${aws_ecs_task_definition.task1.revision}`. The problem is, as far as I remember, the way we match references means we match on the whole value of the attribute, so we can't match on the part before the `:`.

PR #1920, opened 2022-07-27
feat: add metadata support for additional VCS systems

hugorut:
Adds further VCS metadata for:

* Azure DevOps
* Bitbucket
* Atlantis
* Circle CI

### DIFF
--- a/internal/vcs/metadata.go
+++ b/internal/vcs/metadata.go
@@ -487,17 +487,18 @@ func (f *metadataFetcher) getBitbucketMetadata(path string) (Metadata, error) {
 
 	m.Pipeline = &Pipeline{ID: os.Getenv("BITBUCKET_BUILD_NUMBER")}
 	m.PullRequest = &PullRequest{
-		VCSProvider: "bitbucket",
-		ID:          os.Getenv("BITBUCKET_PR_ID"),
+		VCSProvider:  "bitbucket",
+		ID:           os.Getenv("BITBUCKET_PR_ID"),
+		SourceBranch: os.Getenv("BITBUCKET_BRANCH"),
+		BaseBranch:   os.Getenv("BITBUCKET_PR_DESTINATION_BRANCH"),
+		URL:          fmt.Sprintf("%s/pull-requests/%s", os.Getenv("BITBUCKET_GIT_HTTP_ORIGIN"), os.Getenv("BITBUCKET_PR_ID")),
+
 		// we're unable to fetch these without calling the Bitbucket API endpoint:
 		// https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/
 		// However, calling the API requires authentication with variables that we don't
 		// have access to in the pipeline (e.g. username, password for basic auth).
-		Title:        "",
-		Author:       "",
-		SourceBranch: os.Getenv("BITBUCKET_BRANCH"),
-		BaseBranch:   os.Getenv("BITBUCKET_PR_DESTINATION_BRANCH"),
-		URL:          fmt.Sprintf("%s/pull-requests/%s", os.Getenv("BITBUCKET_GIT_HTTP_ORIGIN"), os.Getenv("BITBUCKET_PR_ID")),
+		Title:  "",
+		Author: "",
 	}
 
 	return m, nil
--- a/internal/vcs/metadata.go
+++ b/internal/vcs/metadata.go
@@ -41,6 +41,7 @@ var (
 
 	scpSyntax        = regexp.MustCompile(`^([a-zA-Z0-9-._~]+@)?([a-zA-Z0-9._-]+):([a-zA-Z0-9./._-]+)(?:\?||$)(.*)$`)
 	mergeCommitRegxp = regexp.MustCompile(`(?i)^merge\s([\d\w]+)\sinto\s[\d\w]+`)
+	startsWithMerge  = regexp.MustCompile(`(?i)^merge`)
 )
 
 type keyMutex struct {
@@ -672,7 +673,7 @@ type Metadata struct {
 }
 
 func isMergeCommit(message string) bool {
-	return strings.Contains(strings.ToLower(message), "merge")
+	return startsWithMerge.MatchString(message)
 }
 
 // urlStringToRemote returns the provided string as a Remote struct, with the host extracted from the URL.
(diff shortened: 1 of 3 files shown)


internal/vcs/metadata.go
```diff
@@ -223,12 +288,332 @@ func (f *metadataFetcher) getLocalGitMetadata(path string) (Metadata, error) {
...
+	SourceRefName string    `json:"sourceRefName"`
+	TargetRefName string    `json:"targetRefName"`
+	MergeStatus   string    `json:"mergeStatus"`
+	MergeId       string    `json:"mergeId"`
+}
+
+// getAzureRepoPRInfo attempts to get the azurePullRequestResponse using Azure DevOps Pipeline variables.
+// This method is expected to often fail as Azure DevOps requires users to explicitly pass System.AccessToken as
+// an env var on the job step.
+func (f *metadataFetcher) getAzureRepoPRInfo() azurePullRequestResponse {
```
vdmgolub, line 378: Just to keep naming consistent
```suggestion
func (f *metadataFetcher) getAzureReposPRInfo() azurePullRequestResponse {
```

internal/vcs/metadata.go
```diff
@@ -223,12 +288,332 @@ func (f *metadataFetcher) getLocalGitMetadata(path string) (Metadata, error) {
...
+		logging.Logger.WithError(err).Debugf("could not decode response body from Azure DevOps pull request API call to '%s'", apiURL)
+	}
+
+	return out
+}
+
+func (f *metadataFetcher) getAzureReposGithubMetadata(path string) (Metadata, error) {
+	m, err := f.getLocalGitMetadata(path)
+	if err != nil {
+		return m, fmt.Errorf("AzureRepos metadata error, could not fetch initial metadata from local git %w", err)
```
vdmgolub, line 416:
```suggestion
		return m, fmt.Errorf("Azure Repos metadata error, could not fetch initial metadata from local git %w", err)
```

internal/vcs/metadata.go
```diff
@@ -223,12 +288,332 @@ func (f *metadataFetcher) getLocalGitMetadata(path string) (Metadata, error) {
...
+	}
+
+	if m.Branch.Name == "HEAD" {
```
vdmgolub, line 419:
```suggestion
	if strings.toLower(m.Branch.Name) == "head" {
```

internal/vcs/metadata.go
```diff
@@ -223,12 +288,332 @@ func (f *metadataFetcher) getLocalGitMetadata(path string) (Metadata, error) {
...
+		// provided to fetch this information from the GitHub API.
+		Title:  "",
+		Author: "",
+	}
+
+	return m, nil
+}
+
+// transformAzureDevOpsMergeCommit sets the first non merge Commit that metadataFetcher can find on
+// Metadata m. transformAzureDevOpsMergeCommit fetches a specific commit sha if it iss referenced
```
vdmgolub, line 446:
```suggestion
// Metadata m. transformAzureDevOpsMergeCommit fetches a specific commit sha if it is referenced
```

internal/vcs/metadata.go
```diff
@@ -223,12 +288,332 @@ func (f *metadataFetcher) getLocalGitMetadata(path string) (Metadata, error) {
...
+		return fmt.Sprintf("https://gitlab.com/%s/%s/-/merge_requests/%s", owner, project, pullNumber)
+	}
+
+	if strings.Contains(remote.Host, "azure") {
+		return fmt.Sprintf("https://dev.azure.com/%s/base/_git/%s/pullrequest/%s", owner, project, pullNumber)
+	}
+
+	if strings.Contains(remote.Host, "bitbucket") {
+		return fmt.Sprintf("https://bitbucket.org/%s/%s/pull-requests/%s", owner, project, pullNumber)
+	}
```
vdmgolub, line 612: This method would cover cloud/hosted versions of GitHub/Gitlab/Bitbucket, but it seems for Enterprise versions we need to check their respective env vars like `ATLANTIS_GH_HOSTNAME` or `ATLANTIS_BITBUCKET_BASE_URL`. Or maybe you tried that and it's not possible for some reason? I took the env var names from here https://www.runatlantis.io/docs/server-configuration.html#environment-variables.

internal/vcs/metadata.go
```diff
@@ -270,11 +655,57 @@ type Pipeline struct {
...
 type Metadata struct {
+	Remote      Remote
 	Branch      Branch
 	Commit      Commit
 	PullRequest *PullRequest
 	Pipeline    *Pipeline
 }
+
+func isMergeCommit(message string) bool {
+	return strings.Contains(strings.ToLower(message), "merge")
```
vdmgolub, line 675: This can provide false positives for commit messages like "feat: Merge X and Y into Z". Should we use a regex that ensures that the message starts with `/^merge/i`?

Another alternative might be checking commit's parents - merge commit should have multiple. So if we have `c` as go-git's commit struct, `len(c.ParentHashes()) > 1` would mean it's a merge commit (in theory, I haven't checked myself). What do you think?

vdmgolub (approved):
Looking great! Added mostly typo/copy update comments, but there is a question about Enterprise VCSs in Atlantis.

internal/vcs/metadata.go
hugorut: yup, nice catch, will update the URL string to use those vars if they exist.

internal/vcs/metadata.go
hugorut: Good shout with `regexp`. Yeah I tried using the `c.ParentHashes` approach but this seemed to be inconsistent as to which was the hash of the `source` and the `target` branch, so I skipped it.

internal/vcs/metadata.go
hugorut: Ah crap, yeah I remember, these aren't actually available to us in the run environment. These are purely server config vars, they don't seem to be exposed to the `run` step. I guess we could try and infer the enterprise versions from the `remote.Host` value

internal/vcs/metadata.go
hugorut: resolved in https://github.com/infracost/infracost/pull/1920/commits/15c83f1965286b7e7cc6e9e06dd19cec9a8fa136

PR #1934, opened 2022-07-29
feat: add TFC run task vcs metadata

hugorut:
This change enables TFC run tasks to show up in Infracost Cloud. It uses OS vars passed from the Cloud worker CLI run.

Cloud changes here: https://github.com/infracost/dashboard/pull/208

### DIFF
--- a/internal/vcs/metadata.go
+++ b/internal/vcs/metadata.go
@@ -125,6 +125,12 @@ func (f *metadataFetcher) Get(path string) (Metadata, error) {
 		return f.getAtlantisMetadata(path)
 	}
 
+	_, ok = lookupEnv("TFC_RUN_ID")
+	if ok {
+		logging.Logger.Debug("fetching TFC run task metadata")
+		return f.getTFCMetadata(path)
+	}
+
 	logging.Logger.Debug("could not detect a specific CI system, fetching local Git metadata")
 	return f.getLocalGitMetadata(path)
 }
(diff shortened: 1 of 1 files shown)


aliscott (approved):

internal/vcs/metadata.go
```diff
@@ -592,6 +598,79 @@ func (f *metadataFetcher) getAtlantisMetadata(path string) (Metadata, error) {
 	return m, nil
 }
 
+// getTFCMetadata returns Metadata built from a TFC request: https://www.terraform.io/cloud-docs/api-docs/run-tasks/run-tasks-integration#request-body
+// We cannot rely on any local git information we run this process in the Infracost Cloud. All OS variables are populated
+// by the Infracost Cloud run task worker, which passes them as environment flags when running the Infracost CLI.
+func (f *metadataFetcher) getTFCMetadata(path string) (Metadata, error) {
+	remote := urlStringToRemote(os.Getenv("TFC_VCS_URL"))
```
aliscott, line 605: Since we're creating these env vars why not follow the same format as the existing `INFRACOST_VCS_` ones?
`INFRACOST_VCS_PULL_REQUEST_URL`
`INFRACOST_VCS_REPOSITORY_URL`
...

aliscott (commented):
Looks good. I'd not seen the remote parsing stuff before, it looks neat!

internal/vcs/metadata.go
hugorut: Yeah can do, thought it would be more explicit - but it makes sense to reuse existing ones. Ill fixup now

aliscott: I think there's maybe two other reasons to use `INFRACOST_VCS_` for these:
1. It scopes it to Infracost, so we make sure we don't ever conflict with other vars.
2. If we have generic `INFRACOST_VCS_` env vars we can use them for any deployment system to integrate with Infracost Cloud.
What do you think?

hugorut: resolved in https://github.com/infracost/infracost/pull/1934/commits/0fd966d2375e72c8c97f158857920ca8849df8ad

aliscott (approved):
LGTM

Issue #1938, opened 2022-07-31
terraform.workspace parameter not parsed --terraform-workspace 


danielgadd:
Hi Infracost Team!

The terraform parameter `terraform.workspace` doesn't appear to be parsed correctly when using the `--terraform-workspace` argument?
I use workspaces (dev/stage/prod)  to differentiate things like instance_types (without the need for additional variables) - If I have the below code snippet:

```
resource "aws_elasticache_replication_group" "redis" {
  replication_group_id       = "rep-group-1"
  description                = "ElastiCache Cluster"
  node_type                  = terraform.workspace == "prod" ? "cache.t3.medium" : "cache.t3.small"
... snipped
}
```

And run infracost: (both locally compiled, and latest release binary)

```
$ infracost breakdown --path . --show-skipped --terraform-workspace prod
```

The parser doesn’t seem to evaluate the `terraform.workspace` parameter correctly and I always get the false value (cache.t3.small)
**note;** It works correctly if I use an input variable instead of the workspace variable (`var.environment == "prod ? "cache.t3.medium" : "cache.t3.small"`)

aliscott:
Thanks for writing up the issue @danielgadd 🙏

hugorut:
Thanks @danielgadd for raising this issue. We've merged a fix for this bug into master, and it will go out with the next release (0.10.10). We'll notify you here when it's available.

danielgadd:
Thanks @hugorut - Tested in my environment with the latest commit in master and can confirm its working 👍

tim775:
@danielgadd v0.10.10 is live and includes this fix.

PR #1954, opened 2022-08-08
fix: use terraform workspace for evaluator if provided

hugorut:
fixes: https://github.com/infracost/infracost/issues/1938

This change resolves `--terraform-workspace` flag being ignored for the HCL provider. User provided
workspaces are now set on the `Parser` so they can be used in context evaluation.


aliscott (approved):
LGTM!

Issue #1946, opened 2022-08-03
panic: `not a string` in HCL evaluator


aliscott:
Infracost version: v0.10.8

Stacktrace:
(I've fixed to get the actual file paths in https://github.com/infracost/infracost/pull/1929, this is happening in evaluator.go):
```
1: running [Created by errgroup.(*Group).Go @ errgroup.go:72]
   runtime/debug:24                                                Stack()
   main:297                                                        (*parallelRunner).run.func1.1()
   :838                                                            panic()
   github.com/zclconf/go-cty/cty:1256                              Value.AsString()
   github.com/zclconf/go-cty/cty:1326                              Value.AsValueMap()
   github.com/infracost/infracost/internal/hcl:603                 expandedEachBlockToValue()
   github.com/infracost/infracost/internal/hcl:556                 (*Evaluator).evaluateResource()
   github.com/infracost/infracost/internal/hcl:532                 (*Evaluator).getValuesByBlockType()
   github.com/infracost/infracost/internal/hcl:226                 (*Evaluator).evaluateStep()
   github.com/infracost/infracost/internal/hcl:200                 (*Evaluator).evaluate()
   github.com/infracost/infracost/internal/hcl:180                 (*Evaluator).Run()
   github.com/infracost/infracost/internal/hcl:344                 (*Parser).ParseDirectory()
   github.com/infracost/infracost/internal/providers/terraform:237 (*HCLProvider).Modules()
   github.com/infracost/infracost/internal/providers/terraform:205 (*HCLProvider).LoadPlanJSONs()
   github.com/infracost/infracost/internal/providers/terraform:160 (*HCLProvider).LoadResources()
   main:469                                                        (*parallelRunner).runProjectConfig()
   main:305                                                        (*parallelRunner).run.func1()
   golang.org/x/sync/errgroup:75                                   (*Group).Go.func1()
```

PR #1956, opened 2022-08-09
fix: catch evaluator panic if value map is not a `cty.ObjectType`

hugorut:
fixes: https://github.com/infracost/infracost/issues/1946

Resolves issue where calling `AsValueMap` for a non-valid type will panic. We now skip the `AsValueMap` call and send a debug write:

```
DEBU[0000] skipping unexpected value value type: 'cty.ListVal([]cty.Value{cty.StringVal("test")})' for existing for_each context value  block="test.test[\"two\"]" module_name=root ...
```

This unexpected value case should, in reality, never happen, as we always expect the existing values to be a valid `ObjectType`. So there's probably a hidden bug somewhere here. However, I could not replicate it with any HCL test cases. These log lines should give us valid context for the future so that we could solve any pricing bugs that arise from this case.

### DIFF
--- a/internal/hcl/evaluator.go
+++ b/internal/hcl/evaluator.go
@@ -556,7 +556,7 @@ func (e *Evaluator) evaluateResource(b *Block, values map[string]cty.Value) cty.
 
 	if k := b.Key(); k != nil {
 		e.logger.Debugf("expanding block %s to be available for for_each key %s", b.FullName(), *k)
-		valueMap[stripCount(labels[1])] = expandedEachBlockToValue(b, valueMap)
+		valueMap[stripCount(labels[1])] = e.expandedEachBlockToValue(b, valueMap)
 		return cty.ObjectVal(valueMap)
 	}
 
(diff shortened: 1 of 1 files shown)


internal/hcl/evaluator.go
```diff
@@ -593,16 +593,26 @@ func expandCountBlockToValue(b *Block, existingValues map[string]cty.Value) cty.
...
+			e.logger.WithFields(logrus.Fields{
+				"block": b.Label(),
+			}).Debugf("skipping unexpected cty value type '%s' for existing for_each context value", eachMap.GoString())
+
+			ob[*k] = b.Values()
+			return cty.ObjectVal(ob)
+		}
+
 		for e, v := range eachMap.AsValueMap() {
 			ob[e] = v
```
vdmgolub, line 617: I guess it's better to rename this `e` variable as it's the same as the struct now.

vdmgolub (approved):

internal/hcl/evaluator.go
hugorut: good catch 👍

Issue #1932, opened 2022-07-29
panic: `invalid memory address or nil pointer dereference` in google_container_node_pool


aliscott:
Infracost version: v0.10.7

Stacktrace:
```
runtime error: invalid memory address or nil pointer dereference
goroutine 23 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.18.3/x64/src/runtime/debug/stack.go:24 +0x65
main.(*parallelRunner).run.func1.1()
	/home/runner/work/infracost/infracost/cmd/infracost/run.go:304 +0x46
panic({0x1da70a0, 0x3713930})
	/opt/hostedtoolcache/go/1.18.3/x64/src/runtime/panic.go:838 +0x207
github.com/infracost/infracost/internal/schema.MultiplyQuantities(0xc00043edc0, {0xc00a0b0c78?, 0x22a3215?})
	/home/runner/work/infracost/infracost/internal/schema/resource.go:111 +0x5e
github.com/infracost/infracost/internal/resources/google.(*ContainerNodePool).BuildResource(0xc007ee6bc0)
	/home/runner/work/infracost/infracost/internal/resources/google/container_node_pool.go:65 +0x3b6
github.com/infracost/infracost/internal/providers/terraform/google.newContainerNodePool(0xc00043e790, 0xc009affd10?)
	/home/runner/work/infracost/infracost/internal/providers/terraform/google/container_node_pool.go:40 +0x292
github.com/infracost/infracost/internal/providers/terraform.(*Parser).createResource(0xc0007baf20, 0xc00043e790, 0x0)
	/home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:57 +0x12b
github.com/infracost/infracost/internal/providers/terraform.(*Parser).parseJSONResources(_, _, {_, _, _}, _, {0x5, {0xc000aa6000, 0x208cc}, {0x0, ...}, ...}, ...)
	/home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:105 +0x5ef
github.com/infracost/infracost/internal/providers/terraform.(*Parser).parseJSON(0xc0007baf20, {0xc003820000, 0x208cc, 0x20b67}, 0xc0010f9a00?)
	/home/runner/work/infracost/infracost/internal/providers/terraform/parser.go:145 +0x648
[...]

PR #1960, opened 2022-08-10
fix: panic caused from invalid guest accelerator cost component

hugorut:
fixes: https://github.com/infracost/infracost/issues/1932

This change resolves a panic where a nil `CostComponent` was being attributed to a `Resource` if a
`guest_accelerator.type` was an unexpected string. Fixes include:

* Altered all callers to expect that `guestAcceleratorCostComponent` could return nil if the `type` is invalid.
* Added unhandled `nvidia-tesla-a100` type to list of supported valid guest accelerator types

### DIFF
--- a/internal/resources/google/compute_cost_component_helpers.go
+++ b/internal/resources/google/compute_cost_component_helpers.go
@@ -156,7 +156,7 @@ func guestAcceleratorCostComponent(region string, purchaseOption string, guestAc
 		name = "NVIDIA Tesla K80"
 		descPrefix = "Nvidia Tesla K80 GPU"
 	case "nvidia-tesla-a100":
-		name = " NVIDIA Tesla A100"
+		name = "NVIDIA Tesla A100"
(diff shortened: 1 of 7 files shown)


internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -153,7 +155,11 @@ func guestAcceleratorCostComponent(region string, purchaseOption string, guestAc
 	case "nvidia-tesla-k80":
 		name = "NVIDIA Tesla K80"
 		descPrefix = "Nvidia Tesla K80 GPU"
+	case "nvidia-tesla-a100":
+		name = " NVIDIA Tesla A100"
```
aliscott, line 159:
```suggestion
		name = "NVIDIA Tesla A100"
```

aliscott (approved):
Looks good. One tiny change.

Thanks for fixing the unsupported option 🙏

PR #1964, opened 2022-08-11
enhance: Truncate GitHub comment if it's too large

vdmgolub:
GitHub has a limit of 262,144 characters per comment body.
https://github.com/orgs/community/discussions/27190#discussioncomment-3254953

### DIFF
--- a/internal/output/markdown.go
+++ b/internal/output/markdown.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"sort"
 	"text/template"
+	"unicode/utf8"
 
 	"github.com/infracost/infracost/internal/ui"
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


internal/output/markdown.go
```diff
@@ -189,13 +197,21 @@ func ToMarkdown(out Root, opts Options, markdownOpts MarkdownOptions) ([]byte, e
...
 		markdownOpts})
 	if err != nil {
 		return []byte{}, err
 	}
 
 	bufw.Flush()
-	return buf.Bytes(), nil
+	msg := buf.Bytes()
+
+	msgLength := utf8.RuneCountInString(string(msg))
```
hugorut, line 210:
```suggestion
	msgLength := utf8.RuneCount(msg)
```

hugorut (approved):

hugorut:
hold up - one of my review comments got lost for some reason.

internal/output/markdown.go
```diff
@@ -107,6 +108,13 @@ func ToMarkdown(out Root, opts Options, markdownOpts MarkdownOptions) ([]byte, e
 		return nil, errors.Wrap(err, "Failed to generate diff")
 	}
 
+	diffMsg := ui.StripColor(string(diff))
+	if opts.truncateLength > 0 {
+		truncateMsg := "\n\n...(truncated due to message size limit)...\n\n"
+		truncateLength := utf8.RuneCountInString(diffMsg) - opts.truncateLength - 1000
+		diffMsg = truncateMiddle(diffMsg, truncateLength, truncateMsg)
+	}
```
hugorut, line 116: So I know I suggested this `truncateLength` approach, but here's an alternative. You could to truncate the msg before the second `ToMarkdown` call and pass it as an optional `opts.diffMsg`. If this is set, you could skip the `ToDiff` call at the top of the function and use this. Doing this removes the unnecessary `ToDiff` call on the second `ToMarkdown` call.

I'll leave it to you to decide what to do tho.

hugorut (approved):

internal/output/markdown.go
vdmgolub: Looks like this: https://github.com/infracost/infracost/pull/1964/commits/3e140695e317aed174bbd16af4c6249b7f9feb88

Issue #1968, opened 2022-08-12
Terraform resources with count=0 appear in the costs


oliver-sentianse:
Awesome tool and presentation. 

BUT we noticed a cost breakdown error when we were trying it out today, so with today's version as installed per your instructions using curl. 

We have 2 terraform-managed stacks based on the same underlying terraform code, one stack has count=0 for the redshift DB, the other has count=1. I checked with `terraform state list` that redshift is only in the stack with count=1. Yet `infracost breakdown --path=.` in each folder reports redshift cost even for stack where count=0. Since redshift is almost 90% of the cost of that stack, it is major issue for us. 
```
$ infracost --version
Infracost v0.10.9
$ terraform version
Terraform v1.2.6
on linux_amd64
+ provider registry.terraform.io/hashicorp/aws v4.23.0
...
```

I really hope this can be fixed so we can use infracost!

tim775:
Hi @oliver-dss, thanks for the issue!  

I checked that `count = 0` is working at least in simple cases.  Could you share a sanitized version of the terraform code that causes the problem?  There are some simple [redshift examples here](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/testdata/redshift_cluster_test/redshift_cluster_test.tf) if you prefer to start from scratch. 

It also can be helpful if you run with `--log-level debug` and share the output.

Cheers

schollii:
Ah good to know. I'm sure it's nothing to do with redshift specifically, it was just obvious in that particular case. 

I suspect it is because the `count` is set from json loaded via `jsondecode(file(...))`. 

I might be afk for next days so if you don't get to try the above sooner, I'll try to pare down my stack and post the code when I get back. 

Thanks for a quick reply!

oliver-dss:
So this is super weird: I was able to pare it down to just one small .tf file and a .json file, and I could see the error. Then I added a couple more cases, one using a json string hardcoded and another using a variable with default value, and indeed those 2 were good. Then the error on the file case was gone. I commented out the 2 extra cases and the error is gone. 

My code was this: 
```
variable "tfvars_count" {
  type = number
  default = 0
}

locals {
  json_str = jsondecode("{\"count\": 0}")
  json_file = jsondecode(file("test.json"))
}

resource "aws_redshift_cluster" "test_from_str" {
  count = local.json_str.count

  cluster_identifier = "test_from_str"
  node_type = "dc2.large"
}

[...]

PR #1972, opened 2022-08-15
fix: fix HCL parsing values within local objects

aliscott:
This is a potential fix for https://github.com/infracost/infracost/issues/1760.

This fixes 2 underlying issues:
1. When evaluating local objects with mocks, bad variables from objects weren't supported, which meant the whole object is mocked instead of specififc values
2. When merging object variables the arg order was incorrect, which meant a mocked value was overritten as an unknown type

### DIFF
--- a/cmd/infracost/hcl_test.go
+++ b/cmd/infracost/hcl_test.go
@@ -114,3 +114,15 @@ func TestHCLModuleForEach(t *testing.T) {
 		nil,
 	)
 }
+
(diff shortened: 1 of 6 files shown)


hugorut (approved):
Solid work, could also be a fix for: https://github.com/infracost/infracost/issues/1968

aliscott:
> could also be a fix for: https://github.com/infracost/infracost/issues/1968

Unfortunately it doesn't seem to fix this case.

PR #1979, opened 2022-08-18
feat: resource type usage data

### LINKED ISSUE #1978: Add resource type usage to CLI, default to that if available
Based on the discussion here:
https://infracost-community.slack.com/archives/C01G63RG63D/p1660313212869199

balazs-marjan:
Fixes #1978 

- [ ] Infracost team to review/test PR
- [ ] @alikhajeh1 to create corresponding docs PR

### DIFF
--- a/internal/schema/usage_data.go
+++ b/internal/schema/usage_data.go
@@ -1,9 +1,11 @@
 package schema
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
+	"github.com/imdario/mergo"
 	jsoniter "github.com/json-iterator/go"
 	"github.com/tidwall/gjson"
 )
@@ -148,4 +150,34 @@ func ParseAttributes(i interface{}) map[string]gjson.Result {
 	}
 
 	return a
+}
+
+func MergeAttributes(dst *UsageData, src *UsageData) {
+	for key, srcAttr := range src.Attributes {
+		if _, has := dst.Attributes[key]; has {
+			switch srcAttr.Type {
+			case gjson.Null:
+				fallthrough
+			case gjson.True:
+				fallthrough
+			case gjson.False:
+				fallthrough
+			case gjson.Number:
+				fallthrough
+			case gjson.String:
+				// Should be safe to override
+				dst.Attributes[key] = srcAttr
+			case gjson.JSON:
+				var destJson map[string]interface{}
+				var srcJson map[string]interface{}
+				json.Unmarshal([]byte(dst.Attributes[key].Raw), &destJson)
+				json.Unmarshal([]byte(srcAttr.Raw), &srcJson)
+				mergo.Map(&destJson, srcJson)
+				src, _ := json.Marshal(destJson)
+				dst.Attributes[key] = gjson.Parse(string(src))
+			}
+		} else {
+			dst.Attributes[key] = srcAttr
+		}
+	}
 }
--- a/internal/schema/usage_data.go
+++ b/internal/schema/usage_data.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/imdario/mergo"
 	jsoniter "github.com/json-iterator/go"
+	log "github.com/sirupsen/logrus"
 	"github.com/tidwall/gjson"
 )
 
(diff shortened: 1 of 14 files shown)


hugorut:
Thanks @balazs-marjan for this. I'll take a peek at this tomorrow morning (GMT) if that's ok.

alikhajeh1:
@balazs-marjan and @hiltol thanks for this! We discussed it internally today and we like it a lot :) 
We just need time to review, test (e.g. with precedence and duplicates and other edge cases) and document it. We're heads-down with our August goals but I've added it to our September goals so it'll be done in the next few weeks. Hopefully you're not blocked until then as you can use this in your own fork.

aliscott:
@balazs-marjan this is great thanks. I've just been testing this out. I've added one case below that I think we should address.

When I have the following s3 bucket:
```hcl
resource "aws_s3_bucket" "my-bucket" {
  bucket = "my-tf-test-bucket"
}
```

And a usage file:
```yml
version: 0.1
resource_type_usage:
  aws_s3_bucket:
    standard:
      storage_gb: 100

resource_usage:
  aws_s3_bucket.my-bucket:
    standard:
      monthly_tier_1_requests: 1000
```

I see the following result:
```
 Name                                             Monthly Qty  Unit                    Monthly Cost

 aws_s3_bucket.my-bucket
 └─ Standard
    ├─ Storage                              Monthly cost depends on usage: $0.023 per GB
    ├─ PUT, COPY, POST, LIST requests                       1  1k requests                    $0.01
    ├─ GET, SELECT, and all other requests  Monthly cost depends on usage: $0.0004 per 1k requests
    ├─ Select data scanned                  Monthly cost depends on usage: $0.002 per GB
    └─ Select data returned                 Monthly cost depends on usage: $0.0007 per GB

 OVERALL TOTAL                                                                                $0.01
```

I would expect the `storage_gb` value from the `resource_type_usage` to be used if the `resource_usage` doesn't specify that key. What do you think? Maybe we need to deep merge the `resource_usage` into the `resource_type_usage`.

cc @alikhajeh1 if he has any opinions.

balazs-marjan:
@alikhajeh1 @aliscott @hiltol
Sure we could do that. I think this only depends on what your vision is for the UX of the CLI. Personally I think that if it behaved as a merge too, then a more extensive documentation is needed with the feature. In my experience, defaults and overrides are usually not straightforward to users.

hiltol:
I would expect both values (storage_gb and monthly_tier_1_requests) in the usage-file to be respected, unless the default one is overridden.

aliscott:
Thanks @balazs-marjan and @hiltol, after chatting with @alikhajeh1 we agree that the `resource_type_usage` should be the default for all resources that have that type and `resource_usage` should override (be deep-merged) into that to get the usage for each resource.

I think this will increase the usability since it means the user doesn't need to copy all the resource type usages for any resource they want to override specific values for.

To make this functionality clearer I suggest we also rename `resource_type_usage` to `resource_type_default_usage`.

balazs-marjan:
@aliscott can you review again?

internal/usage/usage_file.go
```diff
@@ -122,10 +135,22 @@ See https://infracost.io/usage-file/ for docs`,
...
+		&u.RawResourceTypeUsage,
 		resourceUsagesKeyNode,
 		&u.RawResourceUsage,
 	)
 
+	// Add a comment to the first commented-out resource
+	for _, node := range u.RawResourceTypeUsage.Content {
+		if isNodeMarkedAsCommented(node) {
+			node.HeadComment = `##
+## The following usage values are all commented-out, you can uncomment resource types and customize as needed.
```
alikhajeh1, line 148:
```suggestion
## The following usage values apply to each resource of the given type, which is useful when you want to define defaults.
## All values are commented-out, you can uncomment resource types and customize as needed.
```

balazs-marjan: I adjusted the comments

aliscott (commented):
Thanks @balazs-marjan, this looks great! One thing we noticed when testing:

1. Run `make run -s ARGS="breakdown --path examples/terraform --usage-file infracost-usage.yml --sync-usage-file"`
2. Change `resource_type_default_usage` to uncomment and add a usage parameter.
3. Also change `resource_usage` to uncomment and add a usage parameter.
4. Re-run `make run -s ARGS="breakdown --path examples/terraform --usage-file infracost-usage.yml --sync-usage-file"`

The `resource_type_default_usage` value from step 2 is overwritten. We would expect this to not be overwritten when re-synced.

We'd expect the functionality to be the same as the `resource_usage` value from step 3, which is not overwritten.

balazs-marjan:
@aliscott I restructured the merging solution as it was never going to work properly, the bug that you found was indicative of that. I think I fixed that too now.

Once again, can you review? Thanks!

aliscott (commented):
Thanks for fixing this up @balazs-marjan, the changes look 🔥. Just some small linting issues and I think this is good to merge!

balazs-marjan:
@aliscott I added handling for the possible error cases. That should fix the linting issues found, suggestions obviously welcome.

internal/schema/usage_data.go
```diff
@@ -149,3 +152,50 @@ func ParseAttributes(i interface{}) map[string]gjson.Result {
...
+			case gjson.String:
+				// Should be safe to override
+				dst.Attributes[key] = srcAttr
+			case gjson.JSON:
+				var err error
+				var destJson map[string]interface{}
+				var srcJson map[string]interface{}
+				err = json.Unmarshal([]byte(dst.Attributes[key].Raw), &destJson)
+				if err != nil {
+					log.Error(err)
```
aliscott, line 177:
```suggestion
					log.Errorf("Error merging attribute '%s': %v", key, err)
```
I think for these log lines we should add some additional context, like above, so it helps us debug if there is any issue.

aliscott (commented):
Just a small comment on context in the log lines.

internal/schema/usage_data.go
balazs-marjan: Done

aliscott (approved):

aliscott:
Thanks @balazs-marjan! 🎉 This will go out in the next release.

Issue #1987, opened 2022-08-24
Nil panic when detecting local git metadata on windows


hiran-filho:
`debug: initializing workspace
debug: running Infracost in project: d:\
debug: running Infracost cmd cmd /C "set INFRACOST_CLI_PLATFORM=vscode && infracost breakdown --path "d:\" --format json --log-level info"
error: Infracost cmd error trace Error: Command failed: cmd /C "set INFRACOST_CLI_PLATFORM=vscode && infracost breakdown --path "d:\" --format json --log-level info"

[91mError:[0m An unexpected error occurred

runtime error: invalid memory address or nil pointer dereference`

dont show the cost in vscode

hugorut:
Hi @hiran-filho thanks for opening this issue. I've tracked down the error and it seems to be an error coming from within a library we use to detect local git metadata. It seems it's struggling to properly handle an error when it can't find the local `.git` directory in the workspace.

I've added this to our grooming agenda to discuss with the team, so we can prioritise accordingly. The panic seems to have been resolved in `go-git` (the lib we use) in this commit: https://github.com/go-git/go-git/commit/935af59cf64fbe49eb8baba9fe80e6b236daf593, but for some reason we don't seem to have these lines the infracost go module we've downloaded. Even though this is the latest release, hopefully it's just a simple go mod update/cache clean. I'm adding the stacktrace below and problem code for any engineers who do pick this up.

```
  :24                                             Stack()
   :68                                             Run.func1()
   :838                                            panic()
   :283                                            dotGitToOSFilesystems()
   :241                                            PlainOpenWithOptions()
   :309                                            (*metadataFetcher).getLocalGitMetadata()
   :153                                            (*metadataFetcher).Get()
   :93                                             runMain()
   :43                                             breakdownCmd.func1()
   :856                                            (*Command).execute()
   :974                                            (*Command).ExecuteC()
   :902                                            (*Command).Execute()
   :86                                             Run()
   :34                                             main()
```

hugorut:
Hi @hiran-filho, I've merged a fix for this issue into master, and it will go out with the next release. I'll ping you here when it's available. Thanks.

hiran-filho:
thank @hugorut , I will wait anxiously

hiran-filho:
thank @hugorut,  I will wait anxiously

hugorut:
[...]

PR #1989, opened 2022-08-24
chore: update go-git to use the latest master hash

hugorut:
fixes: https://github.com/infracost/infracost/issues/1987

This change updates the `go-git` module to the latest master snapshot. Which includes fixes for the panics reported by a user.


aliscott (approved):
Thanks!

Issue #1976, opened 2022-08-15
CLI updates for Infracost Cloud with Atlantis, Jenkins and other deployment systems


alikhajeh1:
1. The Infracost CLI needs to support the following env variables as overrides so Atlantis (and other deployment systems) can provide them manually. We already have `INFRACOST_VCS_REPOSITORY_URL`, `INFRACOST_VCS_SUB_PATH`, `INFRACOST_VCS_PULL_REQUEST_URL` and `INFRACOST_TERRAFORM_WORKSPACE`. Once the following is done, all top-level [run metadata](https://www.infracost.io/docs/features/cli_commands/#examples) can be overridden with env vars.
    ```
    === If a user has pull requests

    Required:
    INFRACOST_VCS_PROVIDER # github, gitlab, bitbucket, azure_repos
    INFRACOST_VCS_REPOSITORY_URL
    INFRACOST_VCS_PULL_REQUEST_URL
    INFRACOST_VCS_PULL_REQUEST_AUTHOR
    INFRACOST_VCS_PULL_REQUEST_TITLE

    Optional:
    INFRACOST_VCS_BASE_BRANCH
    INFRACOST_VCS_BRANCH
    INFRACOST_VCS_PIPELINE_RUN_ID
    INFRACOST_VCS_COMMIT_SHA
    INFRACOST_VCS_COMMIT_MESSAGE
    INFRACOST_VCS_COMMIT_TIMESTAMP
    INFRACOST_VCS_COMMIT_AUTHOR_EMAIL
    INFRACOST_VCS_COMMIT_AUTHOR_NAME
    INFRACOST_VCS_SUB_PATH
    INFRACOST_TERRAFORM_WORKSPACE

    === If the user does not have pull requests

    Required:
    INFRACOST_VCS_PROVIDER # github, gitlab, bitbucket, azure_repos
    INFRACOST_VCS_REPOSITORY_URL
    INFRACOST_VCS_BRANCH
    INFRACOST_VCS_COMMIT_SHA # added _SHA
    INFRACOST_VCS_COMMIT_MESSAGE
    INFRACOST_VCS_COMMIT_TIMESTAMP
    INFRACOST_VCS_COMMIT_AUTHOR_EMAIL
    INFRACOST_VCS_COMMIT_AUTHOR_NAME

    Optional:
    INFRACOST_VCS_PIPELINE_RUN_ID
    INFRACOST_VCS_SUB_PATH
[...]

PR #1990, opened 2022-08-24
feat: additional VCS env vars

hugorut:
fixes: https://github.com/infracost/infracost/issues/1976

* Adds support for additional env vars. This means users can specify incomplete metadata in deployment and VCS platforms, which don't supply valid metadata.
* Refactors the metadata package to merge user specified metadata no matter which platform is identified
* Removes TFC metadta fetching method as this is now redundant as all the information will be supplied by env vars
* Additional test coverage for env var merging
* renames `vcsRepoUrl` -> `vcsRepositoryUrl`


aliscott (commented):
This is a great refactor, thanks.

A couple of notes after talking to @alikhajeh1 about this.

1. The Atlantis example we have just uses `INFRACOST_VCS_PULL_REQUEST_URL` and `INFRACOST_VCS_PULL_REQUEST_AUTHOR`. I think we should extract the PR nujmber from the PR URL if `INFRACOST_VCS_PULL_REQUEST_URL` is passed and only override that if `INFRACOST_VCS_PULL_REQUEST_ID` is passed?
2. I'm confused by the difference between `INFRACOST_` and `INFRACOST_VCS_`. I understand some data (repo URL, branch and commit data) is retrieved locally using git, but it can also be set from the VCS provider's env variables/context. I think it's clearer if we're consistent with these as they are all metadata relating to the VCS system/repo.
3. Not got a strong opinion on this but `INFRACOST_VCS_REPO_URL` is more consistent with the term "Repo" we use in Infracost Cloud
4. I think `INFRACOST_VCS_COMMIT_SHA` is clearer than `INFRACOST_VCS_COMMIT`. What do you think?

aliscott:
> 3. Not got a strong opinion on this but INFRACOST_VCS_REPO_URL is more consistent with the term "Repo" we use in Infracost Cloud

After talking to @hugorut I take this back. Changing this would be a breaking change, so let's leave as `INFRACOST_VCS_REPOSITORY_URL`

hugorut:
1. Sounds good
2. Discussed on slack and we'll progress with the `VCS` naming
4. Agreed

hugorut:
@aliscott I've made the changes requested

internal/output/metadata.go
```diff
@@ -10,14 +10,14 @@ import (
...
-	Branch            string    `json:"branch"`
-	Commit            string    `json:"commit"`
-	CommitAuthorName  string    `json:"commitAuthorName"`
-	CommitAuthorEmail string    `json:"commitAuthorEmail"`
-	CommitTimestamp   time.Time `json:"commitTimestamp"`
-	CommitMessage     string    `json:"commitMessage"`
-
-	VCSRepoURL           string `json:"vcsRepoUrl,omitempty"`
+	Branch            string    `json:"vcsBranch"`
+	Commit            string    `json:"vcsCommit"`
```
aliscott, line 14:
```suggestion
	CommitSHA            string    `json:"vcsCommitSha"`
```

aliscott (approved):
One tiny change.

internal/output/metadata.go
hugorut: 👍

PR #1992, opened 2022-08-25
Fix `azurerm_function_app` and  `azurerm_mysql_flexible_server`

aliscott:
Following changes to the Azure Price List API


hugorut (approved):
Nice work

PR #2016, opened 2022-09-01
fix: Fix project name generation from complex URLs

vdmgolub:
My regex-fu is not strong enough, made it in steps but with comments :( For such small strings, it should not affect the speed much, right?


hugorut (commented):
Alrighty, great job, this works. However, I think i'd prefer if we stuck it in the `metadata` pkg and reused `Remote` ultimately the `nameFromRepoURL` just uses the URL built from the `Remote.URL`. If we stick it there it means we can remove all this regex compilation and be a bit more performant.

I just did a quick play around with the metadata `urlStringToRemote` function and I think we can solve these port issues with the following:


```go
// urlStringToRemote returns the provided string as a Remote struct, with the host extracted from the URL.
// urlStringToRemote is designed to work with both HTTPS and SCP remote url strings.
func urlStringToRemote(remote string) Remote {
	if remote == "" {
		return Remote{}
	}

	u, err := url.Parse(remote)
	if err == nil {
		return Remote{
			Host: u.Host,
			URL:  fmt.Sprintf("https://%s%s", u.Host, u.Path),
		}
	}

	logging.Logger.WithError(err).Debugf("parsing remote '%s' as SCP string", remote)

	match := scpSyntax.FindAllStringSubmatch(remote, -1)
	if len(match) == 0 {
		logging.Logger.Debug("remote did not match SCP regexp")
		return Remote{}
	}

	m := match[0]
	if len(m) < 4 {
		logging.Logger.Debug("SCP remote was malformed")
		return Remote{}
	}

	host := m[2]
	path := strings.TrimRight(m[3], "/")
	port := ""

	if _, err := strconv.ParseInt(path, 10, 64); err == nil {
		port = ":" + path
		path = m[4]
	}

	if strings.Contains(host, "azure") {
		host = strings.TrimLeft(m[2], "ssh.")
		path = versionRegxp.ReplaceAllString(path, "")
		pieces := strings.Split(path, "/")
		if len(pieces) == 3 {
			path = strings.Join([]string{pieces[0], pieces[1], "_git", pieces[2]}, "/")
		}
	}

	return Remote{
		Host: host,
		Name: strings.TrimRight(path, ".git"),
		URL: fmt.Sprintf("https://%s%s/%s", host, port, path),
	}
}
```

the added lines are:

```diff
+	path := strings.TrimRight(m[3], "/")
+	port := ""
+
+	if _, err := strconv.ParseInt(path, 10, 64); err == nil {
+		port = ":" + path
+		path = m[4]
+	}
```

and 

```diff
	return Remote{
		Host: host,
+		Name: strings.TrimRight(path, ".git"),
~		URL:  fmt.Sprintf("https://%s%s/%s", host, port, path),
	}
```

I just added a quick test case for the bitbucket url and it seems to work fine:


```go
{
	input: "git@bitbucket.org:8888/~test.infracost.io/infracost-bitbucket-pipeline.git",
	want: Remote{
		Host: "bitbucket.org",
		Name: "~test.infracost.io/infracost-bitbucket-pipeline",
		URL:  "https://bitbucket.org:8888/~test.infracost.io/infracost-bitbucket-pipeline.git",
	},
},
```

```
    --- PASS: Test_urlStringToRemote/git@bitbucket.org:8888/~test.infracost.io/infracost-bitbucket-pipeline.git (0.00s)

```

Doing it this way means we can delete the `nameFromRepoURL` and keep the metadata generation/naming in one pkg which simplifies and makes it easier to understand.

internal/schema/project.go
```diff
@@ -116,17 +116,33 @@ func nameFromRepoURL(rawURL string) string {
...
-		if m[1] == "dev.azure.com" || m[1] == "ssh.dev.azure.com" {
-			n = parseAzureDevOpsRepoPath(m[2])
+	// Removes port if present
+	r = regexp.MustCompile(`:\d+/`)
+	rawURL = r.ReplaceAllString(rawURL, ":")
+
+	// Delimits domain from repo name with `:`
+	if !strings.Contains(rawURL, ":") {
+		rawURL = strings.Replace(rawURL, "/", ":", 1)
+	}
```
hugorut, line 134: If you don't want to do my suggestion then I'd recommend moving the `MustCompile` to the top of the pkg as this means the compilation only happens once, not per function call.

internal/schema/project.go
```diff
@@ -116,17 +116,33 @@ func nameFromRepoURL(rawURL string) string {
...
-	r := regexp.MustCompile(`(?:\w+@|http(?:s)?:\/\/)(?:.*@)?([^:\/]+)[:\/]([^\.]+)`)
-	m := r.FindStringSubmatch(rawURL)
+	// Removes schema and/or credentials
+	r := regexp.MustCompile(`^.*@|http(?:s)?:\/\/`)
+	rawURL = r.ReplaceAllString(rawURL, "")
 
-	if len(m) > 2 {
-		var n = m[2]
+	// Removes extension if present
+	r = regexp.MustCompile(`\.git$`)
```
hugorut, line 124: and this can just be replaced with

```
strings.TrimRight(rawURL, ".git")
```

which is more performant

hugorut:
Actually we'll need to rework the metadata, `urlStringToRemote` no matter what, as without the port parsing will show the remote as `https://bitbucket.org:8888/`

internal/schema/project.go
vdmgolub: Nah, it makes sense to do all this craziness in one place. Thanks for the suggestion 👍

internal/schema/project.go
```diff
@@ -30,16 +29,16 @@ func (m *ProjectMetadata) WorkspaceLabel() string {
...
 }
 
-func (m *ProjectMetadata) GenerateProjectName(repoURL string, dashboardEnabled bool) string {
-	// If the VCS repo is set, create the name from that
-	if repoURL != "" {
-		n := nameFromRepoURL(repoURL)
+func (m *ProjectMetadata) GenerateProjectName(remote *vcs.Remote, dashboardEnabled bool) string {
+	// If the VCS repo is set, use its name
+	if remote != nil && remote.Name != "" {
+		name := remote.Name
```
hugorut, line 35: why are we using a pointer here? It seems fine just to check that the `remote.Name` is not an empty string?

hugorut (approved):
Nice work

internal/schema/project.go
vdmgolub: I was cautious in case the remote is not initialized. But I guess if metadata is present, then remote should always be available. I'll change to pass the value instead of the pointer then 👍

PR #2017, opened 2022-09-02
feat: custom currency format option is provided to for cost formatting

fatihtokus:


aliscott:
Thanks @fatihtokus, I'll assign this to @tim775 since he's familiar with the currency support.

internal/output/format.go
```diff
@@ -23,12 +26,46 @@ func formatCost(currency string, d *decimal.Decimal) string {
 	if d == nil {
 		return "-"
 	}
+
+	if os.Getenv("INFRACOST_CURRENCY_FORMAT") != ""{
```
tim775, line 30: This should be loaded as an [`envconfig`](https://github.com/infracost/infracost/blob/5ea3f9d6de8737d81930c836ea1f0fa60c379c83/internal/config/config.go#L81) like the INFRACOST_CURRENCY env.

internal/output/format.go
```diff
@@ -23,12 +26,46 @@ func formatCost(currency string, d *decimal.Decimal) string {
...
+        return customFormatCost(d)
```
tim775, line 31: Instead of returning a fixed format, setting the custom format should adjust the formatter that is used by the other formatting functions.  (The reason is because this module is used to format prices (which can be very small) and costs which can be large so there isn't really a one-size-fits-all format.)

Maybe the best way to do this is at the start of the output.FormatOutput function use the `money.AddCurrency` to update the formatter for the currently selected currency.

tim775 (changes requested):
Thanks @fatihtokus this is great!  I think it needs a few changes though, let me know if you have questions.

internal/output/format.go
fatihtokus: I made some changes, not sure if that is enough, I have just started coding in go :)

internal/output/format.go
fatihtokus: I have amended it accordingly, please have a look

tim775 (approved):
This is great, thanks so much.  Could you run `make fmt` and make sure you can run `make build` (it's complaining about a few unused imports).  Thanks again!

fatihtokus:
Thanks @tim775, removed unused imports. But couldn't run make build which is failing with following error:
process_begin: CreateProcess(NULL, env bash C:\Users\fatih\IdeaProjects\infracost_fork\infracost\scripts\get_version.sh HEAD, ...) failed.
CGO_ENABLED=0 go build -ldflags="-X 'github.com/infracost/infracost/internal/version.Version='" -v -o build/infracost github.com/infracost/infracost/cmd/infracost
'CGO_ENABLED' is not recognized as an internal or external command,
operable program or batch file.
make: *** [build] Error 1

PR #2037, opened 2022-09-12
feat: add HCL warnings to project metadata

hugorut:
This change introduces a new `warning` field to the project metadata. This is used to report any non-critical run errors to Infracost cloud. The initial use case for this is to capture missing Terraform variable warnings, but additional warnings could be added in the future.

* returns `Module` warnings as part of the evaluation step
* defines a new `Warning` field on `ProjectMetadata` which is populated in `hcl_provider` add metadata
* removes `printWarning` logic in the `hcl_parser` now that we return warnings in the `Module`


internal/hcl/module.go
```diff
@@ -24,6 +29,57 @@ type Module struct {
...
+	}
+}
+
+func joinQuotes(elems []string) string {
+	switch len(elems) {
+	case 0:
+		return ""
+	case 1:
+		return fmt.Sprintf("%q", elems[0])
+	}
```
aliscott, line 75: Why is the `case 1` needed, is it not already handled below?

An alternative would be to loop through and quote the strings and then call `strings.Join(arr, ", ")`.

internal/providers/terraform/testdata/hcl_provider_test/populates_warnings_on_missing_vars/expected.json
```diff
@@ -0,0 +1,74 @@
+{
```
aliscott, line 1: I was expecting to see a warning in this output file, but it doesn't look like there is one.

hugorut: This is plan JSON output (which the hcl parser builds to pass to the json parser) so the warnings don't live in here. I'm testing they come through on the parser here: internal/providers/terraform/hcl_provider_test.go. But I'll also add a breakdown integration test so that we have coverage.

internal/hcl/module.go
hugorut: no, this will panic with the range index, but I'll change to your suggestion using `strings.Join` as it's easier to understand

internal/providers/terraform/testdata/hcl_provider_test/populates_warnings_on_missing_vars/expected.json
aliscott: Ah okay, that makes sense. I think I was thrown off by the `infracost_metadata` we add in here.

internal/hcl/module.go
hugorut: updated in https://github.com/infracost/infracost/pull/2037/commits/82b9d735eeb1f8e806cd8056a715e483061ca76c

internal/providers/terraform/testdata/hcl_provider_test/populates_warnings_on_missing_vars/expected.json
hugorut: added `breakdown` format JSON test in https://github.com/infracost/infracost/pull/2037/commits/020801f203d427bb8be8d2f9de436aaee5420426

aliscott (approved):

PR #2060, opened 2022-09-27
fix: if "Module path" is column is empty it should be hidden

### LINKED ISSUE #1829: Comment showing "Module path" column even though it's empty
Infracost comment is showing the "Module path" column even though it's empty.

Screenshot 2022-07-07 at 10 41 33

**Expected result:**
If "Module path" is column is empty it should be hidden

fatihtokus:
Fixed #1829


aliscott:
@fatihtokus thanks 🙏! Is this ready to be reviewed? Looks like it needs `make fmt` run on it.

fatihtokus:
> @fatihtokus thanks 🙏! Is this ready to be reviewed? Looks like it needs `make fmt` run on
Np, done

fatihtokus:
@aliscott , could you please review the PR

tim775 (changes requested):
Hi Fatih, thanks for the PR.  I updated the test case you wrote so that the project without a metadata path has changes.  This exposes a corner case where the empty "module path" column is still shown.

I like where you're going with your fix, I'm thinking we can update the `calculateMetadataToDisplay` function to ignore projects that are "skipped".  What do you think?

internal/output/markdown.go
```diff
@@ -87,6 +87,10 @@ func calculateMetadataToDisplay(projects []Project) (hasModulePath bool, hasWork
 	// check if any projects that have the same name have different path or workspace
 	for i, p := range sprojects {
 		if i > 0 { // we compare vs the previous item, so skip index 0
+			if p.Diff == nil || len(p.Diff.Resources) == 0 { // ignore the projects that are skipped
+				continue
```
tim775, line 91: I think this doesn't quite cover all cases since when comparing against the 'previous project' (i-1), that could be a project that was supposed to be skipped.   For example, there are two projects with the same name and the first one has no diff, but the second one does, it would still end up as `hasModulePath = true`.

I guess it might be best to filter out the projects without a diff before sorting.

tim775 (approved):

tim775:
Thanks @fatihtokus!

Issue #2003, opened 2022-08-31
`for_each` enumerates properties instead of resource keys


hinaloe:
For resources that use `for_each` to define multiple items, `for_each` will enumerate the resource properties instead of its keys and resources.

## Example

There is a simple sample.

```tf
provider "aws" {
  region = "us-east-1"
}

locals {
  az = {
    a = {
      az         = "us-east-1a"
      cidr_block = "10.0.16.0/24"
    }
    b = {
      az         = "us-east-1b"
      cidr_block = "10.0.32.0/24"
    }
  }
}

resource "aws_vpc" "main" {
  cidr_block       = "10.0.0.0/16"
  instance_tenancy = "default"
}

resource "aws_internet_gateway" "main" {
  vpc_id = aws_vpc.main.id
}

resource "aws_subnet" "public" {
  for_each = local.az

  vpc_id                  = aws_vpc.main.id
  cidr_block              = each.value.cidr_block
  availability_zone       = each.value.az
  map_public_ip_on_launch = true

  tags = {
    Name = "public-subnet-${each.value.az}"
  }

}

resource "aws_nat_gateway" "public_nat" {
  for_each      = aws_subnet.public # 
  subnet_id     = each.value.id
  allocation_id = aws_eip.nat_gateway[each.key].id

  depends_on = [aws_internet_gateway.main]

  tags = {
    Name = "public-nat-${each.value.availability_zone}"
  }
}

resource "aws_eip" "nat_gateway" {
  for_each   = aws_subnet.public
  vpc        = true
  depends_on = [aws_internet_gateway.main]
}
```

### expected result
```
Name                                                      Monthly Qty  Unit              Monthly Cost

 aws_nat_gateway.public_nat["a"]
 ├─ NAT gateway                                                    730  hours                   $32.85
 └─ Data processed                                      Monthly cost depends on usage: $0.045 per GB

 aws_nat_gateway.public_nat["b"]
 ├─ NAT gateway                                                    730  hours                   $32.85
 └─ Data processed                                      Monthly cost depends on usage: $0.045 per GB

OVERALL TOTAL                                                                                   $65.70
```

### actual result
```
$ infracost breakdown --path .
Evaluating Terraform directory at .
  ✔ Downloading Terraform modules
  ✔ Evaluating Terraform directory
  ✔ Retrieving cloud prices to calculate costs

Project: .

 Name                                                      Monthly Qty  Unit              Monthly Cost

 aws_nat_gateway.public_nat["arn"]
[...]

PR #2078, opened 2022-10-05
fix: nested foreach iterating over resource properties

hugorut:
fixes: #2003

* Expands for_each attributes by first wiping the parent blocks values from the context. Then it sets the correct nested object into the context value.
* We now check if the block for_each attribute references another block that is yet to be expanded. If it does, skip the expansion.
* Change the `expandBlock` functionality into a context loop so that blocks that reference each other are eventualy expanded.


tim775 (approved):
🙇

PR #2079, opened 2022-10-06
test: fix intermittent currency format failure

tim775:
@hugorut  I haven't been able to reproduce the FAIL but I'm guessing this will help.


hugorut (approved):
LGTM

PR #2081, opened 2022-10-07
feat: Upload cloud resource ids when usage api is enabled

tim775:
This collects and uploads aws arns to the usage API when it is enabled.  There is additional work needed to collect non-arn aws ids and (eventually) google and azure ids.

### DIFF
--- a/internal/providers/terraform/parser.go
+++ b/internal/providers/terraform/parser.go
@@ -52,6 +52,7 @@ func (p *Parser) createPartialResource(d *schema.ResourceData, u *schema.UsageDa
 					NoPrice:     true,
 					SkipMessage: "Free resource.",
 				},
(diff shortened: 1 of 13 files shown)


internal/providers/terraform/parser.go
```diff
@@ -52,6 +52,7 @@ func (p *Parser) createPartialResource(d *schema.ResourceData, u *schema.UsageDa
 					NoPrice:     true,
 					SkipMessage: "Free resource.",
 				},
+				CloudResourceIds: registryItem.CloudResourceIDFunc(d),
```
aliscott, line 55:
```suggestion
				CloudResourceIDs: registryItem.CloudResourceIDFunc(d),
```

aliscott (approved):
LGTM!

PR #2090, opened 2022-10-14
fix(aws/fsx-windows): no prices for SSD

### LINKED ISSUE #2086: FSX SSD storage unable to be priced
On a recent CI run when pricing an FSX for windows resource with SSD storage the following appeared in the log and the storage cost for the resource was set to 0:

` time="2022-10-12T13:09:06Z" level=warning msg="No products found for module.fsx.aws_fsx_windows_file_system.storage SDD storage, using 0.00"`

When the storage is set to use HDD then the pricing can be retrieved successfully and when testing with the cloud pricing API, a response is returned when using SSD

Presumably the below should be "SSD" rather than "SDD" to retrieve a price: https://github.com/infracost/infracost/blob/367f1fe9c7e8601f5c1b31ba8ec696593569d71b/internal/resources/aws/fsx_windows_file_system.go#L56

jgrumboe:
This will fix #2086


hugorut (approved):
LGTM

PR #2096, opened 2022-10-19
refactor: s3 bucket to CoreResource type

hugorut:


tim775:
This refactor was also done in #2097

PR #2123, opened 2022-11-03
feat: Add ability to override region via ENV var

vdmgolub:


internal/providers/terraform/parser.go
```diff
@@ -327,6 +332,28 @@ func parseTags(resourceType string, v gjson.Result) map[string]string {
...
+
+	switch providerPrefix {
+	case "aws":
+		region = config.AWSOverrideRegion
+	case "azurerm":
+		region = config.AzureOverrideRegion
+	case "google":
+		region = config.GoogleOverrideRegion
+	default:
+		log.Debugf("Unsupported provider %s", providerPrefix)
```
hugorut, line 347:
```suggestion
		logging.Logger.Debugf("Unsupported provider %s", providerPrefix)
```

internal/providers/terraform/parser.go
```diff
@@ -327,6 +332,28 @@ func parseTags(resourceType string, v gjson.Result) map[string]string {
...
+	}
+
+	if region != "" {
+		log.Debugf("Overriding region (%s) for %s", region, addr)
```
hugorut, line 351:
```suggestion
		logging.Logger.Debugf("Overriding region (%s) for %s", region, addr)
```

internal/providers/terraform/parser.go
```diff
@@ -327,6 +332,28 @@ func parseTags(resourceType string, v gjson.Result) map[string]string {
 	}
 }
 
+func overrideRegion(addr string, resourceType string, config *config.Config) string {
+	region := ""
+	providerPrefix := strings.Split(resourceType, "_")[0]
```
hugorut, line 337:
```suggestion
	providerPrefix := strings.Split(resourceType, "_")
	if len(providerPrefix) == 0 {
	    return ""
	}
```

so we don't panic with invalid provider strings (vscode)

vdmgolub: Going to do the same in other methods 👍

hugorut (approved):

PR #2174, opened 2022-12-06
fix: bump max context iterations

hugorut:
fixes: https://github.com/infracost/infracost/issues/1968

* Increases max context iterations to accomodate for complex modules that require further analysis loops to evaluate properly. e.g. https://github.com/Invicton-Labs/terraform-null-deepmerge
* Changes debug function signature to take in a name so that its more useful when debugging


aliscott (approved):
🚀

Issue #1223, opened 2021-12-23
Support for AWS_spot_instance_request


zglin:
Infracost Version: 0.9.15

**Issue in question**

AWS_Spot_Instance_request isn't estimated. Is this working as designed? 

**Current output :**

> 3 cloud resources were detected:
> ∙ 0 were estimated
> ∙ 1 wasn't estimated, report it in https://github.com/infracost/infracost:
>   ∙ 1 x aws_spot_instance_request

**Proposed solution**: Override aws_instance to support spot instances (You could just price it as a full priced instance or take the current spot rates from https://aws.amazon.com/ec2/spot/pricing/)

alikhajeh1:
@zglin thanks for creating this! We can add the mapping for that resource - the [Cloud Pricing API](https://github.com/infracost/cloud-pricing-api) already has AWS EC2 spot instance prices. We might be able to re-use the same logic we have for [aws_instance](https://github.com/infracost/infracost/blob/master/internal/resources/aws/instance.go) (please let us know if you'd like to [work](https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md) on this; we'd be happy to help).

oneslash:
@alikhajeh1 I can try if you don't mind

alikhajeh1:
Hi @oneslash - that would be much appreciated!

aliscott:
@oneslash thanks! Here's some quick thought on how we could support this resource:

[This file](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/aws/instance.go) pulls out the Terraform attributes and constructs an `aws.Instance`. You could handle the `aws_spot_instance_request` in the same way, by extracting the `instance_type` and `ami` attributes and building an `aws.Instance` struct with `PurchaseOption: "spot"`

The question is what we do about the `spot_price`. What do you think the expected behavior should be here? Some ideas I've thought of are:
1. Always show the cost based on the price in the `spot_price` attribute?
2. Ignore the `spot_price` attribute and always show the cost based on the current spot price
3. Show the cost based on the current spot price but cap it at the value of the `spot_price` attribute
4. Show the current spot price, but show this resource as skipped if the current spot price > the `spot_price` attribute.

cc @zglin if you have any thoughts on the behavior of the `spot_price` attribute.

alikhajeh1:
@oneslash in case it helps, our team hangs out in https://www.infracost.io/community-chat too if you have any questions

lucastercas:
[...]

PR #2178, opened 2022-12-08
feat: Add aws spot instance request

lucastercas:
Add support for `aws_spot_instance_request`

Fixes: #1223


hugorut:
Thanks for PR this @lucastercas, I'll take a peek at this when I get a moment in the next few days.

internal/providers/terraform/aws/testdata/spot_instance_request_test/spot_instance_request_test.golden
```diff
@@ -0,0 +1,13 @@
+
+ Name                                             Monthly Qty  Unit   Monthly Cost 
+                                                                                   
+ aws_spot_instance_request.spot_instance_1                                         
+ └─ Instance usage (Linux/UNIX, spot, m3.medium)          730  hours         $4.89 
+                                                                                   
+ aws_spot_instance_request.spot_instance_2                                         
+ └─ Instance usage (Linux/UNIX, spot, m3.large)           730  hours        $22.85
```
hugorut, line 8: Sadly this method won't work, as the spot prices change often and will cause the tests to fail. Instead you'll need to test against the price hash. See here: https://github.com/infracost/infracost/blob/59b89e4f850b9f9f1e7930cb2e191549555478c0/internal/providers/terraform/aws/autoscaling_group_test.go#L23 as an example of how to do this.

hugorut (changes requested):
@lucastercas this looks great, thanks. However, we just need to change the way we test the output price. A golden file test won't work unfortunately. Thanks

internal/providers/terraform/aws/testdata/spot_instance_request_test/spot_instance_request_test.golden
lucastercas: Yeah, no problem. 
I was just waiting for the github pipeline to go back so I can continue, thank you for the help!

hugorut: done

internal/providers/terraform/aws/testdata/spot_instance_request_test/spot_instance_request_test.golden
```diff
@@ -0,0 +1,13 @@
+
```
hugorut, line 1: We can just remove this directory now - given we're using the hash tests

hugorut (commented):
Seems like the tests are failing as you just need to add in a check for the `t3_large` spot resource you've got in your tests

internal/providers/terraform/aws/testdata/spot_instance_request_test/spot_instance_request_test.tf
```diff
@@ -0,0 +1,22 @@
+provider "aws" {
```
hugorut, line 1: this file can be removed

internal/providers/terraform/aws/testdata/spot_instance_request_test/spot_instance_request_test.usage.yml
```diff
@@ -0,0 +1,2 @@
+version: 0.1
```
hugorut, line 1: this file can be removed

alikhajeh1:
Any update on this?

tim775 (approved):

hugorut (approved):

PR #2182, opened 2022-12-12
feat: track git changes in project metadata

hugorut:
Adds support to track git changes for HCL projects. If changes can be detected, we set the `vcsCodeChanged` project metadata field on any projects that have been modified. This behaviour is works with the `show-changed` flag that `infracost comment` supports.

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -68,8 +68,8 @@ func addRunFlags(cmd *cobra.Command) {
 
 	cmd.Flags().StringSlice("exclude-path", nil, "Paths of directories to exclude, glob patterns need quotes")
 	cmd.Flags().Bool("include-all-paths", false, "Set project auto-detection to use all subdirectories in given path")
-	cmd.Flags().String("git-diff", "master", "Show only costs that have git changes compared to the provided branch. Use the name of the current branch to fetch changes from the last two commits")
-	_ = cmd.Flags().MarkHidden("git-diff")
+	cmd.Flags().String("git-diff-target", "master", "Show only costs that have git changes compared to the provided branch. Use the name of the current branch to fetch changes from the last two commits")
+	_ = cmd.Flags().MarkHidden("git-diff-target")
 
 	cmd.Flags().Bool("no-cache", false, "Don't attempt to cache Terraform plans")
 
@@ -93,7 +93,7 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 	}
 
 	repoPath := runCtx.Config.RepoPath()
-	metadata, err := vcs.MetadataFetcher.Get(repoPath, runCtx.Config.ChangeTarget)
+	metadata, err := vcs.MetadataFetcher.Get(repoPath, runCtx.Config.GitDiffTarget)
 	if err != nil {
 		logging.Logger.WithError(err).Debugf("failed to fetch vcs metadata for path %s", repoPath)
 	}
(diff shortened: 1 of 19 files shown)


internal/config/config.go
```diff
@@ -148,7 +149,7 @@ func DefaultConfig() *Config {
 // RepoPath returns the filepath to either the config-file location or initial path provided by the user.
 func (c *Config) RepoPath() string {
 	if c.ConfigFilePath != "" {
-		return c.ConfigFilePath
+		return strings.TrimRight(c.ConfigFilePath, filepath.Base(c.ConfigFilePath))
```
aliscott, line 152: Why do we need this? We should also update the docs for this function since they seem inconsistent now.

internal/config/config.go
```diff
@@ -148,7 +149,7 @@ func DefaultConfig() *Config {
 // RepoPath returns the filepath to either the config-file location or initial path provided by the user.
```
hugorut, line 149:
```suggestion
// RepoPath returns the file path to either the config-file directory or initial path provided by the user.
```

cmd/infracost/run.go
```diff
@@ -68,6 +68,8 @@ func addRunFlags(cmd *cobra.Command) {
 
 	cmd.Flags().StringSlice("exclude-path", nil, "Paths of directories to exclude, glob patterns need quotes")
 	cmd.Flags().Bool("include-all-paths", false, "Set project auto-detection to use all subdirectories in given path")
+	cmd.Flags().String("git-diff", "master", "Show only costs that have git changes compared to the provided branch. Use the name of the current branch to fetch changes from the last two commits")
```
aliscott, line 71: It's not clear from the name what this is. I'd consider something like `git-base-ref`. I realise it only supports branches at the moment, but I'm assuming we will support other references later?

internal/vcs/metadata.go
```diff
@@ -534,12 +541,88 @@ func (f *metadataFetcher) getLocalGitMetadata(path string) (Metadata, error) {
...
+		}
+	} else {
+		commitIter, err := r.Log(&git.LogOptions{Order: git.LogOrderCommitterTime})
+		if err != nil {
+			logging.Logger.WithError(err).Debugf("could not get local git file changes, could not call git log using currentCommit hash %s", currentCommit.Hash)
+			return nil
+		}
+
+		// call Next() twice to get the commit after the currentCommit.
+		_, _ = commitIter.Next()
```
aliscott, line 582: Is an error case here possible?

internal/config/config.go
```diff
@@ -100,6 +100,7 @@ type Config struct {
 	SyncUsageFile   bool       `yaml:"sync_usage_file,omitempty" ignored:"true"`
 	Fields          []string   `yaml:"fields,omitempty" ignored:"true"`
 	CompareTo       string
+	ChangeTarget    *string
```
aliscott, line 103: It seems we have `git-diff`, `ChangeTarget` and `ChangeBase` which (I think) all mean the same thing. Could we make the naming more consistent?

aliscott (approved):
Nice!

cmd/infracost/run.go
hugorut: yes that makes sense, and yes we'd support refs later

internal/config/config.go
hugorut: yeah, I think these all got mixed up when I was playing around with naming. I think `GitDiffTarget` is the best. So i'll change to that

PR #2183, opened 2022-12-12
enhance: add cache for Terraform source files.

aliscott:
This means that Terragrunt doesn't need to redownload the source files multiple times if the same source is specified

### DIFF
--- a/internal/providers/terraform/terragrunt_hcl_provider.go
+++ b/internal/providers/terraform/terragrunt_hcl_provider.go
@@ -24,6 +24,7 @@ import (
 	hcl2 "github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
 	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/otiai10/copy"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/zclconf/go-cty/cty"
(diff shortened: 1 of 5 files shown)


internal/providers/terraform/terragrunt_hcl_provider.go
```diff
@@ -649,16 +660,41 @@ func (p *TerragruntHCLProvider) downloadTerraformSource(source string, terragrun
...
+	// 1. Terraform lock files - these are normally committed to source control
+	// 2. .terragrunt-source-version files - these are used to determine if the source has changed
+	// 3. .infracost dir - this contains any cached third party modules. We can remove this when we move this directory to the root path
+	opt := copy.Options{
+		Skip: func(src string) (bool, error) {
+			if filepath.Base(src) == util.TerraformLockFile || filepath.Base(src) == terragruntSourceVersionFile || filepath.Base(src) == config.InfracostDir {
+				return false, nil
+			}
+
+			return strings.HasPrefix(filepath.Base(src), "."), nil
```
hugorut, line 680:
```suggestion
            base := filepath.Base(src)
			if base == util.TerraformLockFile || base == terragruntSourceVersionFile || base == config.InfracostDir {
				return false, nil
			}

			return strings.HasPrefix(base, "."), nil
```

hugorut (approved):
LGTM. Are we going to address the Terraform redownload of modules in a multi-project setting in another PR?

aliscott:
> Are we going to address the Terraform redownload of modules in a multi-project setting in another PR?

I think that is a separate optimization we can make, but not in scope of this PR. We want that optimization to work for non-Terragrunt projects as well, and will need it to work with the parallelisation as well.

hugorut:
@aliscott yup, sounds good. Just wanted to clarify if we're waiting on that for the `0.10.15` release or not, which we shouldn't.

PR #2208, opened 2023-01-02
feat: change module download directory to reside at context root

hugorut:
This PR changes the module download and caching functionality to reside at the top level of multi-project repositories. This departs from the old logic of using a .infracost folder in each sub-project. This change is significant as it departs from Terraform logic in favour of speed and efficiency. This change should speed up most Infracost runs by some measure, as we will no longer need to download multiple copies of modules per project or per multi-project context.

I've also added concurrency to the module loader, to further boost download performance.


internal/hcl/modules/loader.go
```diff
@@ -72,38 +71,63 @@ func NewModuleLoader(path string, credentialsSource *CredentialsSource, logger *
...
+				// we'll look up the modules based on the cache path at the Infracost root (where the infracost.yml
+				// resides or where the --path autodetect started for multi-project)
+				for i, module := range manifest.Modules {
+					dir := path
+					if m.cachePath != "" {
+						dir, _ = filepath.Rel(m.cachePath, path)
+					}
+
+					manifest.Modules[i].Dir = filepath.Join(dir, module.Dir)
+				}
```
hugorut, line 129: I wasn't sure here if we should copy these to the `.infracost` directory instead. I've left that out for now. But I think we should probably migrate away from relying on the `.terraform` directory and instead do a conversion step when we first find it.

hugorut:
The test failures look unrelated to these changes, so please ignore

aliscott (commented):
1. Add `version` key to manifest.json instead of manifest.v2.json, similar to the state file and config file.
2. Add `project` key to manifest's modules instead of separate manifest files.

hugorut:
As discussed 

1. I've changed the version to reside in the file https://github.com/infracost/infracost/pull/2208/commits/c0a95d8f55fcc4b63b1afe717ffa39d8df4bd2c6
2. we're leaving the manifest per project change, as it's a bit too involved to have a singular manifest.json

internal/hcl/modules/loader.go
```diff
@@ -72,38 +73,63 @@ func NewModuleLoader(path string, credentialsSource *CredentialsSource, logger *
...
-func (m *ModuleLoader) manifestFilePath() string {
-	return filepath.Join(m.Path, manifestPath)
+func (m *ModuleLoader) manifestFilePath(projectPath string) string {
+	if m.cachePath == projectPath {
+		return filepath.Join(m.cachePath, manifestPath)
+	}
+
+	rel, _ := filepath.Rel(m.cachePath, projectPath)
+	sum := md5.Sum([]byte(rel)) //nolint
+	return filepath.Join(m.cachePath, ".infracost/terraform_modules/", fmt.Sprintf("%x.json", sum))
```
aliscott, line 87:
```suggestion
	return filepath.Join(m.cachePath, ".infracost/terraform_modules/", fmt.Sprintf("manifest-%x.json", sum))
```
Should we prefix this with `manifest-` so it's clear it's a manifest file.

aliscott (approved):
Awesome!

internal/hcl/modules/loader.go
hugorut: good idea

PR #2209, opened 2023-01-03
feat: run auto-detected projects in parallel

hugorut:
Changes the auto-detection functionality to run discovered projects in parallel. This is faster and inline with the `config-file` behaviour.


aliscott (approved):
LGTM

PR #2212, opened 2023-01-06
fix: default to the correct default provider based on the resource type

aliscott:


aliscott:
Test failures are unrelated. Will look into them separately.

hugorut (approved):

PR #2220, opened 2023-01-11
fix: fix object variables passed with a subset of keys

aliscott:


internal/hcl/evaluator.go
```diff
@@ -524,61 +523,32 @@ func (e *Evaluator) evaluateVariable(b *Block) (cty.Value, error) {
...
-			return def.Value(), nil
-		}
-		return convert.Convert(def.Value(), ty)
+		return e.convertType(b, def.Value(), attrType)
 	}
 
-	return convertType(cty.DynamicVal, attrType), errorNoVarValue
-}
-
-func convertType(val cty.Value, attribute *Attribute) cty.Value {
```
aliscott, line 546: From what I can tell, we don't need this logic any more.

hugorut (approved):

PR #2225, opened 2023-01-16
refactor: scope errors by project

hugorut:
Changes the Terraform and Terragrunt providers to scope evaluation and execution errors to projects, rather than failing the whole run. We now set a metadata field on the project to indicate if the project encountered errors and then continue to evaluate other detected or configured projects. Once the run has finished and piped to an output we indicate that certain projects have failed, and display errors.

This change means that pipeline runs won't fail in a multi-project setting if only a select few projects have failed. If all project error, we bubble the errors up the stack and fail the cmd.


internal/output/table.go
```diff
@@ -51,14 +51,31 @@ func ToTable(out Root, opts Options) ([]byte, error) {
 
 		s += "\n"
 
-		tableOut := tableForBreakdown(out.Currency, *project.Breakdown, opts.Fields, includeProjectTotals)
+		if len(project.Metadata.Errors) > 0 {
```
aliscott, line 54: Should we have the same logic in the diff output? The other option would be something like:
```
The following projects could not be evaluated:
infracost/infracost/cmd/infracost/testdata/breakdown_multi_project_with_error/dev (Module path: dev)
Run the following command to see more details: infracost breakdown --path=/path/to/code
```

internal/output/markdown.go
```diff
@@ -207,6 +211,13 @@ func ToMarkdown(out Root, opts Options, markdownOpts MarkdownOptions) ([]byte, e
 		}
 	}
 
+	erroredProjectCount := 0
```
aliscott, line 214: Should we exclude anything with `len(p.Metadata.Errors) == 0` from the skipped project count so the comment just shows:

`1 project could not be evaluated`.

aliscott (commented):
Love reading your code.

internal/output/table.go
hugorut: sounds good

internal/output/markdown.go
hugorut: 👍

internal/providers/terraform/hcl_provider.go
```diff
@@ -309,61 +316,57 @@ func (p *HCLProvider) Modules() ([]*hcl.Module, error) {
...
-	if err != nil {
-		return nil, err
-	}
+	wg.Wait()
 
 	if p.config.CacheParsingModules {
 		p.cache = mods
 	}
 
+	// TODO: handle nil modules from error return
```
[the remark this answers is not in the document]
hugorut: Yeah I forgot to remove this line - we shouldn't ever hit this as the default return from `ParseDirectory` is a safe default `Module`

aliscott (approved):

PR #2228, opened 2023-01-17
fix: project errors markdown output

hugorut:
resolves:

* The extra project counts lines are now one comma separated line
* Fixes errored projects having a diff meaning they are displayed in the table


aliscott (approved):
LGTM!

PR #2232, opened 2023-01-19
fix: module for_each attributes not expanding correctly

hugorut:
Resolves various issue within the Evaluator which was causing module for_each attributes to not be expanded correctly.

* Solves issue where for_each that references a Block that doesn't need to be expanded was never getting expanded as we were always checking if Block.expanded was true even if the Block had no count/for_each
* When modules are expanded we now set the EvaluationContext to have the correct Variables. Prior to this change expanded modules were being set with their expansion key in the name. e.g:
  ```
  module "foo" {
    source = "../foo"
  
    for_each = {
       "one" = {
           "bar" = "bat"
       }
       "two" = {
          ...
       }
    }
    bar      = each.value.bar
  }
  ```
  would have been set in the context to:
  ```
  module
     foo
     foo[one]
       bar: bat
     foo[two]
       ...
  ```
  instead of of the correct format:
  
  ```
  module
     foo
       one
         bar: bat
       two
         ...
  ```
* Removes logic that duplicates expansion of modules. We expand modules before the `loadModules` call in Evaluator, but fail to override the `Module.Blocks` with these expanded Blocks. This means that we expand these module outputs twice which can lead to incorrect EvaluationContext, e.g:

  ```
  module
     foo
       one
         bar: bat
         # after duplicate expansion
         one
            bar: bat
         two
            ...
       two
         ...
  ```

### DIFF
--- a/internal/hcl/block.go
+++ b/internal/hcl/block.go
@@ -98,7 +98,13 @@ func (b referencedBlocks) Less(i, j int) bool {
 	return false
 }
 
-// ModuleBlocks returns all the Blocks of type module. The returned Blocks
+// ModuleBlocks is a wrapper around SortedByCaller that selects just Modules to be sorted.
+func (blocks Blocks) ModuleBlocks() Blocks {
+	justModules := blocks.OfType("module")
+	return justModules.SortedByCaller()
+}
+
+// SortedByCaller returns all the Blocks of type module. The returned Blocks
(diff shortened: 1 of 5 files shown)


hugorut:
Test failures are unrelated - they are happening on master

aliscott (approved):

PR #2241, opened 2023-01-25
fix: reload ModuleCall when `each.value` changes

hugorut:
When expanding a module we pass the `each.value` to it's definition block to set as initial vars in the child `Evaluator`. If the module `for_each` depends on a value that is yet to be known in the evaluation `Context` then the `each.value` will be incomplete.

This PR fixes this issue by:
1. Checking if `Blocks` in the evaluation loop `HaveChanged` to their `for_each` attribute. If they do then we expand these blocks again, setting the lower level `each.value` to the correct value.
3. We now keep track of `ModuleCalls` as a map, which allows us to lookup if the changed `Blocks` correspond to a loaded `ModuleCall`. If they do we reset the `Definition` block to the newly expanded one with the correct value.
4. This means that in `evaluateModules` call in the `Evaluator` that we see that the input vars have changed and reload the module

It also changes the `Registry` `GetDefaultRefIDFunc` behaviour to fallback to generating a reference from `id` this means unsupported resources can still be indexed as references.

### DIFF
--- a/internal/hcl/attribute.go
+++ b/internal/hcl/attribute.go
@@ -44,7 +44,8 @@ type Attribute struct {
 	Verbose bool
 	Logger  *logrus.Entry
 	// newMock generates a mock value for the attribute if it's value is missing.
-	newMock func(attr *Attribute) cty.Value
+	newMock       func(attr *Attribute) cty.Value
+	previousValue cty.Value
 }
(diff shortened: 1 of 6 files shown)


tim775 (approved):
🤯

Issue #2242, opened 2023-01-25
Update price mapping for Google Storage Bucket Class A/Class B operations


vdmgolub:
Google is updating pricing for the operations. They will take effect on Apr 1st 2023: https://cloud.google.com/storage/pricing-announce-apr23#sku-mapping.

Currently, the cost components support existing single pricing. Closer to the deadline, we will need to update [these mappings](https://github.com/infracost/infracost/blob/master/internal/resources/google/storage_bucket.go#L159-L193) to support regional/multi-region/dual-region variations.

The location should probably specify this variation, like [this block does](https://github.com/infracost/infracost/blob/master/internal/resources/google/storage_bucket.go#L84-L94), but this needs to be tested.

PR #2243, opened 2023-01-25
fix: Fix tests for cloud resources

vdmgolub:
See comment messages for detailed descriptions:

1. AWS - new SKUs were matched by the price filter.
2. Google - new SKUs with new pricing. Fix makes the cost component to use the current pricing, but we'll need to update it closer to Apr 1st. Here's the issue to track the progress: #2242.
3. Azure - attribute changed the name.


hugorut (approved):
LGTM

PR #2245, opened 2023-01-26
fix: adds infracost_resource_changes to fix policy api

hugorut:
Reworks the `resource_changes` from the HCLProvider to become `infracost_resource_changes` so that diff is ignored. We then rewrite this as `resource_changes` when sending to the Policy API so that this is consistent with a standard Plan JSON output.


hugorut:
Note - this does not fix the resource override bug that @vdmgolub found. I have changes inbound for that.

aliscott (approved):
🙌

PR #2246, opened 2023-01-26
fix: copy CoreResource before costing policy

hugorut:
Resolves issue where policies that set attributes on CoreResources was affecting the Infracost output. This was because the underlying CoreResource was a pointer. So setting struct fields would alter the resource when before fetching the prices.


aliscott (approved):
LGTM!

PR #2251, opened 2023-01-27
feat(installer): Allow installer to download any version

sylwit:
This PR allows the installer to download a specific version of infracost based on the INFRACOST_VERSION variable. Default to latest

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -19,9 +19,10 @@ if [ "$arch" = "aarch64" ]; then
   arch="arm64"
 fi
 
-url="https://infracost.io/downloads/latest"
+version=${INFRACOST_VERSION:-latest}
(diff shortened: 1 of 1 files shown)


sylwit:
@tim775 following our discussion on Slack

tim775:
Makes sense to me but we'll need to update our redirect rules since they don't support individual versions right now.

@aliscott I think we can just add a redirect from `/downloads/<*>` -> `https://github.com/infracost/infracost/releases/download/<*>` at the end of the list, right?

aliscott:
> @aliscott I think we can just add a redirect from `/downloads/<*>` -> `https://github.com/infracost/infracost/releases/download/<*>` at the end of the list, right?

Yep, looks like this should work okay.

PR #2266, opened 2023-02-01
fix: don't query resources twice if past resources are the same as current resources

aliscott:
This prevents the warning logs from being outputted twice.

We could improve this further by checking the existence of each resource individually, but for now this only handles the case where there is no diff.

### DIFF
--- a/internal/providers/terraform/parser.go
+++ b/internal/providers/terraform/parser.go
@@ -1,6 +1,8 @@
 package terraform
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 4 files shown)


internal/providers/terraform/parser.go
```diff
@@ -831,3 +843,23 @@ func parseKnownModuleRefs(resData map[string]*schema.ResourceData, conf gjson.Re
 		}
 	}
 }
+
+func gjsonEqual(a, b gjson.Result) bool {
+	var err error
+
+	var aOut bytes.Buffer
+	err = json.Indent(&aOut, []byte(a.Raw), "", "")
```
hugorut, line 851:
```suggestion
	err = json.Compact(&aOut, []byte(a.Raw))
```

internal/providers/terraform/parser.go
```diff
@@ -170,6 +172,16 @@ func (p *Parser) parseJSON(j []byte, usage map[string]*schema.UsageData) ([]*sch
...
 	}
 
+	if parsed.Get("prior_state").Exists() == false {
+		return nil, resources, nil
+	}
+
+	// Check if the prior state is the same as the planned state
+	// and if so we can just return pointers to the same resources
+	if gjsonEqual(parsed.Get("prior_state.values.root_module"), parsed.Get("planned_values.root_module")) {
+		return resources, resources, nil
```
hugorut, line 182: might get messy in the future if we manipulate the resource pointer, but I guess we can copy the resources when it becomes a problem

hugorut (approved):

internal/providers/terraform/parser.go
aliscott: Oooh, nice

Issue #2138, opened 2022-11-16
AWS Lambda - missing cost configurations


robh007:
Currently AWS Lambda supports both x86 & Arm architectures. They have different pricing options, also lambda now supports storage options & this too is missing from the `lambda_function` resource. Concurrency is also not supported within the `aws_lambda_provisioned_concurrency_config` tf resource.

robh007:
@alikhajeh1 @aliscott @tim775 I'll pick this one up.

alikhajeh1:
Thanks @robh007! 

BTW, when running `infracost breakdown --path . --usage-file infracost-usage.yml` with the following main.tf + usage-file, I noticed the output is not showing request costs despite the usage value `monthly_requests` being set. If you get the time, it would be awesome if you can check that issue too as it's in the lambda resource logic.


```
# === main.tf
provider "aws" {
  region                      = "us-east-1" # <<<<< Try changing this to eu-west-1 to compare the costs
  skip_credentials_validation = true
  skip_requesting_account_id  = true
  access_key                  = "mock_access_key"
  secret_key                  = "mock_secret_key"
}

resource "aws_lambda_function" "hello_world" {
  function_name = "hello_world"
  role          = "arn:aws:lambda:us-east-1:account-id:resource-id"
  handler       = "exports.test"
  runtime       = "nodejs12.x"
  filename      = "function.zip"
  memory_size   = 1024 # <<<<< Try changing this to 512 to compare costs
}


# === Usage file

version: 0.1
resource_usage:
  aws_lambda_function.hello_world:
    monthly_requests: 100000000 # Monthly requests to the Lambda function.
    request_duration_ms: 250    # Try changing this to 100 (milliseconds) to compare costs


# === Current output

aws_lambda_function.hello_world
 ├─ Requests                                            Monthly cost depends on usage: $0.20 per 1M requests
 └─ Duration (first 6B)                                       25,000,000  GB-seconds                 $416.67
```

PR #2269, opened 2023-02-02
Feat: (AWS) Adds support for ARM architectures

robh007:
Lambda resource would always provide cost estimates for x86 only. ARM architecture as a different cost tiers. Also fixed a bug where request usage wasn't being set at all. So whatever you set for requests it would never report the cost. 

Fixes part of #2138


aliscott:
@robh007 thanks 🙏 , Can you try regenerating the golden file tests? It looks like the sync usage functionality is messing with it, so maybe try in a shell that doesn't have any `AWS_PROFILE` exported.

robh007:
I've ran the `make test_update_usage`, not sure if these are required but the azure files have updated for some reason.

robh007:
@aliscott not sure what's going on here. `TestWAFWebACLGoldenFile` Looks to be failing, I've ran the updated the AWS golden files too.

aliscott (approved):

aliscott:
@robh007, thanks 🙏 ! I've looked through this and the test failures and it all seems unrelated. Will look into fixing up all the tests on master.

PR #2278, opened 2023-02-06
fix(azure): fix Application Gateway v2 costs

### LINKED ISSUE #2277: Wrong pricing azure application gateway v2
Two problems here:

1) pricing of CU is completely off, this is calculated per hour, but should be calculated per month
2) when you do not set capacity on the sku (optional if you use autoscaling), there is no price for the application gateway, it seems that the capacity is multiplying the usage cost (so capacity=2 means 1460 hours of application gateway costs) , but this should impact the CU cost of (1)

aliscott:
* Capacity units are priced hourly not monthly
* For v2 Capacity units shouldn't affect the fixed costs

Fixes #2277


internal/providers/terraform/azure/application_gateway.go
```diff
@@ -20,13 +20,16 @@ func GetAzureRMApplicationGatewayRegistryItem() *schema.RegistryItem {
...
 	region := lookupRegion(d, []string{})
-	var monthlyDataProcessedGb, monthlyCapacityUnits *decimal.Decimal
+	var monthlyDataProcessedGb *decimal.Decimal
 	skuName := d.Get("sku.0.name").String()
 	var sku, tier string
 	costComponents := make([]*schema.CostComponent, 0)
 	tierLimits := []int{10240, 30720}
 
-	capacity := d.Get("sku.0.capacity").Int()
+	capacityUnits := d.Get("sku.0.capacity").Int()
```
vdmgolub, line 29: Should this take precedence over usage value when it's > 0? It's optional for autoscaling, so that's where the usage value can be handy. However, if this is set explicitly (for example = 2), can it go over to the usage value?

Also, when the capacity attribute is missing and the usage value is not set, should we default `capacityUnits` to 1? For this case, we need another example with `autoscale_configuration` [block](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/application_gateway#autoscale_configuration).

aliscott: Yep, updated to be that precedence, and also use the `autoscale_configuration > min_capacity` if nothing else is provided.

internal/providers/terraform/azure/application_gateway.go
```diff
@@ -20,13 +20,25 @@ func GetAzureRMApplicationGatewayRegistryItem() *schema.RegistryItem {
...
 
-	capacity := d.Get("sku.0.capacity").Int()
+	var capacityUnits int64 = 1
+
+	if d.Get("autoscale_configuration.0.min_capacity").Type != gjson.Null {
+		capacityUnits = d.Get("autoscale_configuration.0.min_capacity").Int()
+	}
+
+	if u != nil && u.Get("capacity_units").Type != gjson.Null {
+		capacityUnits = u.Get("capacity_units").Int()
```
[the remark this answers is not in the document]
aliscott: I think if the user sets the usage file value it should override in this case.

PR #2281, opened 2023-02-07
Feat:(AWS) Adds Lambda provisioned concurrency

robh007:
Adds new resource `aws_lambda_provisioned_concurrency_config` to support provisioned concurrency in lambda. I think there's potentially more that can be done to merge with `aws_lambda_function`. I did try to use the references logic within the code but couldn't get it to work. This resource provides 2 cost components, duration & concurrency. 

Closes more of #2138

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -262,6 +262,13 @@ resource_usage:
   aws_lambda_function.my_function:
     monthly_requests: 100000 # Monthly requests to the Lambda function.
     request_duration_ms: 500 # Average duration of each request in milliseconds.
+  
+  aws_lambda_provisioned_concurrency_config.my_config:
+    monthly_duration_hrs: 100 # Number of hours in a month that provisioned concurrency will be enabled.
+    request_duration_ms: 350 # Average duration of each request in milliseconds during the enabled period.
+    monthly_requests: 10000000 # Number of requests sent to the function during the enabled period.
+    architecture: arm64 # Architecture of the Lambda function.
+    memory_mb: 512 # Memory size of the Lambda function.
 
   aws_alb.my_alb:
     new_connections: 10000    # Number of newly established connections per second on average.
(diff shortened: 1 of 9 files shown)


aliscott (commented):
Hi @robh007 thanks this looks excellent. I had a look through the pricing and agree with this approach and not using the references.

In the AWS calculator I'm also seeing $2 'Requests' charge for the example you have in the tests, so it looks like we need another cost component for that.
Screenshot 2023-02-10 at 11 54 30

robh007:
> Hi @robh007 thanks this looks excellent. I had a look through the pricing and agree with this approach and not using the references.
> 
> In the AWS calculator I'm also seeing $2 'Requests' charge for the example you have in the tests, so it looks like we need another cost component for that. Screenshot 2023-02-10 at 11 54 30

Hi @aliscott,

Yes that cost component is requests, the requests are needed to generate the Duration  cost component. I can add that in no problem. I probably need to document something somewhere, because if you had something like this in your code.

```hcl
resource "aws_lambda_provisioned_concurrency_config" "without_usage" {
  function_name                     = aws_lambda_function.name
  provisioned_concurrent_executions = 100
  qualifier                         = 1
}

resource "aws_lambda_function" "lambda" {
  function_name = "lambda_function_name"
  role          = "arn:aws:lambda:us-east-1:account-id:resource-id"
  handler       = "exports.test"
  filename      = "function.zip"
  runtime       = "nodejs12.x"
}

```

With these usage parameters

```yml
  aws_lambda_function.my_function:
    monthly_requests: 10000000 # Monthly requests to the Lambda function.
    request_duration_ms: 350 # Average duration of each request in milliseconds.
  
  aws_lambda_provisioned_concurrency_config.my_config:
    monthly_duration_hrs: 730 # Number of hours in a month that provisioned concurrency will be enabled.
    request_duration_ms: 350 # Average duration of each request in milliseconds during the enabled period.
    monthly_requests: 10000000 # Number of requests sent to the function during the enabled period.
    architecture: arm64 # Architecture of the Lambda function.
    memory_mb: 512 # Memory size of the Lambda function.
```

The estimates for requests are duplicated, I think that's why I wanted to use something with the reference. It probably just needs to highlight the differences somewhere for now. Thoughts?

aliscott (approved):
Sorry for the delay @robh007. This looks great, thanks!

PR #2285, opened 2023-02-08
Enable use of infracost breakdown with ARM whatif file

Yeseh:
Only azure_app_service_plan works for now


Yeseh:
Woops, misclicked please ignore :)

PR #2286, opened 2023-02-08
feat: wildcard support for modules in usage file

hugorut:
This PR adds support for wildcard references in the Infracost usage file.

Users can now specify usage file references such as `module.mod[*].aws_lambda_function.test[*]`.

This functionality is enabled by the introduction of `schema.UsageMap` into the application. This builds a regexp from the usage file name (escaping all regular expression metacharacters apart from *) if a direct match is not made. This means we can support many layers of wildcard references, e.g: `module.mod[*].module.another[*].aws_lambda_function.test[*]`.

### DIFF
--- a/cmd/infracost/testdata/breakdown_terraform_sync_usage_file/breakdown_terraform_sync_usage_file.golden
+++ b/cmd/infracost/testdata/breakdown_terraform_sync_usage_file/breakdown_terraform_sync_usage_file.golden
@@ -8,7 +8,7 @@ Project: infracost/infracost/cmd/infracost/testdata/breakdown_terraform_sync_usa
  └─ Insights queries data scanned                    Monthly cost depends on usage: $0.005 per GB   
                                                                                                     
  aws_cloudwatch_log_group.production_logs["media"]                                                  
- ├─ Data ingested                                    Monthly cost depends on usage: $0.50 per GB    
+ ├─ Data ingested                                             1,000  GB                     $500.00 
  ├─ Archival Storage                                          1,000  GB                      $30.00 
  └─ Insights queries data scanned                    Monthly cost depends on usage: $0.005 per GB   
                                                                                                     
@@ -28,7 +28,7 @@ Project: infracost/infracost/cmd/infracost/testdata/breakdown_terraform_sync_usa
  ├─ NAT gateway                                                 730  hours                   $32.85 
  └─ Data processed                                                2  GB                       $0.09 
                                                                                                     
- OVERALL TOTAL                                                                              $251.12 
+ OVERALL TOTAL                                                                              $751.12 
 ──────────────────────────────────
 6 cloud resources were detected:
 ∙ 6 were estimated, all of which include usage-based costs, see https://infracost.io/usage-file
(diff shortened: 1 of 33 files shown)


hugorut:
@aliscott I'm not sure what we're thinking of doing with the `sync` functionality and if these wildcard changes need to change that as well. For now I've not touched it.

internal/schema/usage_data.go
```diff
@@ -149,8 +151,64 @@ func (u *UsageData) CalcEstimationSummary() map[string]bool {
...
+	if err == nil {
+		val, ok := usage[parsedAddress.ResourceSpec.Type]
+		if ok {
+			data = val
+		}
+	}
+
+	if ud := usage[address]; ud != nil {
+		if data != nil {
+			MergeAttributes(data, ud)
```
hugorut, line 172: It seems from the code that I extracted this from that we were only ever merging attributes from the `type` spec if this was a direct hit. I'm not sure if we should keep this behaviour, or merge the usage data if we also get a hit on the wildcard matches. I think we probably should, but I'm unsure why this wasn't done in the first place?

internal/schema/usage_data.go
```diff
@@ -149,8 +151,64 @@ func (u *UsageData) CalcEstimationSummary() map[string]bool {
...
+		return ud
+	}
+
+	if strings.HasSuffix(address, "]") {
+		lastIndexOfOpenBracket := strings.LastIndex(address, "[")
+
+		if arrayUsageData := usage[fmt.Sprintf("%s[*]", address[:lastIndexOfOpenBracket])]; arrayUsageData != nil {
+			return arrayUsageData
+		}
+	}
```
hugorut, line 184: technically we don't "need" this now with the regxp matching below, however this is more performant lookup. So I'd prefer to keep it, so we catch things before moving to the matching functionality. However, I'm ok to remove as well.

internal/schema/usage_data.go
hugorut: Discussed on zoom, this is a bug.

cmd/infracost/testdata/breakdown_terraform_sync_usage_file/breakdown_terraform_sync_usage_file.golden
```diff
@@ -8,7 +8,7 @@ Project: infracost/infracost/cmd/infracost/testdata/breakdown_terraform_sync_usa
  └─ Insights queries data scanned                    Monthly cost depends on usage: $0.005 per GB   
                                                                                                     
  aws_cloudwatch_log_group.production_logs["media"]                                                  
- ├─ Data ingested                                    Monthly cost depends on usage: $0.50 per GB    
+ ├─ Data ingested                                             1,000  GB                     $500.00
```
hugorut, line 11: This seems like this was a bug before that was fixed by this PR, the resource defaults are as follows:

```
resource_type_default_usage:
  aws_cloudwatch_log_group:
    monthly_data_ingested_gb: 1000.0
```

aliscott (approved):
Nice tests!

PR #2290, opened 2023-02-13
feat: show all errored project

hugorut:
This change removes the early exit in the CLI if all projects have errored. This means that the CLI output will always return a successful list of projects even if all of them have errored. This means that the output JSON will be generated correctly if all projects have errored. This change also removes the output table from the comment if there are no projects in the output JSON.


aliscott (approved):
LGTM

PR #2304, opened 2023-02-21
Feat:(AWS) Adds support for Ephemeral Storage

### LINKED ISSUE #2138: AWS Lambda - missing cost configurations

robh007:
Should close #2138


robh007:
@aliscott The build seems to be failing because it can't access AWS.

aliscott:
@robh007 sorry for the delay, I've been struck by the virus. Those errors look expected. It looks like the sync usage stuff which is okay for the tests. The test failures seem to be because the usage data was updated. Did you have AWS credentials exported in your shell when running the update the tests?

robh007:
> @robh007 sorry for the delay, I've been struck by the virus. Those errors look expected. It looks like the sync usage stuff which is okay for the tests. The test failures seem to be because the usage data was updated. Did you have AWS credentials exported in your shell when running the update the tests?

Not that I'm aware of.

aliscott:
@robh007 I hope you don't mind I just pushed a commit to this branch to update the tests to see if that helps. I ran the following:
```
unset AWS_PROFILE
unset AWS_ACCESS_KEY_ID
unset AWS_SECRET_ACCESS_KEY
make test_update_cmd
make test_update_usage
```

aliscott:
@robh007 the failing tests are now fixed in master, so this so pass the tests when rebased.

internal/resources/aws/lambda_function.go
```diff
@@ -164,3 +184,22 @@ func (a *LambdaFunction) durationCostComponent(displayName string, usageTier str
 		},
 	}
 }
+
+func (a *LambdaFunction) storageCostComponent(quantity *decimal.Decimal, storageType string) *schema.CostComponent {
+	return &schema.CostComponent{
+		Name:            "Ephemeral Storage",
```
aliscott, line 190:
```suggestion
		Name:            "Ephemeral storage",
```

internal/resources/aws/lambda_function.go
```diff
@@ -86,6 +91,16 @@ func (a *LambdaFunction) BuildResource() *schema.Resource {
 	},
 	)
 
+	if a.MonthlyRequests != nil {
+		monthlyRequests = decimalPtr(decimal.NewFromInt(*a.MonthlyRequests))
+		gbSeconds = decimalPtr(calculateGBSeconds(memorySize, averageRequestDuration, *monthlyRequests))
+		storageGBSeconds = decimalPtr(calculateStorageGBSeconds(storageSize, *gbSeconds))
+
+		costComponents = append(costComponents, a.storageCostComponent(storageGBSeconds, storageType))
```
aliscott, line 99: I think we can move this logic down to be in the same block that we add the duration cost component, since we're already calculating `monthlyRequests` and `gbSeconds` there.

internal/resources/aws/lambda_function.go
```diff
@@ -86,6 +91,16 @@ func (a *LambdaFunction) BuildResource() *schema.Resource {
...
+	} else {
+		costComponents = append(costComponents, a.storageCostComponent(decimalPtr(decimal.NewFromInt(0)), storageType))
```
aliscott, line 101:
```suggestion
		costComponents = append(costComponents, a.storageCostComponent(nil, storageType))
```
I think this means it will show the pricing instead of a zero cost, which is what we want if there's no usage data added.

aliscott (commented):
Thanks @robh007 this is awesome! I added a couple of comments. It also looks like the rebase went a bit weird.

robh007:
Finally!!! Hopefully!!! @aliscott

aliscott (approved):

aliscott:
Thanks @robh007!

PR #2307, opened 2023-02-22
chore: add more ignored variables to evaluator

hugorut:
* adds more ignored variables to list
* changes pattern to match both `_` and `-` separated vars
* ignores case of variable

### DIFF
--- a/internal/hcl/evaluator.go
+++ b/internal/hcl/evaluator.go
@@ -39,7 +39,7 @@ var (
 		"data":     {},
 	}
 
-	sensitiveRegxp = regexp.MustCompile(`_token$|^token_|_secret$|^secret_|_password$|^password_|_username$|^username_|api_key|expiration_date`)
+	sensitiveRegxp = regexp.MustCompile(`^image[_|-]|[_|-]image$|saml[_|-]role$|secrets$|aws[_|-]profile$|key[_|-]id$|secret[_|-]key$|access[_|-]key$|[_|-]token$|^token[_|-]|[_|-]secret$|^secret[_|-]|[_|-]password$|^password[_|-]|[_|-]username$|^username[_|-]|api[_|-]key|expiration[_|-]date`)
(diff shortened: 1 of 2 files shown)


internal/hcl/evaluator.go
```diff
@@ -39,7 +39,7 @@ var (
 		"data":     {},
 	}
 
-	sensitiveRegxp = regexp.MustCompile(`_token$|^token_|_secret$|^secret_|_password$|^password_|_username$|^username_|api_key|expiration_date`)
+	sensitiveRegxp = regexp.MustCompile(`^image[_|-]|[_|-]image$|saml[_|-]role$|secrets$|aws[_|-]profile$|key[_|-]id$|secret[_|-]key$|access[_|-]key$|[_|-]token$|^token[_|-]|[_|-]secret$|^secret[_|-]|[_|-]password$|^password[_|-]|[_|-]username$|^username[_|-]|api[_|-]key|expiration[_|-]date`)
```
vdmgolub, line 42: Can we do something like this for better readability?
```suggestion
	sensitiveRegxp = regexp.MustCompile(strings.Join([]string{
		"^image[_|-]",
		"[_|-]image$",
		"saml[_|-]role$",
		"secrets$",
		"aws[_|-]profile$",
		"key[_|-]id$",
		"secret[_|-]key$",
		"access[_|-]key$",
		"[_|-]token$",
		"^token[_|-]",
		"[_|-]secret$",
		"^secret[_|-]",
		"[_|-]password$",
		"^password[_|-]",
		"[_|-]username$",
		"^username[_|-]",
		"api[_|-]key",
		"expiration[_|-]date",
	}, "|"))
```

vdmgolub (approved):

internal/hcl/evaluator.go
hugorut: 👍

PR #2310, opened 2023-02-25
Add MSSQL Elastic Pool

aliscott:
This adds support for the `azurerm_mssql_elasticpool` resource and the deprecated `azurerm_sql_elasticpool` resource.
It also fixes some issues with the existing `azurerm_mssql_database` resource:
1. Fix zone redundancy costs. These are added on top of the compute costs, not instead of.
2. Don't show extra data prices if there's no usage.
3. Support the different storage types for long term retention
4. Add PITR cost component
5. Show DTU cost components as hours so the units and costs match the Azure dashboard and pricing calculator.
6. Fix some issues to support non-case-sensitive attribute values

I think we can remove the `extra_data_storage_gb` usage param from `azurerm_mssql_database` since its value is always defined in Terraform. I haven't done that as part of this PR though.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -144,6 +144,7 @@ require (
 	github.com/creack/pty v1.1.11 // indirect
 	github.com/dimchansky/utfbom v1.1.1 // indirect
 	github.com/emirpasic/gods v1.12.0 // indirect
+	github.com/fatih/camelcase v1.0.0 // indirect
 	github.com/frankban/quicktest v1.11.3 // indirect
 	github.com/go-errors/errors v1.0.2-0.20180813162953-d98b870cc4e0 // indirect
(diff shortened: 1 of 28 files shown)


internal/resources/azure/sql_elasticpool.go
```diff
@@ -0,0 +1,42 @@
+package azure
```
hugorut, line 1: Do we need this file - it doesn't look like we use it?

hugorut (approved):
Good work 🚀

internal/resources/azure/sql_elasticpool.go
aliscott: 👍 cheers, have deleted

PR #2316, opened 2023-02-27
feat: Infracost template file support

hugorut:
Adds support for Infracost template files using go text/templates.

### DIFF
--- a/cmd/infracost/generate.go
+++ b/cmd/infracost/generate.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/infracost/infracost/internal/config/template"
+	"github.com/infracost/infracost/internal/ui"
+)
+
+type generateConfigCommand struct {
+	projectPath  string
+	templatePath string
+	outFile      string
+}
+
+func newGenerateConfigCommand() *cobra.Command {
+	var gen generateConfigCommand
+
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "Generate Infracost YML file from a template file",
+		Long:  "Generate Infracost YML file from a template file",
+		Example: `
+      infracost generate config --project-path . --template-path infracost.yml.tmpl
+      `,
+		ValidArgs: []string{"--", "-"},
+		RunE:      gen.run,
+	}
+
+	cmd.Flags().StringVar(&gen.projectPath, "project-path", ".", "Path to the Terraform project you want to run the template file on")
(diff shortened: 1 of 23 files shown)


cmd/infracost/generate.go
```diff
@@ -0,0 +1,95 @@
...
+		ValidArgs: []string{"--", "-"},
+		RunE:      gen.run,
+	}
+
+	cmd.Flags().StringVar(&gen.projectPath, "project-path", ".", "Path to the Terraform project you want to run the template file on")
+	cmd.Flags().StringVar(&gen.templatePath, "template-path", "", "Path to the Infracost template file which will generate the yml output")
+	cmd.Flags().StringVar(&gen.outFile, "out-file", "", "Path to the Infracost template file which will generate the yml output")
+
+	return cmd
+}
```
aliscott, line 38: @alikhajeh1 would be good if you review this bit as well.

cmd/infracost/generate.go
```diff
@@ -0,0 +1,95 @@
...
+		Short: "Generate Infracost YML file from a template file",
+		Long:  "Generate Infracost YML file from a template file",
+		Example: `
+      infracost generate config --project-path . --template-path infracost.yml.tmpl
+      `,
+		ValidArgs: []string{"--", "-"},
+		RunE:      gen.run,
+	}
+
+	cmd.Flags().StringVar(&gen.projectPath, "project-path", ".", "Path to the Terraform project you want to run the template file on")
```
aliscott, line 33: Not sure if `project` is the right term here. We just use `--path` elsewhere to refer to the repository path, but understand we might want to make this more explicit. Any ideas @alikhajeh1?

aliscott (approved):

cmd/infracost/generate.go
alikhajeh1: @aliscott and @hugorut yeah `project` is confusing here as we use that term to mean another thing in Infracost.
Maybe `--repo-path` with `Path to the Terraform repo or directory you want to run the template file on`

cmd/infracost/generate.go
```diff
@@ -0,0 +1,95 @@
...
+		Long:  "Generate Infracost config file from a template file",
+		Example: `
+      infracost generate config --project-path . --template-path infracost.yml.tmpl
+      `,
+		ValidArgs: []string{"--", "-"},
+		RunE:      gen.run,
+	}
+
+	cmd.Flags().StringVar(&gen.projectPath, "project-path", ".", "Path to the Terraform project you want to run the template file on")
+	cmd.Flags().StringVar(&gen.templatePath, "template-path", "", "Path to the Infracost template file which will generate the yml output")
```
alikhajeh1, line 34: @hugorut maybe `Path to the Infracost template file that will generate the config-file yaml output` ?

PR #2318, opened 2023-02-28
Add moved block into terraform schema 

lyonsden:
Adding `moved` as a valid block type into the `terraformSchemaV012` schema to prevent a silent failure.

Previous behavior would result in "no cloud resources were detected" if there was a `moved` declaration in the terraform file. By adding it to the schema normal processing is allowed to continue.

This block has no impact on the cost calculations so no additional processing is necessary.


examples/terraform/main.tf
```diff
@@ -31,6 +31,11 @@ resource "aws_lambda_function" "hello_world" {
   memory_size   = 1024 # <<<<< Try changing this to 512 to compare costs
 }
 
+moved {
+  from = aws_lambda_function.hola_mundo
+  to   = aws_lambda_function.hello_world
+}
+
```
hugorut, line 38:
```suggestion
```

can we remove this from the `examples/terraform` file as this is used as a getting started point for many users.

Perhaps we can add this moved block to `cmd/infracost/testdata/breakdown_multi_project_autodetect/dev/main.tf` instead as this will be covered by one of the breadkdown tests.

But in this case it would need to be >

```
moved {
  from = aws_instance.foo
  to   = aws_lambda_function.web_app
}
```

internal/hcl/block.go
```diff
@@ -49,6 +49,9 @@ var (
 				Type:       "data",
 				LabelNames: []string{"type", "name"},
 			},
+			{
```
hugorut, line 52: I think maybe a better solution would be to change `loadBlocksFromFile` to use `PartialContent` over `Content`, as we don't actually support type `moved` https://github.com/infracost/infracost/blob/2d4a16fe989b9c363f16f4295dd4466e85c4fde4/internal/hcl/block.go#L987

e.g:

```go
	contents, _, diags := file.hclFile.Body.PartialContent(schema)
	if diags != nil && diags.HasErrors() {
		return nil, diags
	}
```

PartialContent is like Content except that it permits the configuration to contain additional blocks or attributes not specified in the schema. This means we don't break in the future on new block type additions.

hugorut (approved):
LGTM

tim775:
@lyonsden @clete2 this is released in v0.10.18

PR #2334, opened 2023-03-07
fix: ignore parent terragrunt.hcl files in a Terragrunt directory

hugorut:
Resolves issue when a parent `terragrunt.hcl` file was using any `find` or `get` functions. These functions are called relative to the original Terragrunt config (the child `terragrunt.hcl` file). Thus any file include would fail with a not exists error. We now ignore parent `terragrunt.hcl` from initial evaluation and only read it when it is included by a child `terragrunt.hcl` (expected behaviour).

### DIFF
--- a/internal/providers/terraform/terragrunt_hcl_provider.go
+++ b/internal/providers/terraform/terragrunt_hcl_provider.go
@@ -334,8 +334,16 @@ func (p *TerragruntHCLProvider) prepWorkingDirs() ([]*terragruntWorkingDirInfo, 
 	if err != nil {
 		return nil, err
 	}
+
+	var filtered []string
+	for _, file := range terragruntConfigFiles {
+		if !p.isParentTerragruntConfig(file, terragruntConfigFiles) {
+			filtered = append(filtered, file)
+		}
+	}
+
 	// Filter these config files against the exclude paths so Terragrunt doesn't even try to evaluate them
-	terragruntConfigFiles = p.filterExcludedPaths(terragruntConfigFiles)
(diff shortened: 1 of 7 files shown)


hugorut:
I have an open pr here https://github.com/infracost/infracost/pull/2335 for the intermitently failing golangci-lint errors, so please ignore

internal/providers/terraform/terragrunt_hcl_provider.go
```diff
@@ -361,6 +369,53 @@ func (p *TerragruntHCLProvider) prepWorkingDirs() ([]*terragruntWorkingDirInfo,
...
+// with a find_in_parent_folders call. The find_in_parent_folders function searches up the directory tree
+// from the file and returns the absolute path to the first terragrunt.hcl. This means if it is found
+// we can treat this file as a child terragrunt.hcl.
+func (p *TerragruntHCLProvider) isParentTerragruntConfig(parent string, configFiles []string) bool {
+	for _, file := range configFiles {
+		if !isChildDirectory(parent, file) {
+			continue
+		}
+
+		file, err := os.Open(file)
```
vdmgolub, line 382: Do we need to close these files? Or should it not be that many to care as the execution is usually quick?

vdmgolub (approved):

internal/providers/terraform/terragrunt_hcl_provider.go
hugorut: Yes we should, for some reason the `file.Close` wasn't being recognized with my specific build tags so it got removed with `go fmt` I'm adding back in. Good spot.

PR #2335, opened 2023-03-07
fix: bump golangci-lint version to prevent fmt panic

hugorut:
* bumps version of golangci-lint to 1.51.2 to fix panic
* resolves lint errors that have been introduced with the upgrade
* changes terraform arn used for lambda to fix failing tests due to validation regexp introduction with `terraform-aws-provider` see https://github.com/infracost/infracost/pull/2337


cmd/infracost/comment_github.go
```diff
@@ -54,7 +54,7 @@ func commentGitHubCmd(ctx *config.RunContext) *cobra.Command {
 			}
 
 			tlsConfig.RootCAs = rootCAs
-			tlsConfig.InsecureSkipVerify = tlsInsecureSkipVerify
+			tlsConfig.InsecureSkipVerify = tlsInsecureSkipVerify // nolint: gosec
```
tim775, line 57: nice, i didn't think there were pragmas but i guess that's just golint

tim775 (approved):
🥇

PR #2339, opened 2023-03-07
chore: remove warning about logging info

hugorut:


internal/providers/terraform/hcl_provider.go
```diff
@@ -318,13 +318,7 @@ func (p *HCLProvider) Modules() []HCLProject {
...
 	runInParallel := parallelism > 1 && numJobs > 1
 	if runInParallel && !runCtx.Config.IsLogging() {
-		if runInParallel && !p.config.SuppressLogging {
-			fmt.Fprintln(os.Stderr, "Running multiple projects in parallel, so log-level=info is enabled by default.")
-			fmt.Fprintln(os.Stderr, "Run with INFRACOST_PARALLELISM=1 to disable parallelism to help debugging.")
-			fmt.Fprintln(os.Stderr)
-		}
-
-		p.logger.Logger.SetLevel(log.InfoLevel)
+		// set the config level to info so that the spinners don't report to the console.
```
hugorut, line 321: hacky for now - we should really do something with the spinners set on the child parsers, however, this is a bit more involved and priortising speed.

tim775 (approved):

Issue #2329, opened 2023-03-03
Add support for Azure Web Apps


Scott-Emberson:
Support for Azure Web Apps
Terraform Resource: `azurerm_windows_web_app` 
Terraform Documentation: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/windows_web_app
Azure Pricing: https://azure.microsoft.com/en-gb/pricing/details/app-service/windows/

pacorreia:
And azurerm_linux_web_app as well!

hassankhosseini:
@alikhajeh1 @aliscott - let's add these to be groomed please 🚀

hugorut:
hey @Scott-Emberson and @pacorreia , a fix for this is in master and will go out with the next release. I'll ping you here when it's available. If either of you are using the GH app, these resources should already be available.

tim775:
This is live in v0.10.19

PR #2343, opened 2023-03-09
feat: add remaining azure app service terraform components

hugorut:
closes https://github.com/infracost/infracost/issues/2329

* adds `service_plan` resource
* changes `app_service_plan` resource to pick up new plan types
* marks web apps as free resource as these have no costs associated with then, due to the service plan


tim775 (approved):
Might want to add service plan as a schema.CoreResource.

PR #2349, opened 2023-03-13
feat: support for azurerm_linux_function_app & azurerm_windows_function_app

hugorut:
Closes https://github.com/infracost/infracost/issues/2328

* Adds support for new funciton definitions by extracting old function logic into new shared FunctionApp struct
* Fixes bug with `for_each` attributes not being able to reference parent attributes that used `each.value.*`


internal/providers/terraform/azure/function_linux_app_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/infracost/infracost/internal/providers/terraform/tftest"
+)
+
+func TestAzureRMLinuxAppFunctionGoldenFile(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode")
+	}
+
+	opts := tftest.DefaultGoldenFileOptions()
+	opts.CaptureLogs = true
```
tim775, line 15: Looks like you don't need to capture logs anymore and could add the t.Parallel back in if you want

internal/resources/azure/function_app.go
```diff
@@ -0,0 +1,253 @@
...
+	OSType  string
+
+	MonthlyExecutions   *int64 `infracost_usage:"monthly_executions"`
+	ExecutionDurationMs *int64 `infracost_usage:"execution_duration_ms"`
+	MemoryMb            *int64 `infracost_usage:"memory_mb"`
+	Instances           *int64 `infracost_usage:"instances"`
+}
+
+func (r *FunctionApp) CoreType() string {
+	return "FunctionAPp"
```
tim775, line 47:
```suggestion
	return "FunctionApp"
```

tim775 (approved):

internal/resources/azure/function_app.go
hugorut: whoops

PR #2361, opened 2023-03-16
feat: calculate all Azure Monitor Action Group costs based on the number of time it is triggered

tim775:


internal/providers/terraform/azure/testdata/monitor_action_group_test/monitor_action_group_test.golden
```diff
@@ -1,37 +1,49 @@
...
+ ├─ Email notifications (2)                                     Monthly cost depends on usage: $0.00002 per emails         
+ ├─ ITSM connectors (1)                                         Monthly cost depends on usage: $0.005 per events           
+ ├─ Push notifications (1)                                      Monthly cost depends on usage: $0.00002 per notifications  
+ ├─ Secure web hooks (1)                                        Monthly cost depends on usage: $0.00006 per notifications  
+ ├─ Web hooks (2)                                               Monthly cost depends on usage: $0.000006 per notifications 
+ ├─ SMS messages                                                                                                           
+ │  └─ Country code 1 (1)                                       Monthly cost depends on usage: $0.00645 per messages       
+ └─ Voice calls                                                                                                            
+    └─ Country code 86 (1)                                      Monthly cost depends on usage: $0.06 per calls
```
vdmgolub, line 24: Doesn't read right when the units are in the plural. Maybe worth introducing the notion of single/multiple in units in the future.

vdmgolub (approved):

PR #2369, opened 2023-03-21
fix: Fix for_each blocks not picking up references

tim775:
@hugorut this fixes the issue I was seeing and I think it makes sense - at least the `configuration` block in a terraform generate plan.json doesn't contain the for_each names either.


hugorut (approved):
LGTM

PR #2372, opened 2023-03-23
enhance: scope the config-template functions to the repo directory

aliscott:


internal/config/template/testdata/external_dirs/infracost.yml.tmpl
```diff
@@ -0,0 +1,17 @@
...
+dot_path: success
+{{- else }}
+dot_path: fail
+{{- end }}
+
+{{- if pathExists "../external_dirs" "main.tf" }}
+dot_dot_path: success
+{{- else }}
+dot_dot_path: fail
+{{- end }}
```
hugorut, line 17: This test doesn't actually seem to be testing what we're adding which is scoping to the repo directory. Can we add a few test cases to make sure that's covered, i.e. paths outside the repo directory.

hugorut (commented):
Can we add some test cases to ensure the outside of repo dir functionality is covered. Thanks

internal/config/template/testdata/external_dirs/infracost.yml.tmpl
aliscott: Isn't that what the first case on line 1 is testing?

hugorut: but shouldn't that fail? The golden file output shows it passed?

hugorut: ah sorry I'm reading the if else wrong - that was confusing me. Can we change the statements around so the output is "fail"?

internal/config/template/parser.go
```diff
@@ -123,10 +123,15 @@ func (p *Parser) contains(s, substr string) bool {
...
+		base = filepath.Join(p.repoDir, base)
 	}
 
-	if !filepath.IsAbs(base) {
-		base = filepath.Join(p.repoDir, base)
+	// Ensure the base path is within the repo directory
+	baseAbs, _ := filepath.Abs(base)
+	repoDirAbs, _ := filepath.Abs(p.repoDir)
+	if !strings.HasPrefix(fmt.Sprintf("%s%s", baseAbs, string(filepath.Separator)), fmt.Sprintf("%s%s", repoDirAbs, string(filepath.Separator))) {
+		return false
```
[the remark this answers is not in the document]
hugorut: Ok yup - the abs conversion makes sense then. Let's stick with your prefix way then - but what's the reasoning for sticking "filepath.Separator" at the end of the two paths? This doesn't seem necessary but maybe I'm missing something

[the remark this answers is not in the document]
aliscott: > but what's the reasoning for sticking "filepath.Separator" at the end of the two paths?

To solve the case where `infracost/infra` is a prefix of `infracost/infracost`.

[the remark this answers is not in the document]
aliscott: `.Abs` calls `Clean` so we don't need to strip it manually. I'll add a test case for it.

hugorut (approved):

internal/config/template/testdata/external_dirs/infracost.yml.tmpl
aliscott: 👍 done

PR #2394, opened 2023-04-03
feat: default variables for infracost config template files

hugorut:
Adds support for default data to be passed into the template parser. This allows up to provide top level metadata variables that users can interact with. Currently the only supported variable is `.branch` which represents the current branch name of the repo, e.g. `main`, `test`.

### DIFF
--- a/internal/config/template/parser.go
+++ b/internal/config/template/parser.go
@@ -27,7 +27,7 @@ type Parser struct {
 
 // NewParser returns a safely initialized Infracost template parser, this builds the underlying template with the
 // Parser functions and sets the underlying default template name. Default data can be passed to the parser which
-// we be passed to the template on execution.
+// will be passed to the template on execution.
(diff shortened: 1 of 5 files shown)


internal/config/template/parser.go
```diff
@@ -22,13 +22,15 @@ var (
...
 	repoDir  string
 	template *template.Template
+	data     map[string]interface{}
 }
 
 // NewParser returns a safely initialized Infracost template parser, this builds the underlying template with the
-// Parser functions and sets the underlying default template name.
-func NewParser(repoDir string) *Parser {
+// Parser functions and sets the underlying default template name. Default data can be passed to the parser which
+// we be passed to the template on execution.
```
vdmgolub, line 30:
```suggestion
// will be passed to the template on execution.
```

vdmgolub (approved):

PR #2399, opened 2023-04-05
feat: Add support for Azure SignalR Service

tim775:


internal/providers/terraform/azure/testdata/signalr_service_test/signalr_service_test.tf
```diff
@@ -0,0 +1,35 @@
...
+
+  name                = "tfex-signalr"
+  location            = azurerm_resource_group.example.location
+  resource_group_name = azurerm_resource_group.example.name
+
+  sku {
+    name     = each.value
+    capacity = 5
+  }
+}
```
vdmgolub, line 35: Let's fix this missing newline here. Otherwise, when the next person opens this file in the editor and saves it, it will be added, resolving in a git diff line.

internal/providers/terraform/azure/testdata/signalr_service_test/signalr_service_test.usage.yml
```diff
@@ -0,0 +1,8 @@
+version: 0.1
+resource_usage:
+
+  azurerm_signalr_service.example_with_usage["Standard_S1"]:
+    monthly_additional_messages: 1234567
+
+  azurerm_signalr_service.example_with_usage["Premium_P1"]:
+    monthly_additional_messages: 12345678
```
vdmgolub, line 8: Same here.

internal/resources/azure/signalr_service.go
```diff
@@ -0,0 +1,116 @@
...
+func (r *SignalRService) PopulateUsage(u *schema.UsageData) {
+	resources.PopulateArgsWithUsage(r, u)
+}
+
+// BuildResource builds a schema.Resource from a valid SignalRService struct.
+// This method is called after the resource is initialised by an IaC provider.
+// See providers folder for more information.
+func (r *SignalRService) BuildResource() *schema.Resource {
+	// normalize sku to first letter capitalized
+	sku := cases.Title(language.English).String(strings.ToLower(r.SkuName))
```
vdmgolub, line 49: It might be a great candidate to become a new helper function. I see it's used in several places already. :)

vdmgolub (approved):

internal/providers/terraform/azure/testdata/signalr_service_test/signalr_service_test.tf
tim775: I hear you (though for whatever reason goland doesn't do that) but I feel like these kind of style issues should be addressed through code (i.e. `make fmt`) not code reviews.  There are over 100 .tf files without trailing newline and for whatever reason `terraform fmt` [doesn't want to enforce this](https://github.com/hashicorp/terraform/issues/23223) but I think it should be easy enough to right a `find + awk` that does it so I'll do that in a separate pr.

PR #2409, opened 2023-04-17
fix: support name as a global aws resource ref

hugorut:
This PR adds support to use `name` as a global resource reference. This resolves issues with resources that reference dependent resources with `name`. For example aws_autoscaling_group.launch_configuration:

```
resource "aws_autoscaling_group" "asg_lc_min_size_name_ref" {
  launch_configuration = aws_launch_configuration.lc_basic.name
  max_size             = 3
  min_size             = 2
}
```

Prior to this change `asg_lc_min_size_name_ref` would have blank references. I've decided to put move the `name` into the default resource ID func as I believe referencing a resource by `name` is probably quite a common pattern across a wide range of resources.

### DIFF
--- a/internal/hcl/block.go
+++ b/internal/hcl/block.go
@@ -428,6 +428,10 @@ func SetUUIDAttributes(moduleBlock *Block, block *hcl.Block) {
 				body.Attributes["id"] = newUniqueAttribute("id", withCount, withEach)
 			}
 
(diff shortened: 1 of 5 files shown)


aliscott (approved):
Can we now clean up `CustomRefIDFunc`s that lookup by "name"?

Issue #1981, opened 2022-08-19
Add Support for AWS OpenSearch service


yuha0:
Main TF resource: [aws_opensearch_domain](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/opensearch_domain)
Pricing: https://aws.amazon.com/opensearch-service/pricing/

### Note for community
Please 👍 if you'd like to upvote this or receive updates.
Add a comment if you'd like to [contribute](https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md).

vdmgolub:
Hi! Thank you for creating this issue! Please let us know if you’d like to [contribute](https://github.com/infracost/infracost/blob/master/CONTRIBUTING.md#adding-new-resources), otherwise we’ll open it up for the whole community (we regularly prioritize and work on them too). We usually add details about each resource listing the cost components Infracost should show. If you have ideas what this could be, you could add them to the issue. We have a guide that explains the process: [Resource mapping guide](https://github.com/infracost/infracost/blob/master/contributing/resource_mapping_guide.md).

aliscott:
@yuha0 this has now been released in [v0.10.21](https://github.com/infracost/infracost/releases/tag/v0.10.21). Thank you @ABaboshin for adding this resource 🙏 

cc'ing everyone who upvoted: @gmeligio, @samiles, @oc-stephen-bennett, @bmbferreira, @nebur1989, @Shrivastava-Varsha, @PatMyron, @ciphero, and @bobsut.

PR #2415, opened 2023-04-18
feat(aws): add opensearch support

ABaboshin:
Addressing issue https://github.com/infracost/infracost/issues/1981


internal/resources/aws/opensearch_domain.go
```diff
@@ -0,0 +1,219 @@
+package aws
```
hugorut, line 1: Unless I'm mistaken this file is identical to the `elasticsearch_domain` resource. I'd prefer that we didn't just duplicate this and instead just use the `ElasticsearchDomain` resource. Perhaps we can rename `ElasticsearchDomain` it to `SearchDomain` and add doc blocks to convey that it is used for both `OpensearchDomain` and `ElasticsearchDomain`.

hugorut (changes requested):
Looks good, but let's just remove the duped `OpensearchDomain`. Seems that the tests are also failing as they've been copied over from the elastic search ones. Could we update the terraform and golden file output. Thanks

hugorut (approved):
Thanks @ABaboshin, I've made some minor tweaks to this as we wanted to get this into the release today. Thank you for your help in adding OpenSearch support 🚀

PR #2425, opened 2023-04-25
Add Azure Network Watcher

aliscott:
Closes https://github.com/infracost/infracost/issues/1255

PR #2429, opened 2023-04-26
enhance(azure): add usage example for Azure Network Monitor

aliscott:
Missed this as part of https://github.com/infracost/infracost/pull/2425


hugorut (approved):
My bad - missed this as well

Issue #2359, opened 2023-03-16
GCP: GKE with spot instances in nodepool is detected as on-demand


syphernl:
While setting up a new project on GCP I noticed that Infracost detects spot instances in a GKE Nodepool as "on-demand" rather than spot nodes.

Example Terraform code:

```tf
module "gke" {
  source  = "terraform-google-modules/kubernetes-engine/google"
  version = "25.0.0"

  project_id                      = "test"
  name                            = "test-cluster"
  regional                        = false
  region                          = "europe-west4"
  zones                           = ["europe-west4-a"]
  network                         = "my-test-network"
  subnetwork                      = "my-subnetwork"
  ip_range_pods                   = "my-iprange"
  ip_range_services               = "my-services"
  release_channel                 = "REGULAR"
  remove_default_node_pool        = true

  node_pools = [
    {
      name               = "spot-nodes"
      spot               = true
      initial_node_count = 0
      max_count          = 3
      disk_size_gb       = 25
      enable_secure_boot = true
      disk_type          = "pd-ssd"
      machine_type       = "n1-standard-1"
    }
  ]
}
```

This produces:

```bash
Evaluating Terraform directory at .
  ✔ Downloading Terraform modules 
  ✔ Evaluating Terraform directory 
  ✔ Retrieving cloud prices to calculate costs 

Project: .

 Name                                                       Monthly Qty  Unit   Monthly Cost (EUR) 
                                                                                                   
 module.gke.google_container_cluster.primary                                                       
 ├─ Cluster management fee                                          730  hours              €68.98 
                                                                                                   
 module.gke.google_container_node_pool.pools["spot-nodes"]                                         
 ├─ Instance usage (Linux/UNIX, on-demand, n1-standard-1)           730  hours              €25.25 
 └─ SSD provisioned storage (pd-ssd)                                 25  GB                  €4.42 
                                                                                                   
 OVERALL TOTAL (EUR)                                                                        €98.65 
──────────────────────────────────
7 cloud resources were detected:
∙ 2 were estimated, all of which include usage-based costs, see https://infracost.io/usage-file
∙ 5 were free, rerun with --show-skipped to see details
```

According to the Google Pricing Calculator running a `n1-standard-1` as spot for 730 hours should be `EUR 6,18`  and not `EUR 25,25`

aliscott:
Thanks for reporting @syphernl. Do you have example Terraform code we can use to reproduce this?

syphernl:
[...]

PR #2447, opened 2023-05-05
fix: price spot like preemptible, not on-demand (#2359)

gdubicki:
This is a fix of #2359 for the spot instance pools in GKE, which are now priced like on-demand instances, while they should be priced like the preemptible instances.


aliscott:
Thanks @gdubicki. This looks great 🚀

aliscott (approved):

PR #2455, opened 2023-05-08
fix: `INFRACOST_LOG_LEVEL` overriding `--log-level` flag

aliscott:
When a config file was used, the `INFRACOST_LOG_LEVEL` env var was overriding the `--log-level` flag. The flag should have precendence over the env var.


hugorut (approved):
👍

PR #2464, opened 2023-05-15
fix: Do no cache local modules in the manifest

tim775:


hugorut (approved):
This looks good to me, and I think this should be a sane change. Let's just make sure we put it through it's paces by testing on the runner with some different scenarios. (I'm sure you have, just making it clear that this might be a change we need to be vigilant on)

PR #2476, opened 2023-05-30
fix: for_each keys can be any string (even an empty string)

aliscott:


aliscott:
Test failures are related to the AWS provider update. Will handle in a separate PR.

hugorut (approved):

PR #2496, opened 2023-06-14
feat: traverse parent directories to find .infracost cache folder

hugorut:
attempt to find the cache path in parent directories given a starting path. Resolves issue where multiple `.infracost` directories are created in vscode-infracost extension.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -139,17 +139,17 @@ func TestConfig_CachePath(t *testing.T) {
 		expected string
 	}{
 		{
-			name:     "with no .infracost directory",
+			name:     "should return the root directory with no .infracost directory in either the root or parent directories",
 			path:     "testdata/no_cache_init",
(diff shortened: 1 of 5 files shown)


internal/config/config_test.go
```diff
@@ -130,3 +131,42 @@ projects:
...
+			path:     "testdata/no_cache_init",
+			expected: "testdata/no_cache_init",
+		},
+		{
+			name:     "with parent cache",
+			path:     "testdata/parent_cache_path/parent/child",
+			expected: "testdata/parent_cache_path/parent",
+		},
+		{
+			name:     "with root cache path",
```
tim775, line 152: I don't understand this test case

tim775 (approved):

internal/config/config_test.go
hugorut: it's to define if the `.infracost` directory is already in the folder it shouldn't traverse the parent directories - i'll rename

Issue #1856, opened 2022-07-13
data block `aws_availability_zones` providing `names` attribute as a string


aliscott:
**Terraform code:**
```hcl
provider "aws" {
  region                      = "eu-west-1"
  skip_credentials_validation = true
  skip_requesting_account_id  = true
  access_key                  = "mock_access_key"
  secret_key                  = "mock_secret_key"
}

data "aws_availability_zones" "available" {}

resource "aws_vpc" "tenant_aws_vpc" {
  cidr_block       = "10.0.0.0/16"
  instance_tenancy = "default"
}

resource "aws_subnet" "public_subnet" {
  count             = length(data.aws_availability_zones.available.names)
  vpc_id            = aws_vpc.tenant_aws_vpc.id
  cidr_block        = "172.5.${10 + count.index}.0/24"
  availability_zone = data.aws_availability_zones.available.names[count.index]
}

resource "aws_vpc_endpoint" "with_dynamic_subnet" {
  vpc_id = aws_vpc.tenant_aws_vpc.id
  subnet_ids = [
    for public_subnet in aws_subnet.public_subnet : public_subnet.id
  ]
  service_name      = "com.amazonaws.us-east-1.ec2messages"
  vpc_endpoint_type = "Interface"

  security_group_ids = ["fakeid"]

  private_dns_enabled = true
}
```

**Actual result:**

This shows me the price assuming there's 22 subnets, when there's actually only 3. This looks like it's because the HCL parsing is mocking `data.aws_availability_zones.available.names` to `"availability_zone-mock"` which is a string with length 22.

```
 Name                                    Monthly Qty  Unit            Monthly Cost

 aws_vpc_endpoint.with_dynamic_subnet
 ├─ Data processed (first 1PB)         Monthly cost depends on usage: $0.01 per GB
 └─ Endpoint (Interface)                      16,060  hours                $176.66

 OVERALL TOTAL                                                             $176.66
 ```
 
 **Expected result:**
 
I think the expected result would be to match the availability zone count from AWS, but with HCL parsing this isn't possible since we don't have creds to query the cloud. So maybe a good option initially is just to default the count of arrays to 1 with HCL parsing, and in the future we can add them to the usage file or have some way of overriding them.

**Comment by ckdake** [comment]
A default of one would be nice, but as an infracost user I would be happy to either add an annotation to the file. e.g.:

```hcl
#infracost:expected-count(4)
data "aws_availability_zones" "available" {}
```

[...]

PR #2504, opened 2023-06-21
feat: add support to create well-known attributes for blocks

hugorut:
Goes some way to solving https://github.com/infracost/infracost/issues/1856

Adds support to return well-known values for blocks where mock data is insufficient. This is mostly applicable to data blocks. In particular data blocks such as `aws_availability_zones` where the return values can be inferred and don't require account/api access. This PR allows us to override the `Values` function for the `Block` with a more specific value func.

Initial supported blocks are:

* data.aws_availability_zones
* data.google_compute_zones
* resource.random_shuffle

This pr also resolves an issue where the provider data was being set incorrectly in the evaluation context, using the key `provider` instead of `aws`, `google`, e.t.c.


hugorut:
Note: 

* GCP/Azure test failures are on master - not related to these changes
* `zones_aws` is incomplete at the moment until we enable more regions on our account - I will update when these come online

internal/hcl/block.go
```diff
@@ -1082,3 +1090,116 @@ func loadBlocksFromFile(file file, schema *hcl.BodySchema) (hcl.Blocks, error) {
...
+	defaultGCPRegion = "us-central1"
+
+	blockValueFuncs = map[string]BlockValueFunc{
+		"data.aws_availability_zones": awsAvailabilityZonesValues,
+		"data.google_compute_zones":   googleComputeZonesValues,
+		"resource.random_shuffle":     randomShuffleValues,
+	}
+)
+
+func randomShuffleValues(b *Block) cty.Value {
```
aliscott, line 1108: It's not clear to me what we're doing here. From what I can tell we're essentially slicing the list - I assume that's so the result of this is always consistent?

tools/describezones/main.go
```diff
@@ -0,0 +1,178 @@
...
+			config.WithRegion(name),
+		)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		regionalSvc := ec2.NewFromConfig(regionalConf)
+
+		result, err := regionalSvc.DescribeAvailabilityZones(context.Background(), &ec2.DescribeAvailabilityZonesInput{
+			AllAvailabilityZones: aws.Bool(true),
```
aliscott, line 58: Looks like this returns the wavelength and other zones as well. I guess we only want these if the user has `all_available_zones` on the `data.aws_availability_zones` resource. I guess this is difficult because we have no way of knowing if the user has opted into them or not. So maybe if we find this is an issue with anyone in the future, we can always assume they've not opted into these zones as a default.

aliscott (approved):

internal/hcl/block.go
hugorut: it's a function to mock the values returned from `resource.random_shuffle` https://github.com/hashicorp/terraform-provider-random/blob/main/docs/resources/shuffle.md. This resource uses the `result_count` attribute to return a slice of the `input` attribute, which is randomly sorted. We don't randomly sort it here as it's unecessary for us, and a consistent output is probably better for debuging/tests. I'll add a comment to clarify.

PR #2506, opened 2023-06-22
feat: Add tag policies

tim775:


internal/providers/terraform/hcl_provider.go
```diff
@@ -600,6 +606,28 @@ func (p *HCLProvider) countReferences(block *hcl.Block) *countExpression {
...
+	if err != nil {
+		return ""
+	}
+
+	h := sha256.New()
+	h.Write(serialized)
+
+	return hex.EncodeToString(h.Sum(nil))
+}
+
```
vdmgolub, line 630: @hugorut Could you please review this change? I've updated it from `sha1` to `sha256` to resolve a gosec warning. `// #nosec` didn't work for me. What is the right way to deal with this?

hugorut: LGTM

PR #2512, opened 2023-06-26
fix: override Terragrunt funcs

hugorut:
Resolves issue where Terragrunt functions that are have no impact on Infracost cause program termination. We now provide `Functions` overrides to the `TerragruntOptions` which in turn sets these on the `hcl.EvalContext`.

This change also now changes the go modules to use the Infracost fork of the Terragrunt library. We'll use this for now, but actively work on getting the changes we've made to the library accepted into the offical Terragrunt repo. 

Related Infracost Terragrunt PR: https://github.com/infracost/terragrunt/pull/3


internal/hcl/funcs/yaml.go
```diff
@@ -0,0 +1,44 @@
...
+	},
+	Type: func(args []cty.Value) (cty.Type, error) {
+		if !args[0].IsKnown() {
+			return cty.DynamicPseudoType, nil
+		}
+		if args[0].IsNull() {
+			return cty.NilType, function.NewArgErrorf(0, "YAML source code cannot be null")
+		}
+		val := args[0].AsString()
+		if strings.HasPrefix(val, "mock") {
```
aliscott, line 26: Is it possible for this value to actually start with `mock` and not be a mock?

internal/hcl/funcs/yaml.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"strings"
+
+	yaml "github.com/zclconf/go-cty-yaml"
+	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/function"
+)
+
+var YAMLDecodeFunc = function.New(&function.Spec{
```
aliscott, line 11: Might be good to add some docs to these functions to state why we're overriding them.

aliscott (approved):

internal/hcl/funcs/yaml.go
hugorut: I guess it is theoretically possible but it's highly unlikely as it would require the first key of a yaml/json file to have `mock:` 

e.g:

```
mock:
   - value
```

So I think it's safe to assume it's our internal mocks - I guess longer term we could change our mock string to `infracostmock-...` which would probably mitigate this worry?

internal/hcl/funcs/yaml.go
hugorut: good shout

PR #2526, opened 2023-06-29
Feat/add provider metadata

tim775:


internal/schema/project.go
```diff
@@ -133,6 +142,10 @@ type Project struct {
 	HasDiff              bool
 }
 
+func (p *Project) AddProviderMetadata(metadatas []ProviderMetadata) {
+	p.Metadata.Providers = metadatas
+}
```
hugorut, line 147: Maybe we want to do a nil check here for the p.Metadata since it's a pointer?

hugorut (approved):

Issue #1980, opened 2022-08-18
Validate `--github-api-url`


alikhajeh1:
## Problem
```
infracost comment github --help

      --github-api-url string     GitHub API URL (default "https://api.github.com")
```

When implementing the GitHub comment feature, we struggled with knowing why it wasn’t working. Once we changed the comment mode from “update” to “new” it logged the actual URL it was trying to post to. We use GitHub Enterprise and provided an endpoint that included the “/api/v3/” suffix which Infracost appends on its own. The logging is inconsistent between these functions, and it should probably attempt to detect whether the suffix already exists before appending it.

## Proposed solution
Since we [create github API clients](https://github.com/infracost/infracost/blob/master/internal/comment/github.go#L80-L122), should we check the `--github-api-url` and log a user-friendly error if the user has already added `/api/v3`?

varshneydevansh:
Can you assign this to me?

alikhajeh1:
Thanks @varshneydevansh !

varshneydevansh:
Okay, so from yesterday I set up the project and learned about how it's working with Terraform. (Henceforth, I have also gained some knowledge about it.)

So, according to the above proposed solution, this is what I came up with in the [internal/comment/github.go](https://github.com/infracost/infracost/blob/master/internal/comment/github.go#L80-L122)

```go
// Check if "/api/v3/" already exists in the URL path.
if strings.Contains(u.Path, "/api/v3/") {
	return nil, nil, fmt.Errorf("The /api/v3/ suffix should not be included in the --github-api-url.")
}
```
But, I think if the user is already aware about the /api/v3.
Instead of giving an error, let the user keep going if they put /api/v3 in the apiURL. This would be a good experience because the user already knows about it.

Somewhat like this?
```go
// If "/api/v3/" already exists, do not append "/api/".
if strings.Contains(u.Path, "/api/v3/") {
    apiURL = u.String()
} else {
	// Add api to path if it doesn't exist
	if !strings.HasSuffix(u.Path, "/api/") {
		u.Path += "api/"
	}
	apiURL = u.String()
}
```

Guide me if I am missing anything here.


For the testing, I have downloaded the required file from the https://www.infracost.io/docs/#2-get-api-key
I will do it today.

alikhajeh1:
Thanks! Given that the user might add env vars for this variable in CI, it's probably best if we ensure they set the correct URL as we might use it in other places too.

[...]

Issue #2407, opened 2023-04-13
Infracost uses Premium disk for AKS even though it's not available for given VM family


kamil-mrzyglod:
Hi,

It looks like Infracost makes invalid assumptions regarding OS disks provisioned for AKS.

Infracost version: v0.10.19

### Steps to reproduce:

```
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "=3.0.0"
    }
  }
}

# Configure the Microsoft Azure Provider
provider "azurerm" {
  features {}
  skip_provider_registration = true
}

# Create a resource group
resource "azurerm_resource_group" "example" {
  name     = "example-resources"
  location = "West Europe"
}

resource "azurerm_kubernetes_cluster" "example" {
  name                = "example-aks1"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  dns_prefix          = "exampleaks1"

  default_node_pool {
    name       = "default"
    node_count = 1
    vm_size    = "Standard_D2_v2"
  }

  identity {
    type = "SystemAssigned"
  }

  tags = {
    Environment = "Production"
  }
}
```
Run `infracost breakdown --path .`

Result:
![image](https://user-images.githubusercontent.com/11725731/231722291-43924b6b-72d4-4363-8ae3-08d0d4ea9d3a.png)

### Expected result
`Standard_D2_v2` in Azure doesn't support Premium SSD disks. Estimation value should be ~108.88 USD

tusharrverma:
Can I pick this one up?

aliscott:
Thanks @tusharrverma 🙏, I've assigned you. Please reach out if we can help.

tusharrverma:
Hi team, I've not worked on Golang before, however I can see that for Kubernetes node pools, the DiskType is set to a static value of "Premium" if I'm not wrong. This should be dynamic and needs to come from properties of OsDisk which is being provisioned. Although I'm not sure what changes are to be done exactly here, but I hope this info would help to someone who will make code changes -

<img width="763" alt="image" src="https://user-images.githubusercontent.com/101072123/233860589-a8c86c27-cd53-4046-a8fe-210c6220520b.png">

Feel free to reassign the issue :)

varshneydevansh:
Is this issue still pending?

I did look upon few of the documentation of Azure -

https://learn.microsoft.com/en-us/azure/virtual-machines/vm-usage
[...]

PR #2534, opened 2023-07-03
added /api/v3 check in internal/comment/github.go

varshneydevansh:
Validate --github-api-url #1980

We use GitHub Enterprise and provided an endpoint that included the “/api/v3/” suffix which Infracost appends on its own. The logging is inconsistent between these functions, and it should probably attempt to detect whether the suffix already exists before appending it.

### DIFF
--- a/internal/apiclient/pricing.go
+++ b/internal/apiclient/pricing.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"math"
 	"net/http"
 	"os"
 
@@ -37,6 +38,11 @@ type PriceQueryKey struct {
 type PriceQueryResult struct {
 	PriceQueryKey
 	Result gjson.Result
+}
+
+type BatchRequest struct {
+	keys    []PriceQueryKey
+	queries []GraphQLQuery
 }
 
 func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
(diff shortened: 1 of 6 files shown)


varshneydevansh (commented):
github.go file is updated correctly with the check for the "/api/v3" in the URL. This file change belongs to the issue #2534 

However, in [kubernetes_cluster_node_pool.go](https://github.com/infracost/infracost/pull/2534/files#diff-4a93ab291f5dba270b6b6633bab1425b5936d16b844a0b61b77c77aaaca82d4e) file there could be some edge cases where the `regex` might not work, but the probability of this is low, and also **I haven't run the test cases yet** for this file. 
This change belongs to the issue Infracost uses Premium disk for AKS even though it's not available for given VM family #2407

internal/comment/github.go
```diff
@@ -113,6 +113,10 @@ func newGitHubAPIClients(ctx context.Context, token string, apiURL string, tlsCo
 	if !strings.HasSuffix(u.Path, "/") {
 		u.Path += "/"
 	}
+	// Check if "/api/v3/" already exists in the URL path.
+	if strings.Contains(u.Path, "/api/v3/") {
```
aliscott, line 117:
```suggestion
	if strings.HasSuffix(u.Path, "/api/v3/") {
```
We could use `HasSuffix` here to make sure we are only checking the end of the string.

varshneydevansh:
```go
	// Check if "/api/v3/" already exists in the URL path.
	if strings.HasSuffix(u.Path, "/api/v3/") {
		return nil, nil, fmt.Errorf("the /api/v3/ suffix should not be included in the --github-api-url")
	}
```
After this, we also have to lowercase the first **The** as it will show an error **error string should not be capitalized**

PR #2539, opened 2023-07-04
fix: add cloudUrl to addRun response

hugorut:
resolves issue where workflow CI users cannot see the "View in Infracost dashboard".


vdmgolub (approved):
The k8s are not supposed to be in the PR, right?

PR #2547, opened 2023-07-05
updated check for OS disks provisioned for AKS #2407

varshneydevansh:
### Issue >
 Infracost uses Premium disk for AKS even though it's not available for given VM family #2407

### Changes made> 

[kubernetes_cluster_node_pool.go](https://github.com/infracost/infracost/blob/master/internal/providers/terraform/azure/kubernetes_cluster_node_pool.go)

in the below function, added the instaceType to the calling function of `aksOSDiskSubResource()`
```go
func aksClusterNodePool()
{
    // rest of the code
    osDisk := aksOSDiskSubResource(region, diskSize, instanceType)

}
```

in the below function `aksOSDiskSubResource()` added the call to the new function `aksGetStorageType()` which returns a string whether the OS Disk provisioned is **Premium** or **Standard** based on the naming convention subfamily has any known premium prefixes (DS, GS, and M), and if so, it returns “Premium”. If not, it checks if the subfamily contains an 's' as an 'Additive Feature', as specified in the Azure VM naming conventions with the help of the **regex**. If it finds a match, it also returns “Premium”. Otherwise, it returns “Standard”.

```go
func aksOSDiskSubResource(region string, diskSize int, instanceType string) *schema.Resource {
	diskType := aksGetStorageType(instanceType)
        // rest of the code
}
```

### DIFF
--- a/internal/providers/terraform/azure/kubernetes_cluster_node_pool.go
+++ b/internal/providers/terraform/azure/kubernetes_cluster_node_pool.go
@@ -1,6 +1,7 @@
 package azure
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/shopspring/decimal"
@@ -84,7 +85,7 @@ func aksClusterNodePool(name, region string, n gjson.Result, nodeCount decimal.D
 		if n.Get("os_disk_size_gb").Type != gjson.Null {
 			diskSize = int(n.Get("os_disk_size_gb").Int())
 		}
-		osDisk := aksOSDiskSubResource(region, diskSize)
+		osDisk := aksOSDiskSubResource(region, diskSize, instanceType)
 
 		if osDisk != nil {
 			subResources = append(subResources, osDisk)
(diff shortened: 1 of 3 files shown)


aliscott:
@varshneydevansh this looks great. Is it possible to add a test case for a non-premium instance type here: https://github.com/infracost/infracost/blob/master/internal/providers/terraform/azure/testdata/kubernetes_cluster_node_pool_test/kubernetes_cluster_node_pool_test.tf

varshneydevansh:
Somewhat like this?

```tf
resource "azurerm_kubernetes_cluster_node_pool" "dv3_standard_d2_v31" {
  name                  = "internal"
  kubernetes_cluster_id = azurerm_kubernetes_cluster.example.id
  vm_size               = "Standard_D2_v31"
}
```

or like this -

```tf
resource "azurerm_kubernetes_cluster_node_pool" "non_premium" {
  name                  = "internal"
  kubernetes_cluster_id = azurerm_kubernetes_cluster.example.id
  vm_size               = "Standard_D2_v31"
}
```

I will add the snippet after your confirmation in the file **kubernetes_cluster_node_pool_test.tf**

aliscott:
> or like this -

This one looks good to me!

aliscott:
Thanks @varshneydevansh this looks good. I've pushed a commit to this just to update the golden files to make sure the tests will pass and then will merge 🚀

aliscott (approved):

PR #2552, opened 2023-07-06
fix: fetch merge request author from gitlab API if GITLAB_TOKEN is available

hugorut:
Fetches the correct author using the /api/v4 endpoint if a valid token is present.


aliscott (approved):
Looks great!

PR #2563, opened 2023-07-11
refactor: change auto detect logic to include tfvars

hugorut:
Changes auto detection logic so that terraform var files are automatically taken into consideration:

1. If a single terraform var file is found under a project, we add it to the evaluator
2. If more than one are found under a project, we split the project into multiples. One for each var file discovered.

This functionality is solely for auto detection and once a user defines a config file, we no longer adhear to these rules.

Also added a `jsoniter` refactor into this pr to speed up the parsing logic. Let me know if you want me to separate it out.

### DIFF
--- a/cmd/infracost/breakdown_test.go
+++ b/cmd/infracost/breakdown_test.go
@@ -842,3 +842,15 @@ func TestBreakdownWithLocalPathDataBlock(t *testing.T) {
 		}, &GoldenFileOptions{CaptureLogs: true},
 	)
 }
+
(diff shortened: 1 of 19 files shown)


aliscott (approved):
Looks great.

Not needed for just now, but this got me thinking if we could do something more intelligent here to determine if the var file is a project-specific var file or a shared/global var file. Maybe this could work by looking at the variable names within the var file and seeing if there's multiple overlapping keys.

PR #2569, opened 2023-07-12
enhance: Add originalLength and truncated attributes to infracost-comment event

vdmgolub:


internal/output/markdown.go
```diff
@@ -266,25 +266,34 @@ func ToMarkdown(out Root, opts Options, markdownOpts MarkdownOptions) ([]byte, e
...
 
 	msgByteLength := len(msg)
+	msgRuneLength := utf8.RuneCount(msg)
+
+	var originalSize int
+	if opts.originalSize > 0 {
+		originalSize = opts.originalSize
+	} else {
+		originalSize = msgRuneLength
+	}
```
hugorut, line 283:
```suggestion
	originalSize := msgRuneLength
	if opts.originalSize > 0 {
		originalSize = opts.originalSize
	}
```

internal/output/markdown.go
```diff
@@ -121,15 +121,15 @@ type MarkdownCtx struct {
 	MarkdownOptions              MarkdownOptions
 }
 
-func ToMarkdown(out Root, opts Options, markdownOpts MarkdownOptions) ([]byte, error) {
+func ToMarkdown(out Root, opts Options, markdownOpts MarkdownOptions) ([]byte, error, int, int) {
```
hugorut, line 124: convention is to always put the error as the last return arg, let's change it to that.

I'm also not a massive fan of having a ton of these return args. Can we change it so we return a struct:

```
type MarkdownOutput struct {
     Msg []byte
     RuneLen  int
     OriginalMsgSize int
}
```

This way it's a bit more readable from a consumer stand point and we don't have to have a ton of return args.

internal/output/markdown.go
```diff
@@ -227,7 +235,7 @@ func ToMarkdown(out Root, opts Options, markdownOpts MarkdownOptions) ([]byte, e
 	})
 	_, err := tmpl.ParseFS(templatesFS, "templates/"+filename)
 	if err != nil {
-		return []byte{}, err
+		return MarkdownOutput{Msg: []byte{}, RuneLen: 0, OriginalMsgSize: 0}, err
```
hugorut, line 238: FYI these are the same as having the zero values, it is essentially equal to `MarkdownOutput{}`

vdmgolub: Oh right! 🤦 Sorry for all the noob issues and thanks for the patience 🙏

hugorut (approved):
LGTM, Test failure looks unrelated

PR #2570, opened 2023-07-13
Consistent mock values

aliscott:
This ensures mocks are calculated deterministically and so resource checksums are also consistent between runs.


aliscott:
Test failure looks unrelated

hugorut (approved):
Do we need to do this with child blocks as well?

aliscott:
> Do we need to do this with child blocks as well?

As far as I can tell the child blocks are already consistently ordered.

PR #2571, opened 2023-07-13
feat: support reporting free/unsupported resource checksums & metadata

hugorut:


hugorut:
Test failures are unrelated - I'm looking into it

PR #2574, opened 2023-07-13
Enhance/enable tag policies api implicitly

tim775:


hugorut (commented):
Not sure I follow exactly why we're making this change? Requiring the URL to be set to toggle `ctx.Config.TagPoliciesEnabled` seems like we'll duplicate the tag policy checks. As we later use and empty check against the URL in comment to check the policies from the CLI. Should we also change these to check against `ctx.Config.TagPoliciesEnabled`? We'll also need to explicitly add the policies URL in the runner env to make sure this works.

tim775:
> Requiring the URL to be set to toggle `ctx.Config.TagPoliciesEnabled` seems like we'll duplicate the tag policy checks. 

I don't think that's quite right.  This does two things 1) if `TagsAPIEnabled` from the cloud settings is true, it makes sure the API URL is set 2) if the API URL is set, it sets `TagPoliciesEnabled=true`. 

I agree it's a little convoluted.  Maybe it would be simpler to remove the `TagPoliciesEnabled` flag and just check for `TagPolicyAPIEndpoint != ""`?

> We'll also need to explicitly add the policies URL in the runner env to make sure this works.

The CLI in runner shouldn't ever use the tag policies API since the runner does the check directly.  Runner sets CLOUD_UPLOAD_ENABLED=false to make sure that's the case.

hugorut (approved):
makes sense

PR #2577, opened 2023-07-17
bug: calculate Google Sustained Use Discount based on monthly_hrs

varshneydevansh:
Apologies for the delay in PR as I got stuck in solving another bug for a different project.

I think it's somewhere close to what we want from the `computeCostComponents()` for the sustained use discount using the `monthly_hrs`


aliscott:
Hi @varshneydevansh, sorry for the delay on this. I've been unable to get to it this week so will take a look next week.

varshneydevansh:
No worries. Next week, I will also begin work on the next feature. :)

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -33,19 +33,21 @@ type ContainerNodeConfig struct {
...
 	if strings.HasPrefix(strings.ToLower(machineType), "e2-custom") {
-		return nil, errors.New("Infracost currently does not support E2 custom instances")
+		return nil, errors.New("infracost currently does not support E2 custom instances")
 	}
 
 	sustainedUseDiscount := 0.0
 	fixPurchaseOption := ""
 
-	if strings.ToLower(purchaseOption) == "on_demand" {
+	if strings.ToLower(purchaseOption) == "on_demand" && monthlyHours != nil {
```
aliscott, line 42: We should probably do something like this instead:
```suggestion
        hours := schema.HourToMonthUnitMultiplier
        if monthlyHours != nil {
                hours = *monthlyHours
        }
        
	if strings.ToLower(purchaseOption) == "on_demand" {
```

And then pass in `hours` to `getSustainedUseDiscount`

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -33,19 +33,21 @@ type ContainerNodeConfig struct {
...
 		fixPurchaseOption = "OnDemand"
 		switch strings.ToLower(strings.Split(machineType, "-")[0]) {
 		case "c2", "n2", "n2d":
-			sustainedUseDiscount = 0.2
+			sustainedUseDiscount = getSustainedUseDiscount(monthlyHours, []float64{0, 0.25, 0.50, 0.75}, []float64{1.0, 0.8678, 0.733, 0.6}, []float64{0.2088, 0.1811, 0.1530, 0.1252})
```
aliscott, line 46: Can we add these values as constants and document where they were received from?

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -33,19 +33,21 @@ type ContainerNodeConfig struct {
...
 		case "custom", "n1", "f1", "g1", "m1":
-			sustainedUseDiscount = 0.3
+			sustainedUseDiscount = getSustainedUseDiscount(monthlyHours, []float64{0, 0.25, 0.50, 0.75}, []float64{1.0, 0.80, 0.60, 0.40}, []float64{0.0475, 0.0380, 0.0285, 0.0190})
+		default:
+			return nil, errors.New("unsupported machine type for sustained use discount")
```
aliscott, line 50: I think if we don't support sustained use discounts for a machine type we should fall back to a 0% discount.

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -202,6 +204,37 @@ func computeCostComponents(region, machineType string, purchaseOption string, in
 
 }
 
+func getSustainedUseDiscount(monthlyHours *float64, thresholds []float64, rates []float64, incrementalRates []float64) float64 {
+	// Assume 730 hours(max) if monthly_hrs is not set
+	if *monthlyHours == 0 {
+		*monthlyHours = 730.0
+	}
+	totalHoursInMonth := 730.0
```
aliscott, line 212:
```suggestion
func getSustainedUseDiscount(monthlyHours float64, thresholds []float64, rates []float64, incrementalRates []float64) float64 {
```

If we move the logic to fallback to `schema.HourToMonthUnitMultiplier` above, then we shouldn't need this and we can pass in the hours as a non-pointer float64. If monthlyHours is 0, we should still support that.

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -202,6 +204,37 @@ func computeCostComponents(region, machineType string, purchaseOption string, in
...
+	totalDiscount := 0.0
+	remainingHours := *monthlyHours
+
+	for i, threshold := range thresholds {
```
aliscott, line 216: We should check that the lengths of thresholds, rates and incrementalRates all match to avoid any index errors.

aliscott: We should also check that the lengths are all greater than 0, and if not then return 0.0.

aliscott (commented):
Thanks @varshneydevansh, this looks good. I've added some comments. It would be good make sure there's some tests that cover these discounts in the `compute_instance_test.tf` file.

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -33,19 +33,21 @@ type ContainerNodeConfig struct {
 // computeCostComponent returns a cost component for Compute instance usage.
 func computeCostComponents(region, machineType string, purchaseOption string, instanceCount int64, monthlyHours *float64) ([]*schema.CostComponent, error) {
 	if strings.HasPrefix(strings.ToLower(machineType), "e2-custom") {
-		return nil, errors.New("Infracost currently does not support E2 custom instances")
+		return nil, errors.New("infracost currently does not support E2 custom instances")
```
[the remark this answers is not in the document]
varshneydevansh: Actually, the golang checker was showing me an error for using the Uppercase character, that's why I made it small case.

internal/resources/google/compute_cost_component_helpers.go
varshneydevansh: I was considering storing them in a struct. Let me made this change and also add a comment about the values.

internal/resources/google/compute_cost_component_helpers.go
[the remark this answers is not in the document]
varshneydevansh: image

varshneydevansh:
I have a query regarding the incremental rate, the values which I have taken are from the example -

image

How do we look for price if it got changed by the GCP?

varshneydevansh:
> Thanks @varshneydevansh, this looks good. I've added some comments. It would be good make sure there's some tests that cover these discounts in the `compute_instance_test.tf` file.

I will create the test cases as soon as these changes are confirmed by you and the query which I have regarding the pricing.

internal/resources/google/compute_cost_component_helpers.go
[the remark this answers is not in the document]
aliscott: I don't think go-staticcheck is part of the linting we do, but this should be an exception to that.

aliscott:
> How do we look for price if it got changed by the GCP?

@varshneydevansh, sorry it's not clear to me what you mean by this. Can you please describe the issue in more details.

aliscott:
Apologies for the slow responses @varshneydevansh, I've been out of office.

varshneydevansh:
> > How do we look for price if it got changed by the GCP?
> 
> @varshneydevansh, sorry it's not clear to me what you mean by this. Can you please describe the issue in more details.

The values in the `incremental rate` list which I have made in the code have been taken from the table below -

image


and for both **Sustained use discounts for up to 30% and 20%** it uses the examples - `example: n1-standard-1 VM` and `example: c2-standard-4 VM`

So, my question is the values for the 2 discounts are from the example of different VM types, and how are we gonna make sure that the calculations for the `"custom", "n1", "f1", "g1", "m1"` and `"c2", "n2", "n2d"` are correct?

Do we need to store the values for each type in separate lists? This is the where I am a little lost.

---------------

> Apologies for the slow responses @varshneydevansh, I've been out of office.
>
This is what I was telling my friends that you guys probably out of the office or maybe busy with some other work. When they asked me how my work was going on.

aliscott:
> So, my question is the values for the 2 discounts are from the example of different VM types, and how are we gonna make sure that the calculations for the "custom", "n1", "f1", "g1", "m1" and "c2", "n2", "n2d" are correct?
> Do we need to store the values for each type in separate lists? This is the where I am a little lost.

I would have two different lists: one for sustained usage discounts up to 20% and another for sustained usage discounts up to 30%.

We can then add two test cases: one to cover each list.

Does that make sense?

varshneydevansh:
> I would have two different lists: one for sustained usage discounts up to 20% and another for sustained usage discounts up to 30%.
> 
> We can then add two test cases: one to cover each list.

Maybe I am overthinking this price thing. I did create these two lists based on the line - 

image
[Sustained use Discount](https://cloud.google.com/compute/docs/sustained-use-discounts) | [VM Pricing GCP](https://cloud.google.com/compute/all-pricing?_ga=2.213835474.711896682.1691408112-2046117023.1688470863)


```go
	VarSudUptoTwenty := SudValues{
		thresholds:       [size_SUD]float64{0, 0.25, 0.50, 0.75},
		rates:            [size_SUD]float64{1.0, 0.8678, 0.733, 0.6},
		incrementalRates: [size_SUD]float64{0.2088, 0.1811, 0.1530, 0.1252},
	}

	VarSudUptoThirty := SudValues{
		thresholds:       [size_SUD]float64{0, 0.25, 0.50, 0.75},
		rates:            [size_SUD]float64{1.0, 0.80, 0.60, 0.40},
		incrementalRates: [size_SUD]float64{0.0475, 0.0380, 0.0285, 0.0190},
	}

```
After modifying the code on that day, I just started to rethink whether the code which I wrote is future-proof with the changes or not?

varshneydevansh:
Could anyone help me with this, as it is almost complete?

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -30,6 +30,21 @@ type ContainerNodeConfig struct {
...
+size_SUD is set to 4 as the Usage level (% of month) is 4
+	to make sure we don't overshoot the array
+*/
+const size_SUD = 4
+
+type SudValues struct {
+	thresholds       [size_SUD]float64
+	rates            [size_SUD]float64
+	incrementalRates [size_SUD]float64
+}
```
aliscott, line 46:
```suggestion
sudRateSize is set to 4 as the Usage level (% of month) is 4
	to make sure we don't overshoot the array
*/
```suggestion
const sudRateSize = 4

type sudRates struct {
	thresholds       [sudRateSize]float64
	rates            [sudRateSize]float64
	incrementalRates [sudRateSize]float64
}
```

Updating var names to be more consistent with the format we use elsewhere (camelCase)

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -38,14 +53,37 @@ func computeCostComponents(region, machineType string, purchaseOption string, in
...
+		thresholds:       [size_SUD]float64{0, 0.25, 0.50, 0.75},
+		rates:            [size_SUD]float64{1.0, 0.8678, 0.733, 0.6},
+		incrementalRates: [size_SUD]float64{0.2088, 0.1811, 0.1530, 0.1252},
+	}
+
+	VarSudUptoThirty := SudValues{
+		thresholds:       [size_SUD]float64{0, 0.25, 0.50, 0.75},
+		rates:            [size_SUD]float64{1.0, 0.80, 0.60, 0.40},
+		incrementalRates: [size_SUD]float64{0.0475, 0.0380, 0.0285, 0.0190},
+	}
```
aliscott, line 76:
```suggestion
	sudRate20 := sudRates{
		thresholds:       [sudRateSize]float64{0, 0.25, 0.50, 0.75},
		rates:            [sudRateSize]float64{1.0, 0.8678, 0.733, 0.6},
		incrementalRates: [sudRateSize]float64{0.2088, 0.1811, 0.1530, 0.1252},
	}

	sudRate30 := sudRates{
		thresholds:       [sudRateSize]float64{0, 0.25, 0.50, 0.75},
		rates:            [sudRateSize]float64{1.0, 0.80, 0.60, 0.40},
		incrementalRates: [sudRateSize]float64{0.0475, 0.0380, 0.0285, 0.0190},
	}
```

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -38,14 +53,37 @@ func computeCostComponents(region, machineType string, purchaseOption string, in
...
 
 	if strings.ToLower(purchaseOption) == "on_demand" {
 		fixPurchaseOption = "OnDemand"
 		switch strings.ToLower(strings.Split(machineType, "-")[0]) {
 		case "c2", "n2", "n2d":
-			sustainedUseDiscount = 0.2
+			sustainedUseDiscount = getSustainedUseDiscount(hours, VarSudUptoTwenty)
```
aliscott, line 82:
```suggestion
			sustainedUseDiscount = getSustainedUseDiscount(hours, sudRate20)
```

internal/resources/google/compute_cost_component_helpers.go
```diff
@@ -38,14 +53,37 @@ func computeCostComponents(region, machineType string, purchaseOption string, in
...
 		case "custom", "n1", "f1", "g1", "m1":
-			sustainedUseDiscount = 0.3
+			sustainedUseDiscount = getSustainedUseDiscount(hours, VarSudUptoThirty)
```
aliscott, line 84:
```suggestion
			sustainedUseDiscount = getSustainedUseDiscount(hours, sudRate30)
```

aliscott (commented):
Thanks @varshneydevansh 🙏. This seems more complicated than I originally anticipated. When I run `make test_google` I'm seeing tests fail with instances showing negative costs.

I think the calculation for `getSustainedUseDiscount` is incorrect.

From what I can tell from the [docs](https://cloud.google.com/compute/docs/sustained-use-discounts) the `incremental rate` is just an example of what the hourly rate would be for a n1-standard-1 instance and so we shouldn't be using that in our calculation.

The total discount rate can then be calculated by working out the total hours that are charged, and then dividing that by the total hours the instance is running. So something like this:


```
// thesholds is the percentage of the full monthly hours up to which the rate applies
// rates is the percentage of the rate that is charged for that threshold.
// e.g.
// - if an instance runs for 0.25 of the month it is charged 100% of the rate.
// - if an instance runs for 0.5 of the month, the first 0.25 of the month is charged at 100% of the rate, the second 0.25 of the month is charged at 86.78% of the rate.
sudRates20 := sudRates{
        thresholds: [sudRatesSize]float64{0.25, 0.5, 0.75, 1.0},
        rates:      [sudRatesSize]float64{1.0, 0.8678, 0.733, 0.6},
}

sudRates30 := sudRates{
        thresholds: [sudRatesSize]float64{0.25, 0.5, 0.75, 1.0},
        rates:      [sudRatesSize]float64{1.0, 0.80, 0.60, 0.40},
}
```
...
```
func getSustainedUseDiscount(hours float64, rates sudRates) float64 {
	totalHoursInMonth, _ := schema.HourToMonthUnitMultiplier.Float64()
	
	// Keep track of how many hours we have remaining after each threshold is applied
	remainingHours := hours
	
	// Keep track of the total hours that are charged for
	ratedHours := 0.0

	index := 0
	for remainingHours > 0 {
		// Calculate the percentage of the month that is covered by the current threshold
		lastThreshold := 0.0
		if index > 0 {
			lastThreshold = rates.thresholds[index-1]
		}

		thresholdHours := (rates.thresholds[index] - lastThreshold) * totalHoursInMonth

		// If the remaining hours are less than the threshold, add them and then we are done
		if remainingHours <= thresholdHours {
			ratedHours += remainingHours * rates.rates[index]
			remainingHours = 0
			break
		}

		// Otherwise, add the discount for the current threshold and continue
		ratedHours += thresholdHours * rates.rates[index]
		remainingHours -= thresholdHours
		index++
	}

	// Return the average discount over the hours the instance is running
	return 1 - (ratedHours / hours)
}
```

Let me know if you think this makes sense. It would be good to add some test cases for the 
`getSustainedUseDiscount` function as well to make sure it is working as expected.

varshneydevansh:
> ```go
> func getSustainedUseDiscount(hours float64, rates sudRates) float64 {
> 	totalHoursInMonth, _ := schema.HourToMonthUnitMultiplier.Float64()
> 	
> 	// Keep track of how many hours we have remaining after each threshold is applied
> 	remainingHours := hours
> 	
> 	// Keep track of the total hours that are charged for
> 	ratedHours := 0.0
> 
> 	index := 0
> 	for remainingHours > 0 {
> 		// Calculate the percentage of the month that is covered by the current threshold
> 		lastThreshold := 0.0
> 		if index > 0 {
> 			lastThreshold = rates.thresholds[index-1]
> 		}
> 
> 		thresholdHours := (rates.thresholds[index] - lastThreshold) * totalHoursInMonth
> 
> 		// If the remaining hours are less than the threshold, add them and then we are done
> 		if remainingHours <= thresholdHours {
> 			ratedHours += remainingHours * rates.rates[index]
> 			remainingHours = 0
> 			break
> 		}
> 
> 		// Otherwise, add the discount for the current threshold and continue
> 		ratedHours += thresholdHours * rates.rates[index]
> 		remainingHours -= thresholdHours
> 		index++
> 	}
> 
> 	// Return the average discount over the hours the instance is running
> 	return 1 - (ratedHours / hours)
> }
> ```
> 
> Let me know if you think this makes sense. It would be good to add some test cases for the `getSustainedUseDiscount` function as well to make sure it is working as expected.

This logic makes sense to me, and I will be more careful with the implementations next time. =)

- Regarding the test cases, I did add a few in the `computer_instance_test.tf` are those what we intended for?

aliscott:
Thanks @varshneydevansh. I've pushed a commit just to fix up some tests based on this change. I'll merge it once this PR has finished building 🙌

varshneydevansh:
> Thanks @varshneydevansh. I've pushed a commit just to fix up some tests based on this change. I'll merge it once this PR has finished building 🙌

I was thinking of asking for the update yesterday. =)

I'll now look for some other task.

aliscott (approved):

PR #2584, opened 2023-07-24
refactor: use terragrunt global cache for modules

hugorut:
This change uses the updated infracost Terragrunt DownloadTerraformSource which:

* prevents duplicate downloads for remote and local modules. Terragrunt projects now use a global cache of modules.
* improves local module caching logic. We now don't recopy the a local module if an up to date hash of the module exists in the cache


aliscott (approved):
lovely jubbly!

PR #2587, opened 2023-07-25
Skip commenting if run was earlier

aliscott:
This now shows a warning and skips posting the comment if the run was generated before the latest comment that was posted.

It does this by embedding the generated at timestamp of the run in the comment.


internal/output/combined.go
```diff
@@ -282,7 +287,7 @@ func Combine(inputs []ReportInput) (Root, error) {
 	combined.PastTotalMonthlyCost = pastTotalMonthlyCost
 	combined.DiffTotalHourlyCost = diffTotalHourlyCost
 	combined.DiffTotalMonthlyCost = diffTotalMonthlyCost
-	combined.TimeGenerated = time.Now().UTC()
+	combined.TimeGenerated = lastestGeneratedAt
```
aliscott, line 290: I'm not 100% sure this won't cause issues, but don't see the harm in using the lastest generated at instead of the current time for the combined run data.

internal/comment/handler.go
```diff
@@ -114,107 +138,176 @@ func (h *CommentHandler) matchingComments(ctx context.Context) ([]Comment, error
...
 // NewComment creates a new comment with the given body.
-func (h *CommentHandler) NewComment(ctx context.Context, body string) error {
-	bodyWithTag := h.PlatformHandler.AddMarkdownTag(body, h.Tag)
+func (h *CommentHandler) NewComment(ctx context.Context, body string, opts *CommentOpts) error {
+	var validAt *time.Time
+	if opts != nil {
+		validAt = opts.ValidAt
+
+		if opts.SkipNoDiff {
+			log.Warning("SkipNoDiff option is not supported for new comments")
```
hugorut, line 232: `logging.Logger` please, and let's change all the rest. Thanks. (note, this isn't me just being a pain, it should hopefully make migration away from logrus easier in the future)

internal/comment/tag.go
```diff
@@ -1,19 +1,99 @@
...
+
+	match := re.FindString(body)
+	return match != ""
+}
+
+// extractTagValue extracts the value of a tag from the given body.
+// It returns an empty string if the tag is not found.
+func extractTagValue(body, key string) string {
+	pattern := fmt.Sprintf(`\[\/\/\]: <> \([^)]*\s*%s=([^,\)]+)`, key)
+	re, err := regexp.Compile(pattern)
```
hugorut, line 72: can we move this out of the function body, so it's not compiled each time the function is run. + use `regex.MustCompile`

internal/output/combined.go
hugorut: I don't see a problem as well, famous last words.

hugorut (approved):
minor tweaks

internal/comment/tag.go
aliscott: Not easily - this pattern is generated based on the key passed in.

hugorut: ah yep - ignore me then

PR #2589, opened 2023-07-30
fix: Strip count from module name as configuration key when marshalling

vdmgolub:
Default tags were not applied to the resources of a module with count:
the configuration key contained the count part, but the lookup used just
the module name. Thus the tags were not found.

### DIFF
--- a/internal/providers/terraform/hcl_provider.go
+++ b/internal/providers/terraform/hcl_provider.go
@@ -491,7 +491,7 @@ func (p *HCLProvider) marshalModule(module *hcl.Module) ModuleOut {
 	})
 
 	for _, m := range module.Modules {
-		pieces := strings.Split(m.Name, ".")
+		pieces := strings.Split(removeAddressArrayPart(m.Name), ".")
 		modKey := pieces[len(pieces)-1]
 
 		mo := p.marshalModule(m)
--- a/internal/providers/terraform/hcl_provider_test.go
+++ b/internal/providers/terraform/hcl_provider_test.go
@@ -179,6 +179,9 @@ func TestHCLProvider_LoadPlanJSON(t *testing.T) {
 		{
 			name: "shows correct duplicate variable warning",
 		},
+		{
+			name: "builds module configuration correctly with count",
+		},
 	}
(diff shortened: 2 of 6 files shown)


internal/providers/terraform/parser.go
```diff
@@ -720,19 +720,18 @@ func addressModulePart(addr string) string {
...
-		n = append(n, m[1])
+	result := []string{}
+	for _, p := range strings.Split(modulesPart, ".module.") {
+		if p != "" {
+			result = append(result, p)
+		}
 	}
-
-	return n
+	return result
```
vdmgolub, line 734: @hugorut I'm 99% sure there should be a better way to do it, please advise. After all my debugging it looks like the parsed module name keeps the index, so I changed this method.

@aliscott Do you happen to remember why you excluded the indexes from module names? I'm worried my change can break something else, so more context could help :) Thanks.

hugorut: @vdmgolub this does not appear to be the solution to the actual problem. The module config in an actual plan JSON does not appear to have the keys/indexs. For example take the following terraform:

```
provider "aws" {
  ...
}

module "test" {
  count  = 2
  source = "./modules/test"
}
```

the `configuration.root_module` has the following in the plan JSON:

```
    "root_module": {
      "module_calls": {
        "test": {
           ...
          "module": {...}
         }
      } 
```

wheras when we run the HCL provider it builds the following json output:

```
    "root_module": {
      "module_calls": {
        "test[0]": {
          ...
          "module": {...}
        },
        "test[1]": {
          ...
           "module": {...}
        }
     }
```

See how it is incorrectly adding the keys to the configuration section.

Hence the fix relies not on altering this function but in the `HCLProvider.marshalModule` functionality.

hugorut (changes requested):

internal/providers/terraform/parser.go
vdmgolub: Thanks! I missed comparing the JSON plans 🤦. Here's the update.

hugorut (approved):

PR #2600, opened 2023-08-04
feat: add support for launch_template tags

hugorut:
Propagates tag_specifications to aws_instance and aws_autoscaling_group from launch_templates.

### DIFF
--- a/internal/providers/terraform/parser.go
+++ b/internal/providers/terraform/parser.go
@@ -111,6 +111,8 @@ func (p *Parser) parseJSONResources(parsePrior bool, baseResources []*schema.Par
 	resData := p.parseResourceData(isState, providerConf, vals, conf, vars)
 
 	p.parseReferences(resData, conf)
+	p.parseTags(resData, providerConf, conf)
+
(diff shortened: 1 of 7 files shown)


internal/providers/terraform/parser.go
```diff
@@ -648,6 +631,28 @@ func (p *Parser) parseConfReferences(resData map[string]*schema.ResourceData, co
 	return found
 }
 
+func (p *Parser) parseTags(data map[string]*schema.ResourceData, providerConf gjson.Result, conf gjson.Result) {
+	for _, resourceData := range data {
+		resConf := getConfJSON(conf, resourceData.Address)
+		defaultTags := parseDefaultTags(providerConf, resConf)
```
vdmgolub, line 637: These two lines are related to AWS only. Can we move them under `case "aws"`?

vdmgolub (approved):

internal/providers/terraform/parser.go
hugorut: good point

PR #2607, opened 2023-08-08
fix: ensure that terragrunt remote source modules with refs and versions are downloaded to different directories

hugorut:
Resolves issue where modules using refs where being downloaded to same directory. We now encode the ref/query into the directory hash: https://github.com/infracost/terragrunt/pull/4/commits/a0a234c6387888404e78489071bc78cde25abd06 so the directories are unique. The source lock is updated to use the download directory that terragrunt checks to ensure that goroutines are synced with the downloading logic.


aliscott (approved):
Woop!

PR #2608, opened 2023-08-08
fix: ensure project summaries are set the original summary

aliscott:
Previously this was not persisting since it was set on a copy of the project.


internal/output/combined.go
```diff
@@ -172,13 +172,13 @@ func CompareTo(c *config.Config, current, prior Root) (Root, error) {
 
 	// preserve the summary from the original run
 	currentProjects := make(map[string]Project)
-	for _, p := range prior.Projects {
+	for _, p := range current.Projects {
```
aliscott, line 175: I think `current` makes more sense here, since then the summaries sum to match the overall summaries, which are also set using `current`.

hugorut (approved):

PR #2614, opened 2023-08-10
refactor: add backoff retry logic to comment handler

hugorut:
Retries comment logic if the handler receives an unexpected error.


aliscott (approved):
Looks like a good approach. I think the other option would be to use `retryablehttp` like we do for the Pricing API client.

PR #2618, opened 2023-08-15
fix: avoid race conditions by only downloading tg sources once

tim775:
@hugorut I think this is ok, but can you confirm that it's correct to only download the terragrunt sourceURL once?

### DIFF
--- a/internal/providers/terraform/terragrunt_hcl_provider.go
+++ b/internal/providers/terraform/terragrunt_hcl_provider.go
@@ -45,7 +45,7 @@ var terragruntSourceLock = infSync.KeyMutex{}
 
 // terragruntDownloadedDirs is used to ensure sources are only downloaded once. Use terragruntSourceLock
 // for concurrency safety.
-var terragruntDownloadedDirs = map[string]bool{}
+var terragruntDownloadedDirs = sync.Map{}
 
 type panicError struct {
 	msg string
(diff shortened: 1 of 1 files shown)


hugorut (commented):
I could be wrong, but I thought the code in `DownloadTerraformSource` handles this logic (checking if the source is already downloaded). See: https://github.com/infracost/terragrunt/blob/f27e16c7e6bd558b1981e10ff8b35a7c7642d8d9/cli/download_source.go#L40

internal/providers/terraform/terragrunt_hcl_provider.go
```diff
@@ -43,6 +43,10 @@ import (
 // concurrency safe downloading.
 var terragruntSourceLock = infSync.KeyMutex{}
 
+// terragruntDownloadedDirs is used to ensure sources are only downloaded once. Use terragruntSourceLock
+// for concurrency safety.
+var terragruntDownloadedDirs = map[string]bool{}
```
hugorut, line 48:
```suggestion
var terragruntDownloadedDirs = map[string]bool{}
```

Maps are not concurrent safe this will need to be `sync.Map`

hugorut (changes requested):

tim775:
> I could be wrong, but I thought the code in `DownloadTerraformSource` handles this logic (checking if the source is already downloaded). See: https://github.com/infracost/terragrunt/blob/f27e16c7e6bd558b1981e10ff8b35a7c7642d8d9/cli/download_source.go#L40

It's not thread safe so the call to `util.CopyFolderContents` is causing the race condition (one project is trying to parse while the other project is copying over).  It does check to see if it has already downloaded the latest sources, but that's not thread safe either so there is also a theoretical race condition if anything needs to be downloaded.

hugorut:
@tim775 I'm not sure I'm following - we lock the source for interaction so why would that matter with util copy contents. This sounds like it's an issue with the lock key rather than the redownload.

PR #2628, opened 2023-08-24
Add il-central-1 region

aliscott:


aliscott:
Test failures are unrelated.

internal/resources/aws/global_accelerator_endpoint_group.go
```diff
@@ -61,6 +61,7 @@ var regionCodeMapping = map[string]string{
 	"eu-south-1":      "EU",
 	"eu-west-3":       "EU",
 	"eu-north-1":      "EU",
+	"il-central-1":    "EU",
```
vdmgolub, line 64: Should this be the Middle East? https://aws.amazon.com/about-aws/global-infrastructure/regions_az/?p=ngi&loc=2
```suggestion
	"il-central-1":    "ME",
```

Although I don't know what mapping should be for Cloudfront 🤔

aliscott: AWS groups Israel with Europe for a lot of the pricing (see https://aws.amazon.com/global-accelerator/pricing/). Not sure why - maybe they are using Eurovision geography.

internal/comment/handler.go
```diff
@@ -117,7 +117,7 @@ func (h *CommentHandler) CommentWithBehavior(ctx context.Context, behavior, body
 			break
 		}
 
-		logging.Logger.WithError(err).Debugf("received an error trying to post comment pausing %d seconds then will retry", backoff.Seconds())
+		logging.Logger.WithError(err).Debugf("received an error trying to post comment pausing %v seconds then will retry", backoff.Seconds())
```
hugorut, line 120:
```suggestion
		logging.Logger.WithError(err).Debugf("received an error trying to post comment pausing %2f seconds then will retry", backoff.Seconds())
```

hugorut (approved):

internal/comment/handler.go
aliscott: Any reason why? I think I prefer `%v` or `%g` for these cases so a `float64(1.0)` just gets printed as `1` instead of `1.00`.

hugorut: no worries - %v just uses reflection so it's slightly slower, but sounds good

aliscott: image

PR #2629, opened 2023-08-24
feat: pricing client cache

hugorut:
Introduces an persisent and in memory query cache for the pricing api. As part of this change the PricingApiClient is now shared across the application codebase so that we maintain a consistent cache used across goroutines. Furthermore, we now dedupe pricing api queries sent to the pricing API, further reducing request size made by the CLI.

Caching functionality can be controlled by the new configuration parameters:

* `INFRACOST_PRICING_CACHE_DISABLED` which, if set to true, completely removes caching functionality
* `INFRACOST_PRICING_CACHE_OBJECT_SIZE` which can control the size of the cache. This is useful for memory requirements.

### DIFF
--- a/internal/apiclient/pricing.go
+++ b/internal/apiclient/pricing.go
@@ -181,6 +181,8 @@ func (c *PricingAPIClient) FlushCache() error {
 		return nil
 	}
 
+	logging.Logger.Debugf("writing %d objects to filesystem cache", len(c.cache))
+
 	// we store the cache as a string instead of the gjson.Result so the size is
 	// smaller on the filesystem and gob encoding doesn't have to work as hard.
 	storedCache := map[uint64]string{}
@@ -304,12 +306,17 @@ func (c *PricingAPIClient) PerformRequest(req BatchRequest) ([]PriceQueryResult,
 	// first filter any queries that have been stored in the cache. We don't need to
 	// send requests for these as we already have the results in memory.
 	var serverQueries []pricingQuery
-	if c.cache != nil {
+	if c.cache == nil {
+		serverQueries = queries
+	} else {
+		var hit int
 		for i, query := range queries {
 			c.mu.RLock()
 			v, ok := (*c.cache)[query.hash]
 			c.mu.RUnlock()
 			if ok {
+				logging.Logger.Debugf("cache hit for query hash: %d", query.hash)
+				hit++
 				res[i] = PriceQueryResult{
 					PriceQueryKey: req.keys[i],
 					Result:        v,
@@ -319,6 +326,8 @@ func (c *PricingAPIClient) PerformRequest(req BatchRequest) ([]PriceQueryResult,
 				serverQueries = append(serverQueries, query)
 			}
 		}
+
+		logging.Logger.Debugf("%d/%d queries were built from cache", hit, len(queries))
 	}
 
 	// now we deduplicate the queries, ensuring that a request for a price only happens once.
@@ -344,7 +353,7 @@ func (c *PricingAPIClient) PerformRequest(req BatchRequest) ([]PriceQueryResult,
 	}
 
 	// if the cache is enabled lets store each pricing result returned in the cache.
-	if *c.cache != nil {
+	if c.cache != nil {
 		for i, query := range deduplicatedServerQueries {
 			c.mu.RLock()
 			keyLength := len(*c.cache)
--- a/internal/apiclient/pricing.go
+++ b/internal/apiclient/pricing.go
@@ -3,12 +3,17 @@ package apiclient
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/gob"
 	"fmt"
 	"math"
 	"net/http"
 	"os"
+	"path/filepath"
+	"sync"
+	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/mitchellh/hashstructure/v2"
 
 	"github.com/infracost/infracost/internal/config"
 	"github.com/infracost/infracost/internal/logging"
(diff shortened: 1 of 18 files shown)


internal/apiclient/pricing.go
```diff
@@ -169,26 +270,123 @@ func (c *PricingAPIClient) BatchRequests(resources []*schema.Resource, batchSize
...
+			hash:  key,
+			query: query,
+		}
+	}
 
-	results, err := c.doQueries(req.queries)
+	// first filter any queries that have been stored in the cache. We don't need to
+	// send requests for these as we already have the results in memory.
+	var serverQueries []pricingQuery
+	if c.cache != nil {
```
aliscott, line 307:
```suggestion
	var serverQueries []pricingQuery
	if c.cache == nil {
		serverQueries = queries
	} else {
```

internal/apiclient/pricing.go
```diff
@@ -169,26 +270,123 @@ func (c *PricingAPIClient) BatchRequests(resources []*schema.Resource, batchSize
...
+	}
+	resultsFromServer, err := c.doQueries(rawQueries)
 	if err != nil {
 		return []PriceQueryResult{}, err
 	}
 
-	return c.zipQueryResults(req.keys, results), nil
-}
+	// if the cache is enabled lets store each pricing result returned in the cache.
+	if *c.cache != nil {
```
aliscott, line 347:
```suggestion
	if c.cache != nil {
```
This will cause a nil ptr error if c.cache is nil.

internal/apiclient/pricing.go
```diff
@@ -169,26 +270,123 @@ func (c *PricingAPIClient) BatchRequests(resources []*schema.Resource, batchSize
...
+			c.mu.RUnlock()
+			if ok {
+				res[i] = PriceQueryResult{
+					PriceQueryKey: req.keys[i],
+					Result:        v,
+					filled:        true,
+				}
+			} else {
+				serverQueries = append(serverQueries, query)
+			}
```
aliscott, line 320: Should we debug log if it's a cache hit or not, so we can look through the logs to see what % hit rate we are getting?

internal/apiclient/pricing.go
```diff
@@ -93,6 +118,82 @@ func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
...
+		cache[k] = gjson.Parse(raw)
+	}
+
+	return cache
+}
+
+// FlushCache writes the in memory cache to the filesystem. This allows the cache
+// to be persisted between runs. FlushCache should only be called once, at
+// program termination.
+func (c *PricingAPIClient) FlushCache() error {
```
aliscott, line 179: Would be good to debug log the size of the objects in the cache when we write it.

aliscott (commented):
Tested this on a smallish repo:
```
no cache: 10.442s
cache enabled, first run: 8.355s, 32% hit rate
cache enabled, second run: 7.148s, 100% hit rate
objects in cache: 91
time to load cache: 5.174µs
time to flush cache: 504.213µs
total size of cache on disk: 13555 bytes = 14kb
```

My current thinking is we could default the cache size to 1000 objects. If we can assume we can extrapolate the above numbers then we'd get the below which seems low impact:
```
time to load cache: 56.857µs
time to flush cache: 5540.802µs = 5.5ms
total size of cache on disk: 148956 bytes = 145kb
```

internal/apiclient/pricing.go
```diff
@@ -93,6 +118,82 @@ func NewPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
...
+	cache := make(map[uint64]gjson.Result)
+	info, err := os.Stat(cacheFile)
+	if err != nil {
+		return cache
+	}
+
+	// if the cache is older than a day don't use it
+	if info.ModTime().Before(time.Now().AddDate(0, 0, -1)) {
+		return cache
+	}
```
aliscott, line 154: Will this mean that if a user continuously uses cached queries every 20 hours then they never expire since we load them from the cache and write them to a new cache file each time?

hugorut: yup makes sense, we've disccuseed and I'm changing the Implementation

internal/apiclient/pricing.go
```diff
@@ -169,26 +282,118 @@ func (c *PricingAPIClient) BatchRequests(resources []*schema.Resource, batchSize
...
+		deduplicatedServerQueries = append(deduplicatedServerQueries, query)
+		seenQueries[query.hash] = true
+	}
+
+	// send the deduplicated queries to the pricing API to fetch live prices.
+	rawQueries := make([]GraphQLQuery, len(deduplicatedServerQueries))
+	for i, query := range deduplicatedServerQueries {
+		rawQueries[i] = query.query
+	}
+	resultsFromServer, err := c.doQueries(rawQueries)
```
aliscott, line 358: Not sure if this is easy or not, so might not be worth it. To reduce the round trips it would make more sense to me to do cache lookups and deduping before we batch the queries. So we only batch the queries we actually need to make requests for.

aliscott (approved):
🙌 Here's my results from running this on a small repo.

The timing results are different from before because I ran them in a different way, but it's interesting that the timings to load and flush are a lot less.

Also not sure why the hit rate is different.
```
no cache: 4.31s
cache enabled, first run: 3.83s, 50% hit rate
cache enabled, second run: 3.45s, 100% hit rate
objects in cache: 91
time to load cache: 411ns
time to flush cache: 483ns
total size of cache on disk: 16123 bytes = 16kb
```

internal/apiclient/pricing.go
```diff
@@ -169,26 +282,118 @@ func (c *PricingAPIClient) BatchRequests(resources []*schema.Resource, batchSize
...
+
+	// first filter any queries that have been stored in the cache. We don't need to
+	// send requests for these as we already have the results in memory.
+	var serverQueries []pricingQuery
+	if c.cache == nil {
+		serverQueries = queries
+	} else {
+		var hit int
+		for i, query := range queries {
+			v, ok := c.cache.Get(query.hash)
```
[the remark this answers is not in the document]
hugorut: yeah exactly, I think it's fine not to check here.

internal/apiclient/pricing.go
hugorut: yup I agree, but the way that the queries are tied to the `result`/`cost component` makes this tricky. As we expect the results to match the exact index of the cost component. I think it's something we can optimise later, but I'm going to leave out for now.

PR #2630, opened 2023-08-24
fix: attempt to handle un-Rel-able paths

tim775:


hugorut (approved):
This looks good - but is there any way we can add a test to cover the issue. I.e. forcing cache path to be `.` and seeing if we get the error before the change and if the change fixes

tim775:
> This looks good - but is there any way we can add a test to cover the issue. I.e. forcing cache path to be `.` and seeing if we get the error before the change and if the change fixes

I didn't see an easy way to add a test case and I haven't been able to reproduce the error so this is a "release it and see" kind of situation.  I think the change does no harm since we're just trying to recover from the initial error.

PR #2642, opened 2023-08-31
fix: add timeout for hcpapi requests

tim775:
@aliscott I'm not sure what you were thinking for the timeout.  I looked at cloudwatch for the max TargetResponseTime from the HCPAPI ELBs and 60s covers all outliers, but it does seem ridicuously high.


internal/apiclient/pricing.go
```diff
@@ -116,6 +116,7 @@ func GetPricingAPIClient(ctx *config.RunContext) *PricingAPIClient {
 	client := retryablehttp.NewClient()
 	client.Logger = &LeveledLogger{Logger: logging.Logger.WithField("library", "retryablehttp")}
 	client.HTTPClient.Transport.(*http.Transport).TLSClientConfig = &tlsConfig
+	client.HTTPClient.Timeout = time.Minute
```
aliscott, line 119:
```suggestion
	client.HTTPClient.Timeout = time.Second * 30
```

aliscott (approved):

PR #2652, opened 2023-09-01
enhance: support google user_labels as tags

tim775:


internal/providers/terraform/google/google.go
```diff
@@ -37,15 +37,30 @@ func GetResourceRegion(resourceType string, v gjson.Result) string {
...
+	rUserLabels := r.Get("user_labels").Map()
+
+	_, supportsSettingsUserLabels := provider_schemas.GoogleSettingsUserLabelsSupport[r.Type]
+	rSettingsUserLabels := r.Get("settings.0.user_labels").Map()
+
+	if !supportsLabels && len(rLabels) == 0 &&
+		!supportsUserLabels && len(rUserLabels) == 0 &&
+		!supportsSettingsUserLabels && len(rSettingsUserLabels) == 0 {
 		return nil
 	}
```
hugorut, line 53: I'm a little unsure why we are checking both if the labels are supported and the length of the labels. Surely just checking if they are supported is sufficient? If it is not, I'm unsure why we actually use the supported schemas?

internal/providers/terraform/provider_schemas/tag_support.go
```diff
@@ -51,4 +59,14 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
+
+	err = json.Unmarshal(googleUserLabelsJSON, &GoogleUserLabelsSupport)
```
hugorut, line 63: can we replace this with the `"github.com/json-iterator/go"` unmarshall as it is prefered due to it's performance.

internal/providers/terraform/google/google.go
tim775: This is just a small safety factor since we don't try to match the provider version used by the tf being parsed.  I don't think it hurts anything.

PR #2653, opened 2023-09-01
chore: remove verbose flag from make test_*

tim775:
Does anyone object to removing the `-v` option from our makefile rules?  I feel like it takes a bunch of scrolling or  grepping to find test failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,57 +67,57 @@ clean:
 
 # Run only short unit tests
 test:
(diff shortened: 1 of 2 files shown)


vdmgolub:
I'm not against it, but I think we should keep it in CI. Would be great to modify it so we can pass `-v` in the CI workflow.

vdmgolub:
@tim775 I just checked: when something fails it will output. Yeah, with that `-v` seems to be just noise 👍

aliscott:
+1 from me.

hugorut (approved):
👍

PR #2661, opened 2023-09-13
feat: Add policy v2

tim775:
Adds the policy v2 client to prepare for additional policy scanning.

### DIFF
--- a/internal/apiclient/policyv2.go
+++ b/internal/apiclient/policyv2.go
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/hashicorp/go-retryablehttp"
 	json "github.com/json-iterator/go"
 
 	"github.com/infracost/infracost/internal/config"
+	"github.com/infracost/infracost/internal/logging"
 	"github.com/infracost/infracost/internal/schema"
 )
 
@@ -19,11 +21,14 @@ type PolicyV2APIClient struct {
 
 // NewPolicyV2APIClient retrieves resource allow-list info from Infracost Cloud and returns a new policy client
 func NewPolicyV2APIClient(ctx *config.RunContext) (*PolicyV2APIClient, error) {
+	client := retryablehttp.NewClient()
+	client.Logger = &LeveledLogger{Logger: logging.Logger.WithField("library", "retryablehttp")}
 	c := PolicyV2APIClient{
 		APIClient: APIClient{
-			endpoint: ctx.Config.TagPolicyAPIEndpoint, // use tag policy api endpoint for now.
-			apiKey:   ctx.Config.APIKey,
-			uuid:     ctx.UUID(),
+			httpClient: client.StandardClient(),
+			endpoint:   ctx.Config.TagPolicyAPIEndpoint, // use tag policy api endpoint for now.
+			apiKey:     ctx.Config.APIKey,
+			uuid:       ctx.UUID(),
 		},
 	}
 
--- a/internal/apiclient/policyv2.go
+++ b/internal/apiclient/policyv2.go
@@ -1,0 +1,294 @@
+package apiclient
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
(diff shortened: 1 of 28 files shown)


internal/apiclient/policyv2.go
```diff
@@ -0,0 +1,294 @@
+package apiclient
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
```
hugorut, line 6: can we use jsoniter instead of the base encoding/json package - it is faster.

internal/apiclient/policyv2.go
```diff
@@ -0,0 +1,294 @@
...
+	"fmt"
+	"github.com/infracost/infracost/internal/config"
+	"github.com/infracost/infracost/internal/schema"
+	"sort"
+)
```
hugorut, line 11: nitpick and feel free to ignore - but i like to sort the imports as follows

1. standard pkg
2. 3rd prty pkg
3. current application pkg

You can automate this in goland with the following:

SCR-20230914-b76

internal/apiclient/policyv2.go
```diff
@@ -0,0 +1,294 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to upload filtered past partial resources %w", err)
+		}
+		project.Metadata.PastPolicySha = sha
+	}
+
+	return nil
+}
+
+func (c *PolicyV2APIClient) uploadProjectPolicyData(p2rs []policy2Resource) (string, error) {
```
hugorut, line 65: do we want to add some retry logic to this - or is it already handled in the base client?

internal/apiclient/policyv2.go
```diff
@@ -0,0 +1,294 @@
...
+	Key   string          `json:"key"`
+	Value json.RawMessage `json:"value"`
+}
+
+type policy2Reference struct {
+	Key       string   `json:"key"`
+	Addresses []string `json:"addresses"`
+}
+
+type policy2Resource struct {
```
hugorut, line 123: could we change this to?

```go
type policy2Resource struct {
    *schema.ResourceData
    Values       []policy2Value           `json:"values"`
	References   []policy2Reference       `json:"references"`
	Metadata     policy2InfracostMetadata `json:"infracostMetadata"`
}
```

The embedding means we only need to set the information we need with the policies API and won't potentially mis a struct copy.

I'm not sure if this works for this usecase though and it might be more pain then it's worth - so feel free to ignore.

internal/output/output.go
```diff
@@ -97,9 +97,15 @@ func (p Project) ToSchemaProject() *schema.Project {
 		resources = append(convertOutputResources(p.Breakdown.Resources, false), convertOutputResources(p.Breakdown.FreeResources, true)...)
 	}
 
+	var clonedMetadata *schema.ProjectMetadata
+	if p.Metadata != nil {
+		m := *p.Metadata
+		clonedMetadata = &m
+	}
```
hugorut, line 104: what is the importance of cloning the pointer here? Can we add a comment to clarify?

hugorut (approved):

internal/apiclient/policyv2.go
tim775: I think it's better as is because the policy2Resource is coupled with the graphql api more that it is the ResourceData.  If we add a new field to resource data without changing the graphql, the api call will blow up.  since we're stubbing out the api that will only happen once it gets to prod.

internal/apiclient/policyv2.go
```diff
@@ -0,0 +1,294 @@
...
+			apiKey:   ctx.Config.APIKey,
+			uuid:     ctx.UUID(),
+		},
+	}
+
+	prw, err := c.getPolicyResourceAllowlist()
+	if err != nil {
+		return nil, err
+	}
+	c.allowlist = prw
```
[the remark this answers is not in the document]
tim775: I think it's clear as is and I don't like the idea of modifying the client as a side effect in the method that makes an api call.

PR #2665, opened 2023-09-17
feat: add checkPolicies endpoint to Policies V2 API

tim775:
- Adds a new CheckPolicies call to the PoliciesV2API that returns both Tag Policies and Fin Ops Policies
- Adds the FinOpsPolicies to the output
- Modifies comment & upload commands to use the PoliciesV2API instead of TagPoliciesAPI
- Remove TagPoliciesAPI


internal/apiclient/policyv2.go
```diff
@@ -41,6 +42,123 @@ func NewPolicyV2APIClient(ctx *config.RunContext) (*PolicyV2APIClient, error) {
 	return &c, nil
 }
 
+func (c *PolicyV2APIClient) CheckPolicies(ctx *config.RunContext, out output.Root) ([]output.TagPolicy, []output.FinOpsPolicy, error) {
```
hugorut, line 45: Can we change the signature to the following: 

```suggestion
type PolicyOutput struct {
    TagPolicies []output.TagPolicy
    FinnOpsPolicies []output.FinOpsPolicy
}

func (c *PolicyV2APIClient) CheckPolicies(ctx *config.RunContext, out output.Root) (*PolicyOutput, error) 
```

Probably personal preference, but, I prefer limit return args to maximium 2. This is because:

1. It's more verbose and readable for consumers to understand what's being returned from the function.
2. It's easier to add additional return args in the future by just appending to the output `struct`.
3. If we want to add additional logic to the return args (e.g. helper methods for filtering/comparison) it's easier to do with a struct, and easier for consumers to use.
4. If there's a complex/involved return arg we can document this more easily on as a parameter on the return struct.
5. we can return nil on an error case, which limits mem use (not super impactful here but could be if we add additional args in the future, see point 2)

internal/output/combined.go
```diff
@@ -333,6 +339,30 @@ func mergeTagPolicies(tagPolicies []TagPolicy) []TagPolicy {
...
+	}
+
+	fpMerged := make([]FinOpsPolicy, 0, len(fpMap))
+	// use the original tagPolicies array to iterate over the map so the order is preserved
+	for _, tp := range finOpsPolicies {
+		if mergedTp, ok := fpMap[tp.PolicyID]; ok {
+			fpMerged = append(fpMerged, mergedTp)
+			delete(fpMap, tp.PolicyID)
+		}
+	}
```
hugorut, line 361:
```suggestion
	for _, fp := range finOpsPolicies {
		if mergedFp, ok := fpMap[fp.PolicyID]; ok {
			fpMerged = append(fpMerged, mergedFp)
			delete(fpMap, fp.PolicyID)
		}
	}
```

internal/output/output.go
```diff
@@ -511,6 +539,63 @@ func (p PolicyCheckFailures) Error() string {
...
+		}
+	}
+
+	return out.String()
+}
+
+func (r FinOpsPolicyResource) Failures() []string {
+	var f []string
+	for _, i := range r.Issues {
+		f = append(f, fmt.Sprintf("'%s' currently set to '%s': %s", i.Attribute, i.Value, i.Description))
```
hugorut, line 594:
```suggestion
		f = append(f, fmt.Sprintf("%q currently set to %q: %s", i.Attribute, i.Value, i.Description))
```

internal/output/output.go
```diff
@@ -511,6 +539,63 @@ func (p PolicyCheckFailures) Error() string {
...
+
+	return fopc
+}
+
+func (fopc FinOpsPolicyCheck) Error() string {
+	if len(fopc.Failing) == 0 {
+		return ""
+	}
+
+	out := bytes.NewBuffer([]byte("FinOps policy check failed:\n"))
```
hugorut, line 576:
```suggestion
	out := &strings.Builder{}
	out := out.WriteString("FinOps policy check failed:\n")
```

strings.Builder is optimized to work with strings

hugorut (commented):
looks good - minor changes

hugorut:
Note: I've also added a comment pr against this: https://github.com/infracost/infracost/pull/2666

PR #2666, opened 2023-09-19
feat: add finops policy output to comment

hugorut:
* adds support for finops policies using old comment output
* refactors comment error handling so it's a bit easier to use

Note: this was not moved to the API as previously discussed as Ali wants the old style comment output until
phase 3. See: https://infracost.slack.com/archives/C025MSU0US3/p1695068299141909?thread_ts=1695040182.419469&cid=C025MSU0US3


tim775 (approved):
I like the errors refactor.

PR #2678, opened 2023-09-29
enhance: support nested policy allow lists IC-46

tim775:


internal/apiclient/policyv2.go
```diff
@@ -375,6 +363,45 @@ func filterResource(rd *schema.ResourceData, allowedKeys map[string]bool) policy
...
+			}
+		}
+	}
+
+	b, err := jsonSorted.Marshal(values)
+	if err != nil {
+		logging.Logger.Warn("Failed to marshal filtered values")
+	}
+	return b
+}
```
hugorut, line 403:
```suggestion
func filterValues(rd gjson.Result, allowList map[string]gjson.Result) map[string]interface{} {
	values := make(map[string]interface{}, len(allowList))
	for k, v := range rd.Map() {
		if allow, ok := allowList[k]; ok {
			if allow.IsBool() && allow.Bool() {
				values[k] = json.RawMessage(v.Raw)
			} else if allow.IsObject() {
				nestedAllow := allow.Map()
				if v.IsArray() {
					vArray := v.Array()
					nestedVals := make([]interface{}, 0, len(vArray))
					for _, el := range vArray {
						nestedVals = append(nestedVals, filterValues(el, nestedAllow))
					}
					values[k] = nestedVals
				} else {
					values[k] = filterValues(v, nestedAllow)
				}
			} else {
				logging.Logger.WithField("Key", k).WithField("Type", allow.Type.String()).Warn("Unknown allow type")
			}
		}
	}

	return values
}
```

This way we don't need to do all the unecessary json Marshalling, which is quite heavy. This means we can alter `policy2Resource` as well and actually have a datastructure:

```
type policy2Resource struct {
	ResourceType string                   `json:"resourceType"`
	ProviderName string                   `json:"providerName"`
	Address      string                   `json:"address"`
	Tags         *[]policy2Tag            `json:"tags,omitempty"`
	Values       map[string]interface{}   `json:"values"`
	References   []policy2Reference       `json:"references"`
	Metadata     policy2InfracostMetadata `json:"infracostMetadata"`
}
```
Furthermore, sorting the keys here, just to have the tests pass is a bit unecessary. We can move the sorting element to get the tests pass to the `GraphqlTestServerWithWriter` so that we don't affect application logic.

hugorut (commented):
Interesting, I was thinking we'd do it a completely different way. Where the api specifies a gjson lookup key e.g. `ebs_block_device[*].iops`. I think this would be more performant and means we could just rely on the gjson lib. But this approach works as well and has some merits with the modelling of keys. 

Some minor tweaks that we need to make to remove all the json Marshalling steps.

internal/apiclient/policyv2.go
hugorut: additionally the amount of nested `if`s here quite close to my limit. If we can reduce that somehow that would be appreciated.

tim775: The keys need to be sorted because we're calcuating a sha dashboard side to avoid storing duplicates, I'll add a comment to make that clear.
 
 I don't know that it's better, but if we want Values to be a map[string]interface{} for some reason we could sort the whole thing right before upload.  That would look something like:
 ```
 	q := `
	mutation($policyResources: [PolicyResourceInput!]!) {
		storePolicyResources(policyResources: $policyResources) {
			sha
		}
	}
	`
	jsonSorted := json.Config{SortMapKeys: true}.Froze()
	var b json.RawMessage
	b, err := jsonSorted.Marshal(p2rs)
	if err != nil {
		return "", fmt.Errorf("query storePolicyResources failed to marshal policy2Resources %w", err)
	}

	v := map[string]interface{}{
		"policyResources": b,
	}

	results, err := c.doQueries([]GraphQLQuery{{q, v}})
```

hugorut: ah makes sense - we can just keep the RawMessage field. But do the sorting/marshalling once here: https://github.com/infracost/infracost/pull/2678/files#diff-4aca3a3972a8907848b8e932d39b01d2e9f726f2cbef8371e916d239eb0c9d17R318

hugorut: an alternative is to sort the keys when we actually store them - as that's the actual bit of the application that's interested in the sorting. But that's probably one refactor too many

hugorut: yup 👍

PR #2695, opened 2023-10-04
enhance: Add storage account refs for storage management policies IC-21

tim775:


vdmgolub (approved):

internal/providers/terraform/azure/registry.go
```diff
@@ -137,6 +137,7 @@ var ResourceRegistry []*schema.RegistryItem = []*schema.RegistryItem{
 	getApplicationInsightsStandardWebTestRegistryItem(),
 	getRecoveryServicesVaultRegistryItem(),
 	getBackupProtectedVmRegistryItem(),
+	getStorageManagementPolicy(),
```
vdmgolub, line 140: I'm okay dropping `RegisterItem` part, but was it intended? :)

tim775: oops no i'll fix that

PR #2696, opened 2023-10-04
fix: support modules that have `for_each` based on other module outputs

aliscott:


internal/hcl/evaluator.go
```diff
@@ -516,13 +517,27 @@ func (e *Evaluator) expandBlockForEaches(blocks Blocks) Blocks {
...
 
-				if v, ok := e.moduleCalls[clone.FullName()]; ok {
-					v.Definition = clone
+				if clone.Type() == "module" {
+					if v, ok := e.moduleCalls[clone.FullName()]; ok {
+						v.Definition = clone
+					} else {
+						modCall, err := e.loadModule(clone)
+						if err != nil {
+							e.logger.WithError(err).Debugf("could not load module %s", clone.FullName())
```
hugorut, line 526: can we make this log line a bit different from the last so that we know the error is coming from the for each section. e.g. `failed to create expanded module call, could not load module %s`

internal/hcl/evaluator.go
```diff
@@ -516,13 +517,27 @@ func (e *Evaluator) expandBlockForEaches(blocks Blocks) Blocks {
...
+					}
 				}
+
 				expanded = append(expanded, clone)
 
 				return false
 			})
+
+			if block.Type() == "module" {
+				delete(e.moduleCalls, block.FullName())
```
hugorut, line 539: can we add a log line here just to say which module we're deleting from the calls.

hugorut (approved):

PR #2701, opened 2023-10-13
enhance: Add governance checks

tim775:


cmd/infracost/comment.go
```diff
@@ -58,8 +58,8 @@ func commentCmd(ctx *config.RunContext) *cobra.Command {
 		_ = subCmd.Flags().MarkHidden("show-changed")
 		subCmd.Flags().Bool("skip-no-diff", false, "Skip posting comment if there are no resource changes. Only applies to update, hide-and-new, and delete-and-new behaviors")
 		_ = subCmd.Flags().MarkHidden("skip-no-diff")
-		subCmd.Flags().String("guardrail-check-path", "", "Path to Infracost guardrail data (experimental)")
-		_ = subCmd.Flags().MarkHidden("guardrail-check-path")
+		subCmd.Flags().String("additional-path", "", "Path to additional comment text (experimental)")
```
hugorut, line 61:
```suggestion
		subCmd.Flags().String("additional-comment-data-path", "", "Path to additional comment text (experimental)")
```

Might be a bit more readable, but I guess it's not too much of an issue as this is internal

internal/apiclient/dashboard.go
```diff
@@ -107,20 +118,26 @@ func newRunInput(ctx *config.RunContext, out output.Root) (*runInput, error) {
...
 
 	ri, err := newRunInput(ctx, out)
 	if err != nil {
 		return response, err
 	}
 
+	commentFormat := "MARKDOWN_HTML"
+	if commentBasicSyntax {
+		commentFormat = "MARKDOWN"
+	}
```
hugorut, line 132: If we change the `AddRun` input to accept a `CommentFormat` type rather than a bool this could give us:

1. A closer representation of the graphql enum
2. Means that we could more easily expand options in the future

Code would look something like this (untested):

```suggestion
type CommentFormat string

var (
    CommentFormatMarkdownHTML = "MARKDOWN_HTML"
    CommentFormatMarkdown = "MARKDOWN"
)

func (c *DashboardAPIClient) AddRun(ctx *config.RunContext, out output.Root, commentFormat CommentFormat) (AddRunResponse, error) {
	response := AddRunResponse{}

	ri, err := newRunInput(ctx, out)
	if err != nil {
		return response, err
	}
```

Thoughts?

internal/apiclient/dashboard.go
```diff
@@ -164,44 +184,46 @@ func (c *DashboardAPIClient) AddRun(ctx *config.RunContext, out output.Root) (Ad
...
 
-		if response.GuardrailCheck.TotalChecked > 0 {
-			guardrailStr := "guardrail"
-			if response.GuardrailCheck.TotalChecked > 1 {
-				guardrailStr = "guardrails"
+		for _, gr := range response.GovernanceResults {
+			t := strings.ReplaceAll(gr.Type, "_", " ")
+			if gr.Checked > 0 {
+				maybePluralT := t
+				if gr.Checked != 1 {
```
hugorut, line 196:
```suggestion
				if gr.Checked > 1 {
```

hugorut (approved):

cmd/infracost/comment.go
tim775: yeah I was hoping you would have a better idea for the name

internal/apiclient/dashboard.go
tim775: that's fine. what I really would like is a nice way to optionally request the governanceComment.  we only need it for comment and it's potentially expensive for the dashboard to generate.  Maybe i'll fmtF the query string and conditionally set %s = "governanceComment" or ""

PR #2702, opened 2023-10-13
chore(docker): Build ARM64 images

torqu3e:
Updates the Github action for building the docker images. Uses parallel runners to build image for each platform and then push the manifest after the build. Cross arch build here should be a relative non issue due to the infrequency of action run (~once every 2 months).

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -11,6 +11,12 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: 
+          - linux/amd64
+          - linux/arm64
     steps:
       - uses: actions/checkout@v3
       - name: Set up Go 1.x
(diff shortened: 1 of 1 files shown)


aliscott:
Thanks @torqu3e 🙏. It looks like this only updates the `Dockerfile.ci` (`infracost:ci-*`) images. I think we would want this to work on both the `Dockerfile` and the `Dockerfile.ci` images.

Any ideas how long this pipeline will take to run? One of the reasons we have a infrequent release cycle is because of the time for these pipelines to run so I'm hesitant about making it significantly slower.

torqu3e:
@aliscott I don't see any mention of `Dockerfile` by itself in any other workflows. There's the create-test-image one which I've not seen run. Where are you referring to for the other files?

This will likely add some more time though hopefully not humongously so because of the parallel image builds. The majority of the time presently is in the go binary build which I can try to look into separate from this.

aliscott:
@torqu3e the `Dockerfile` build is here: https://github.com/infracost/infracost/blob/acfffacdb4d4d26bcc54f3c58e07c97f0a64dae2/.github/workflows/create-release.yml#L55.

I'll see if I can run this in a test environment to see how much time it adds, but not sure how easy that will be.

PR #2703, opened 2023-10-14
Experimental: graph evaluator

aliscott:


internal/hcl/context.go
```diff
@@ -80,6 +83,8 @@ func (c *Context) SetByDot(val cty.Value, path string) {
 }
 
 func (c *Context) Set(val cty.Value, parts ...string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
aliscott, line 87: Can this be removed?

internal/hcl/graph_vertex_module.go
```diff
@@ -0,0 +1,128 @@
...
+
+		err = g.Populate(moduleEvaluator)
+		if err != nil {
+			return nil, fmt.Errorf("error populating graph: %w", err)
+		}
+	}
+	visitMu.Unlock()
+
+	g.ReduceTransitively()
+	g.Walk()
```
aliscott, line 103: Can we use multiple graphs here?

internal/hcl/graph.go
```diff
@@ -0,0 +1,325 @@
...
+
+		err := g.dag.AddEdge(g.rootVertex.ID(), vertex.ID())
+		if err != nil {
+			g.logger.Debugf("error adding edge: %s", err)
+		}
+
+		for _, ref := range vertex.References() {
+			var srcId string
+
+			parts := strings.Split(ref.Key, ".")
```
aliscott, line 134: Can we extract this somewhere?

internal/hcl/graph.go
```diff
@@ -0,0 +1,325 @@
...
+			if ref.ModuleAddress != vertex.ModuleAddress() {
+				srcId = ref.ModuleAddress
+			}
+
+			// Strip the count/index suffix from the source ID
+			srcId = stripCount(srcId)
+
+			if srcId == vertex.ID() {
+				continue
+			}
```
aliscott, line 166: Can we extract this somewhere?

internal/hcl/graph.go
```diff
@@ -0,0 +1,325 @@
...
+		return
+	}
+
+	ve := vert.Evaluator()
+	if ve != nil && vert.Evaluator() != nil {
+		vert.Evaluator().AddFilteredBlocks(blocks...)
+	}
+}
+
+func referencesForBlock(b *Block) []VertexReference {
```
hugorut, line 244: I'd prefer if we moved this and the corresponding `referencesForAttribute` to methods on the `Block` and `Attribute` types.

internal/hcl/graph.go
```diff
@@ -0,0 +1,325 @@
...
+	"strings"
+	"sync"
+
+	"github.com/heimdalr/dag"
+	"github.com/sirupsen/logrus"
+	"github.com/zclconf/go-cty/cty"
+)
+
+var (
+	visitMu              sync.RWMutex
```
hugorut, line 15: This is a global mutext across all go routines. This means this will lock visits across projects that use completely different `Graph`s, meaning they will never hit concurrent read/write issue. We can change this to be a `Graph` scoped mutex, this should hopefully speed up the performance.

internal/hcl/graph_vertex_data.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (v *VertexData) Evaluator() *Evaluator {
+	return v.evaluator
+}
+
+func (v *VertexData) References() []VertexReference {
+	return referencesForBlock(v.block)
+}
+
+func (v *VertexData) Evaluate() error {
```
hugorut, line 32: do we not need to lock this?

internal/hcl/graph_vertex_module.go
hugorut: Again - I think this will work better than what we do currently as it will be significantly easier to understand.

internal/hcl/parser.go
```diff
@@ -397,9 +397,31 @@ func (p *Parser) ParseDirectory() (m *Module, err error) {
...
+		err = g.Populate(evaluator)
+		if err != nil {
+			return m, err
+		}
+
+		g.ReduceTransitively()
+		g.Walk()
+		evaluator.module.Blocks = evaluator.filteredBlocks
+		evaluator.module = *evaluator.collectModules()
+		root = &evaluator.module
```
hugorut, line 418: I'd prefer if we have a equivalent interface to the current evaluator so it's easier to understand that the `Graph` and `Evaluator` and equivalent in what they are doing. Somthing like:

```suggestion
		root, err := g.Run(evaluator)
		if err != nil {
			return m, err
		}
```

internal/hcl/graph_vertex_data.go
hugorut: ignore this comment - in favour of the master comment on refactoring the interface.

internal/hcl/parser.go
```diff
@@ -397,9 +397,28 @@ func (p *Parser) ParseDirectory() (m *Module, err error) {
 		nil,
 	)
 
-	root, err := evaluator.Run()
-	if err != nil {
-		return m, err
+	var root *Module
+
+	// Graph evaluation
+	if os.Getenv("INFRACOST_GRAPH_EVALUATOR") == "true" {
```
aliscott, line 403: @hugorut should we move all this logic inside `evaluator.Run`?

This would also fix the (not-important) issue that we're not seeing the `⠴ Evaluating Terraform directory` spinner when using the graph evaluator.

internal/hcl/parser.go
```diff
@@ -397,9 +397,28 @@ func (p *Parser) ParseDirectory() (m *Module, err error) {
...
+		if err != nil {
+			return m, err
+		}
+	} else {
+		// Existing evaluation
+		root, err = evaluator.Run()
+		if err != nil {
+			return m, err
+		}
 	}
```
aliscott, line 422: @hugorut should we move all this logic inside evaluator.Run?

This would also fix the (not-important) issue that we're not seeing the ⠴ Evaluating Terraform directory spinner when using the graph evaluator.

hugorut: I seems a bit wierd to have the `Graph` -> which has a dep on evaluator, to be initialised and used within the `Graph`. I think it makes actually more sense to place it here, with the understanding that `Evaluator.Run` will be removed in the future when we migrate fully to the `Graph`. If you think it cleans up some stuff though, I'm not super against it so feel free to move, but add a comment explaining the whys/whats.

hugorut (approved):
🚀

hugorut (approved):

PR #2744, opened 2023-11-14
feat: add autodetect output option to `generate config`

hugorut:
`infracost generate config` now supports creating a config file from the native "auto-detect" functionality. If no `template-path` or `template` is provided, generate will output the auto-detected projects.


tim775 (approved):
Nice this seems kind of huge.  Do we need tests?

hugorut:
> Nice this seems kind of huge. Do we need tests?

Good shout, added some tests + tweaks

PR #2749, opened 2023-11-16
chore: add remote module calls to project metadata

hugorut:
Changes `Evaluator` logic so that when a module is evaluated we set a `SourceURL` on the definition. This is populated from the ModuleMetadata. The url is either built from the Metadata source or a newly added `DownloadUrl` which is found by calling a registry disco to try and resolve the original source.

The new source url is then propagated to the plan JSON and eventually aggregated into a list on the `Project` `Metadata`.


internal/providers/terraform/parser.go
```diff
@@ -133,13 +133,24 @@ func (p *Parser) populateUsageData(resData map[string]*schema.ResourceData, usag
 	}
 }
 
-func (p *Parser) parseJSON(j []byte, usage schema.UsageMap) ([]*schema.PartialResource, []*schema.PartialResource, []schema.ProviderMetadata, error) {
+type ParsedPlanConfiguration struct {
```
tim775, line 136: 👍

tim775 (approved):

PR #2751, opened 2023-11-16
enhance: Add throughput cost component for Opensearch IC-336

vdmgolub:


internal/resources/aws/search_domain.go
```diff
@@ -238,3 +267,27 @@ func (r *SearchDomain) opensearchifyClusterInstanceType(instanceType string) *st
 	s := strings.Replace(instanceType, ".elasticsearch", ".search", 1)
 	return &s
 }
+
+// Calculates billable throughput
+// - 125 MiB/s free for volumes up to 170 GB
+// - +250 MiB/s free for every 3 TB for volumes above 170 GB (storage is rounded up).
+func (r *SearchDomain) calculateBillableThroughput(throughput decimal.Decimal, storage decimal.Decimal) decimal.Decimal {
```
hugorut, line 274: is this a global amount or per domain? If it's global, we've tended to ignore these free quotas in the past

internal/resources/aws/search_domain.go
```diff
@@ -238,3 +267,27 @@ func (r *SearchDomain) opensearchifyClusterInstanceType(instanceType string) *st
 	s := strings.Replace(instanceType, ".elasticsearch", ".search", 1)
 	return &s
 }
+
+// Calculates billable throughput
```
hugorut, line 271:
```suggestion
// calculateBillableThroughput calculates billable throughput
```

hugorut (approved):

internal/resources/aws/search_domain.go
vdmgolub: Per domain I guess. I was following the calculator as it makes the cost estimate more precise.

PR #2753, opened 2023-11-16
enhance: remove "we have been notified..." to better support custom error messages

tim775:


cmd/infracost/run.go
```diff
@@ -469,7 +469,7 @@ func (r *parallelRunner) runProjectConfig(ctx *config.ProjectContext) (*projectO
 			}
 
 			if e, ok := err.(*apiclient.APIError); ok {
-				return nil, fmt.Errorf("%v\n%s", e.Error(), "We have been notified of this issue.")
+				return nil, fmt.Errorf("%v", e.Error())
```
hugorut, line 472: I believe we don't need to wrap the error if we're just using the `Error()` function. This is basically exactly the same as writing below:

```suggestion
				return nil, e
```

hugorut (approved):

PR #2755, opened 2023-11-18
feat: improve auto-detection support

hugorut:
Changes `ProjectLocator` behaviour to associated nested tfvar files with detected projects. This means projects specified with tfvars files like such:

```
.
└── infra/
    ├── app1/
    │   ├── main.tf
    │   └── envs/
    │       ├── prod.tfvars
    │       └── dev.tfvars
    └── app2/
        └── main.tf
```

Will have proper auto-detection support.

In order to accomodate this funcitonality I've change the `ProjectLocator` so that if walks all the directories, building a list of detected projects and terraform var files. Var files are then associated with a given project if:

1. they exist in the same directory
2. they exist in a child directory (max 2 levels deep) and are not already associated with a project.

I've also change the `TerragruntHCLProvider` behaviour to skip the autodetect logic. We should never need to auto-detect project that have been provided by a Terragrunt output.


internal/hcl/parser.go
```diff
@@ -234,15 +243,33 @@ func LoadParsers(ctx *config.ProjectContext, initialPath string, loader *modules
 
 			var varFiles []string
 			var autoVarFiles []string
+
+			// first remove all the "auto" tfvar files from the list of discovered var files.
+			// These files should not constitute a new "project" as they won't define an
+			// environment but defaults that should be applied across all environments.
 			for _, varFile := range rootPath.TerraformVarFiles {
-				if strings.HasSuffix(strings.TrimSuffix(varFile, ".json"), ".auto.tfvars") {
+				withoutJsonSuffix := strings.TrimSuffix(varFile, ".json")
```
aliscott, line 251:
```suggestion
				withoutJSONSuffix := strings.TrimSuffix(varFile, ".json")
```

internal/hcl/parser.go
```diff
@@ -234,15 +243,33 @@ func LoadParsers(ctx *config.ProjectContext, initialPath string, loader *modules
...
+
+				if strings.HasSuffix(withoutJsonSuffix, ".auto.tfvars") {
+					autoVarFiles = append(autoVarFiles, varFile)
+					continue
+				}
+
+				if strings.HasSuffix(withoutJsonSuffix, "defaults.tfvars") {
```
aliscott, line 258: Should we maintain a list of these? Others I've seen are:
* shared
* global

Although we'd want to exclude the case where they are prefixed with something like `prod-` or `dev-`

aliscott (approved):
This is awesome!

internal/hcl/parser.go
hugorut: Yeah probably a good shout, I'll add a list with the ones you've suggested and check they don't have env prefixs

aliscott (approved):

PR #2756, opened 2023-11-20
feat: add quota exceeded output

hugorut:
Changes infracost outputs to show quota exceeded for markdown and html outputs.


cmd/infracost/run.go
```diff
@@ -468,6 +469,15 @@ func (r *parallelRunner) runProjectConfig(ctx *config.ProjectContext) (*projectO
 				)
 			}
 
+			var apiErr *apiclient.APIError
+			if errors.As(err, &apiErr) {
+				if apiErr.Code != http.StatusForbidden {
```
tim775, line 474: We also return 403/forbidden if they have a bad api key so we need to parse the message or (or maybe we can add additional context to the response) to decide if it is a quota issue.

internal/output/templates/html.tmpl
```diff
@@ -256,6 +256,9 @@ iVBORw0KGgoAAAANSUhEUgAAAMAAAADACAMAAABlApw1AAAABGdBTUEAALGPC/xhBQAAAAFzUkdCAK7O
   </head>
 
   <body>
+    {{- if .RunQuotaExceeded }}
+      <p>Infracost free monthly run limit exceeded. Please contact <a href="mailto:hello@infracost.io">hello@infracost.io</a> to purchase usage pack. </p>
```
tim775, line 260: I think we probably don't want to hardcode this message in the CLI if we can avoid it?

tim775 (changes requested):

internal/output/templates/html.tmpl
hugorut: so where will the messge come from? The `error` field returned by the pricing api?

cmd/infracost/run.go
hugorut: I was originally going to parse the message but thought that would be pretty flaky, as if we change the messaging it will break. I'll do it for now, but in the future i think we need to add a `code` or equivalent field to the hcpapi json response.

hugorut:
@tim775 just handling the api_key change, hold off review for now

tim775 (approved):

PR #2757, opened 2023-11-20
Fix graph cycle errors

aliscott:


internal/hcl/graph.go
```diff
@@ -226,20 +227,31 @@ func (g *Graph) Populate(evaluator *Evaluator) error {
...
-			srcId = stripCount(srcId)
+			dstID := id
 
-			if srcId == id {
-				continue
+			// If the vertex is a module call we should set the edge
+			// to the variable block within the module
+			if _, ok := vertex.(*VertexModuleCall); ok {
+				if ref.AttributeName != "" {
+					dstID = fmt.Sprintf("%s.variable.%s", stripModuleCallPrefix(id), ref.AttributeName)
```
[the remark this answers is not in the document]
aliscott: We should exclude attributes `for_each`, `count` and `source`.

internal/hcl/graph_vertex_module_call.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
+var moduleCallArgs = []string{"source", "version", "for_each", "count", "providers", "depends_on", "lifecycle"}
```
hugorut, line 13: can we change this to a map lookup so we don't have to loop over the list to check if the attr is a moduleCallArg

internal/hcl/graph_vertex_variable.go
```diff
@@ -42,11 +43,17 @@ func (v *VertexVariable) Visit(mutex *sync.Mutex) error {
...
+		// Re-evaluate the matching module input variables for this variable block
+		// to ensure we have the most up-to-date value
 		inputVars := e.inputVars
+
 		if moduleInstance.moduleCall != nil {
-			inputVars = moduleInstance.moduleCall.Definition.values().AsValueMap()
+			attr, ok := moduleInstance.moduleCall.Definition.AttributesAsMap()[v.block.LocalName()]
+			if ok {
+				inputVars = map[string]cty.Value{
+					v.block.LocalName(): attr.value(0),
```
hugorut, line 54: `attr.Value()` should work here

hugorut (approved):

internal/hcl/graph_vertex_module_call.go
aliscott: Is your reasoning for code readability or performance? I find this more readable than using a `map[string]struct{}` type, and think any performance improvements will be negligible (if at all) for small slices.

hugorut: reasoning is for performance, but I also find map lookup more concise, eg:

```
func attrIsVarInput(name string) bool {
     _, exists := moduleCallArgs[name]
     return exists
}
```

vs 

```
func attrIsVarInput(name string) bool {
	for _, arg := range moduleCallArgs {
		if name == arg {
			return false
		}
	}

	return true
}
``

don't have a strong opinion about it so feel free to keep

PR #2761, opened 2023-11-23
fix: only filter dynamic blocks when expanding

hugorut:
resolves issue that more than one dynamic block for the same block caused the expanded blocks to be overwritten. Now we explicitly build the childBlocks when expanding the dynamic blocks, rather than removing the blocks when expanding.


internal/hcl/evaluator.go
```diff
@@ -406,33 +406,78 @@ func (e *Evaluator) expandBlocks(blocks Blocks, lastContext hcl.EvalContext) Blo
...
+
+			for attrName, attr := range content.AttributesAsMap() {
+				b.context.Root().SetByDot(attr.Value(), fmt.Sprintf("%s.%s.%s", b.Reference().String(), blockName, attrName))
 			}
+
+			newChildBlocks = append(newChildBlocks, content)
 		}
 	}
+
+	return &Block{
```
hugorut, line 461: return a new pointer to a block here as the old `Evaluator` logic "reloads" modules using a list of original `RawBlocks`. If we modify the original block passed into this method (which is a pointer) it alters both the `Blocks` and `RawBlocks` lists and causes issues when we run the module again as the child `Block` doesn't have the correct context.

PR #2762, opened 2023-11-27
fix: don't show a double output when uploading to Infracost Cloud

aliscott:
Before we were showing:
```
1 tag policy checked
48 finops policies checked
Estimate uploaded to organization 'org1' in Infracost Cloud
1 tag policy checked
48 finops policies checked
```

Now we show:
```
Estimate uploaded to organization 'org1' in Infracost Cloud
1 tag policy checked
48 finops policies checked
```

It doesn't seem like we need to show the message when we call CheckPolicies anywhere from what I can tell


aliscott:
Closing in favour of https://github.com/infracost/infracost/pull/2764

PR #2781, opened 2023-12-09
enhance: detect projects that use terraform backend blocks as root pr…

aliscott:
…ojects

### DIFF
--- a/internal/hcl/project_locator.go
+++ b/internal/hcl/project_locator.go
@@ -232,18 +232,28 @@ func (p *ProjectLocator) walkPaths(fullPath string, level int) []string {
 	}
 
 	files := hclParser.Files()
-	var providerBlocks bool
+	var hasProviderBlock bool
+	var hasTerraformBackendBlock bool
 
 	for _, file := range files {
-		body, content, diags := file.Body.PartialContent(justProviderBlocks)
+		body, content, diags := file.Body.PartialContent(terraformAndProviderBlocks)
 		if diags != nil && diags.HasErrors() {
 			p.logger.Warn().Err(diags).Msgf("skipping building module information for file %s as failed to get partial body contents", file)
 			continue
 		}
 
-		// only do this after looping through all files
-		if len(body.Blocks) > 0 {
-			providerBlocks = true
+		providerBlocks := body.Blocks.OfType("provider")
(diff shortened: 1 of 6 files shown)


internal/hcl/project_locator.go
```diff
@@ -232,18 +232,27 @@ func (p *ProjectLocator) walkPaths(fullPath string, level int) []string {
...
+		if len(providerBlocks) > 0 {
+			hasProviderBlock = true
+		}
+
+		terraformBlocks := body.Blocks.OfType("terraform")
+		for _, block := range terraformBlocks {
+			backend, _, _ := block.Body.PartialContent(nestedBackendBlock)
+			if len(backend.Blocks) > 0 {
+				hasBackendAttr = true
+			}
```
hugorut, line 255: I believe this will work:

```suggestion
			if block.Type == "backend" {
				hasBackendAttr = true
				break
			}
```

And this means we don't need to decode the nested block, which is a somewhat heavy process

hugorut (commented):
This looks good, minor change/query regarding the block decoding. If we can avoid doing the decoding that would be ideal.

Also the functionality for this detection has been modified in https://github.com/infracost/infracost/pull/2755 so we'll probably get some conflicts.

internal/hcl/project_locator.go
hugorut: scratch that - this won't work as we need the nested body.

hugorut (approved):

internal/hcl/project_locator.go
```diff
@@ -232,18 +232,27 @@ func (p *ProjectLocator) walkPaths(fullPath string, level int) []string {
...
+		providerBlocks := body.Blocks.OfType("provider")
+		if len(providerBlocks) > 0 {
+			hasProviderBlock = true
+		}
+
+		terraformBlocks := body.Blocks.OfType("terraform")
+		for _, block := range terraformBlocks {
+			backend, _, _ := block.Body.PartialContent(nestedBackendBlock)
+			if len(backend.Blocks) > 0 {
+				hasBackendAttr = true
```
hugorut, line 254: can be make sure we `break` here after finding a block so we don't need to do it more than required

PR #2783, opened 2023-12-10
feat: add HCL graph tests

hugorut:
Adds coverage for HCL graph evaluator. Also removes the onlyRunCLI option for the test suite as this is not used.


aliscott:
@hugorut can you give this a review please?

hugorut (commented):
looks good, nice

aliscott (approved):

PR #2785, opened 2023-12-11
Add discovered projects

aliscott:


hugorut (approved):
Nice, like the move to the composability for the `YAML` method that's come out of this

PR #2789, opened 2023-12-13
refactor: force https module get logic

hugorut:
Changes Terraform and Terragrunt getter logic to force a https `Get` before attempting a ssh `Get`. The Terraform logic now uses a custom `GitGetter` which transforms the URL before `Get`. The Terragrunt provider cannot use this as the `Getters` are hidden below in the lib, so we try and transform the `sourceUrl` before Download.

### DIFF
--- a/internal/hcl/modules/fetch.go
+++ b/internal/hcl/modules/fetch.go
@@ -2,6 +2,7 @@ package modules
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -10,6 +11,8 @@ import (
 	getter "github.com/hashicorp/go-getter"
 	"github.com/otiai10/copy"
 	"github.com/rs/zerolog"
+
+	"github.com/infracost/infracost/internal/logging"
 )
 
 // PackageFetcher downloads modules from a remote source to the given destination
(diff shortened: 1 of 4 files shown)


internal/hcl/modules/fetch.go
```diff
@@ -95,3 +103,78 @@ func (r *PackageFetcher) fetch(moduleAddr string, dest string) error {
...
+}
+
+// TransformSSHToHttps transforms a Terraform module source url to an HTTPS
+// equivalent. This only handles source urls prefixed with ssh:: or git::ssh. The
+// shorthand ssh source referenced here:
+// https://developer.hashicorp.com/terraform/language/modules/sources#github e.g.
+// "git@github.com:hashicorp/example.git" in not handled by this method as we
+// expect the source to already be Detected to the valid longhand equivalent
+// before calling this function. This can be achieved by calling
+// getter.Detect(src) before calling TransformSSHToHttps.
```
aliscott, line 160: Just to check:
the `getter.Detect` function handles scp-style addresses like `git::git@example.com/foo/bar` as well?
Also do we handle `git::https://example.com/foo/bar` style addresses?

aliscott (approved):

internal/hcl/modules/fetch.go
hugorut: `git::https://example.com/foo/bar` -> yes we handle this. 
`git::git@example.com/foo/bar` -> it seems that this goes to a file loader instead. I'll see if I can make a change to support this

hugorut: sorry this was my confusion - yes the `Dectect` functionality correctly identifies a scp-style address, e.g: as a ssh one: `git@github.com:hashicorp/example.git` -> `git::ssh://git@github.com/hashicorp/example.git` which is expected

PR #2811, opened 2024-01-05
fix: add support for provider inheritence

tim775:
@hugorut It seems like we need to inherit the providers into parent when creating the hcl.EvalContext for each module.  Let me know what you think.


internal/providers/terraform/hcl_provider.go
```diff
@@ -573,12 +573,17 @@ func (p *HCLProvider) getResourceOutput(block *hcl.Block) ResourceOutput {
...
+		alias := hcl.GetFromProvider(block, block.Provider(), "alias")
+		if alias.Type().Equals(cty.String) && alias.AsString() != "" {
+			providerConfigKey += "." + alias.AsString()
+		}
 		configuration = ResourceData{
 			Address:           stripCountOrForEach(block.LocalName()),
 			Mode:              "managed",
 			Type:              block.TypeLabel(),
 			Name:              stripCountOrForEach(block.NameLabel()),
-			ProviderConfigKey: block.ModuleName() + ":" + block.Provider(),
```
hugorut, line 581: I'm trying to find in Terraform docs what's the significance of the `:` separator for these modules. It seems that when there's a required terraform block in a sub module it prepends the module name, e.g. `module_1:aws` or `module_1:auth0`. I don't think it's a major issue, but can you put a comment above this line just to mark that we looked at this (in case it causes a bug in the future)

cmd/infracost/breakdown_test.go
```diff
@@ -85,6 +86,28 @@ func TestBreakdownFormatJsonWithTags(t *testing.T) {
 	)
 }
 
+func TestBreakdownFormatJsonWithTagsAliasedProvider(t *testing.T) {
+	testName := testutil.CalcGoldenFileTestdataDirName()
```
hugorut, line 90: Let's add sub tests here for the graph evaluator with 

```go
	t.Run("graph evaluator", func(tt *testing.T) {
		tt.Setenv("INFRACOST_GRAPH_EVALUATOR", "true")
		...
	})
```

hugorut (changes requested):
Looks decent, but there's a few things to fix/change before we merge:

1. The new provider functionality hasn't been updated in the graph evaluator, see internal/hcl/graph_vertex_module_call.go. Let's update the logic there and also add a sub test in the `breakdown_test.go` to make sure we cover the graph evaluator. See my comment for more info.
2. Some parser_tests are now failing because we removed the module prefix for the `provider_config_key`. Let's fix those up quickly.
3. We should run this in prod with a debug pod on a few critical repos to make sure that we haven't broken and tag/provider functionality.

internal/providers/terraform/hcl_provider.go
tim775: i'll add the prefix back in, i removed it by mistake.

internal/hcl/evaluator.go
```diff
@@ -110,11 +110,14 @@ func NewEvaluator(
...
 ) *Evaluator {
-	ctx := NewContext(&hcl.EvalContext{
-		Functions: ExpFunctions(module.RootPath, logger),
-	}, parentContext, logger)
+	evalCtx := &hcl.EvalContext{}
+	if providerContext != nil {
+		evalCtx = providerContext.Inner().NewChild()
+	}
+	evalCtx.Functions = ExpFunctions(module.RootPath, logger)
+	ctx := NewContext(evalCtx, providerContext, logger)
```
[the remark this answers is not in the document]
tim775: yeah that's better

tim775:
@hugorut this is ready for re-review and includes fixes for module-defined providers (those weren't really working before).  Sadly, it doesn't work with graph evaluator and probably breaks it in unpleasant ways.  The graph vertex tries to use evaluators from moduleConfig that were created during graph.Populate, but those evaluators don't have the providers set correctly in the context sometimes it panics or gets the wrong provider.

hugorut (changes requested):
These changes look good, just a minor nitpick about the `cty` library. However, I believe the incompatability with the graph evaluator (especially with your mention of the unhandled panics) means we can't merge this. The graph evaluator is currently being used in production so any degredation to this, especially with the tags feature, is a blocker. I'm requesting changes for the time being until we can resolve this or until we discuss this internally and work out what's the best way to move forward with this PR. I believe we can probably find a easy solution to the graph evaluator problem, so I don't think it's a major problem.

aliscott (approved):

PR #2818, opened 2024-01-08
Fix graph evaluator panics

aliscott:


aliscott:
Test failures look unrelated.

internal/hcl/block.go
```diff
@@ -1132,6 +1132,10 @@ func (b *Block) FullName() string {
 }
 
 func (b *Block) Type() string {
+	if b == nil || b.HCLBlock == nil {
+		return ""
+	}
+
```
hugorut, line 1138: hmm this means that somewhere else in the code we've introduced logic which creates a nil `Block` which is something that shouldn't happen. I'm ok with this fix as it's very hard to track down where this happens, but we should keep this in mind for any errors/price inaccuracies we see later with the graph ev.

hugorut (approved):
This looks good, however a small nitpick, in future could we please add some contextual information to the PR comment/description for:

1. what the problem was
2. why this fix

Otherwise it's very tricky for me to review these.

🙏

internal/hcl/block.go
aliscott: Yep, I thought this was odd too, but we seem to have this fix in a lot of places 😕

PR #2820, opened 2024-01-09
feat: add autodetect configuration to config file IC-604

hugorut:
Adds new autodetect section to configuration file which passes configuration down to the `ProjectLocator`. This adds the ability to:

* add projects which haven't been detected with - included_dirs
* exclude bad projects from the autodetection with - excluded_dirs
* change the naming conventions for the env var files with - env_names

The `generate` functionality has had to be changed to accomodate this new functionality. We now need to "partially" unmarshal the template before running autodetect. This is not supported natively by the goyml package so we need to seek the file to build a partial YAML that can be marshalled without errors.

### DIFF
--- a/cmd/infracost/generate.go
+++ b/cmd/infracost/generate.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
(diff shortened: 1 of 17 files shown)


hugorut:
test failures look unrelated

aliscott (approved):
LGTM

PR #2823, opened 2024-01-10
refactor: use different mocking logic for graph evaluator IC-608

hugorut:
This PR changes the mocking logic in the graph evaluator to an expression based approach. Now instead of changing the global Context we instead traverse any expression that is found with a Diagnostic. We walk this expression tree until we hit an expression that is a candidate for mocking. These are:

* hclsyntax.AnonSymbolExpr:
* hclsyntax.LiteralValueExpr
* hclsyntax.ScopeTraversalExpr:

and occasionally hclsyntax.FunctionCallExpr

These are Expressions that likely correspond to "incomplete" values that infracost cannot build. E.g. hclsyntax.ScopeTraversalExpr directly relates to a "missing attribute" or "invalid attribute" diagnostics that are 95% of the cases that we handle. These come about through lack of complete information because attributes/blocks which are populated from the cloud, e.g. data blocks.

Whilst this method works for the normal evaluator, i've only enabled it for the graph evaluator at the moment as this new functionality creates occasional issues with count/for_each attributes that rely on values not already populated in the context - e.g. for_each attributes that rely on module outputs.

### DIFF
--- a/internal/hcl/evaluator.go
+++ b/internal/hcl/evaluator.go
@@ -762,11 +762,19 @@ func (e *Evaluator) evaluateVariable(b *Block, inputVars map[string]cty.Value) (
 
 	attrType := attributes["type"]
 	if override, exists := inputVars[b.Label()]; exists {
-		return e.convertType(b, override, attrType)
+		val, err := e.convertType(b, override, attrType)
+		if err == nil {
+			return val, nil
+		}
+
+		return override, nil
 	}
 
 	if def, exists := attributes["default"]; exists {
-		return e.convertType(b, def.Value(), attrType)
+		val, err := e.convertType(b, def.Value(), attrType)
+		if err == nil {
+			return val, nil
+		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/hcl/evaluator.go
```diff
@@ -762,11 +762,19 @@ func (e *Evaluator) evaluateVariable(b *Block, inputVars map[string]cty.Value) (
 
 	attrType := attributes["type"]
 	if override, exists := inputVars[b.Label()]; exists {
-		return e.convertType(b, override, attrType)
+		val, err := e.convertType(b, override, attrType)
+		if err == nil {
+			return val, nil
+		}
+
+		return override, nil
```
hugorut, line 770: when converting a variable input to the "type" definition, failures to convert results in "nil" values returned. Thus it's safer/more complete to default to returning back the non converted variable input with failures.

aliscott (approved):
Looks like a smart solution 🙌

PR #2831, opened 2024-01-14
Cost component fixes

aliscott:
Fixes various cost component lookups and golden tests based on new prices.

The only important change is for Azure Security Center since some of usage attributes had to be updated to map to the new pricing structure Azure has.

### DIFF
--- a/internal/providers/terraform/azure/testdata/security_center_subscription_pricing_test/security_center_subscription_pricing_test.golden
+++ b/internal/providers/terraform/azure/testdata/security_center_subscription_pricing_test/security_center_subscription_pricing_test.golden
@@ -9,7 +9,7 @@
  └─ Defender for app service                                                                                Monthly cost depends on usage: $14.60 per node             
                                                                                                                                                                        
  azurerm_security_center_subscription_pricing.standard_example["Arm"]                                                                                                  
- └─ Defender for ARM                                                                                        Monthly cost depends on usage: $4.00 per 1M API calls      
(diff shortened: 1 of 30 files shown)


internal/providers/terraform/azure/testdata/security_center_subscription_pricing_test/security_center_subscription_pricing_test.golden
```diff
@@ -6,10 +6,10 @@
  └─ Defender for servers, plan 2                                                                            Monthly cost depends on usage: $14.60 per server           
                                                                                                                                                                        
  azurerm_security_center_subscription_pricing.standard_example["AppServices"]                                                                                          
- └─ Defender for app service                                                                                Monthly cost depends on usage: $14.60 per node             
+ └─ Defender for app service                                                                                Monthly cost depends on usage: $14.60 per nodes
```
vdmgolub, line 9: Can we keep the units singular? Otherwise `per Xs` (for other units too) doesn't read right.

vdmgolub (approved):

internal/providers/terraform/azure/testdata/security_center_subscription_pricing_test/security_center_subscription_pricing_test.golden
aliscott: Can change it back, but it's inconsistent with what we do with other units.

PR #2837, opened 2024-01-15
fix: aft module providers do not evaluate

tim775:
Provider inhertience doesn't work with the `github.com/aws-ia/terraform-aws-control_tower_account_factory` module.


aliscott (approved):
Looks good. I added https://github.com/infracost/infracost/pull/2897 as a separate PR on top of this one if you could take a look.

PR #2840, opened 2024-01-24
enhance: support recognizing env names from var file suffixes

aliscott:
Currently we support grouping var files like `prd.tfvars` with `prd-defaults.tfvars`. This adds similar support to also group `prd.tfvars` with `defaults-prd.tfvars`.

Also for the case where we have vars called `redis-prd.tfvars` and `redis-dev.tfvars`, we still want to add them to the project if there's no other matching `prd.tfvars` or `dev.tfvars` file.

This also includes a fix to update the module loader tests since it looks like TFC registry has switched to using the commit ref instead of tag.


internal/hcl/project_locator.go
```diff
@@ -16,19 +16,20 @@ import (
 )
 
 var (
-	defaultEnvVarNames = regexp.MustCompile(`^(prd|prod|production|preprod|staging|stage|stg|development|dev|release|testing|test|tst|qa|uat|live|sandbox|demo|integration|int|experimental|experiments|trial|validation|perf|sec|dr)`)
+	defaultEnvVarNames = regexp.MustCompile(`^(.*-)?(prd|prod|production|preprod|staging|stage|stg|development|dev|release|testing|test|tst|qa|uat|live|sandbox|demo|integration|int|experimental|experiments|trial|validation|perf|sec|dr)`)
```
aliscott, line 19: Should we also add `(.*-)$` at the end of this to make sure that the env is a separate token.

As well as `-` we might also want to support `_` as a separator.

hugorut (approved):

aliscott:
Test failures look unrelated.

PR #2841, opened 2024-01-24
feat: add terragrunt auto-detect support

hugorut:
Changes the ProjectLocator to support Terragrunt projects. For the time being I have not change the `TerragruntHCLProvider` to take into account that it has a 1-1 link with a Terragrunt project (e.g. no need to find stacks in the Provider). I didn't want to change to many things as this is quite a dramatic change to how we do things. I will follow up later with a refactor to remove unecessary code.

### DIFF
--- a/internal/hcl/project_locator.go
+++ b/internal/hcl/project_locator.go
@@ -1,6 +1,7 @@
 package hcl
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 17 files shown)


internal/hcl/project_locator.go
```diff
@@ -768,6 +799,12 @@ func (p *ProjectLocator) walkPaths(fullPath string, level int) {
 
 		var parseFunc func(filename string) (*hcl.File, hcl.Diagnostics)
 		name := info.Name()
+		if name == "terragrunt.hcl" || name == "terragrunt.hcl.json" {
```
aliscott, line 802: We used to check for the `TERRAGRUNT_CONFIG` value as well but I think that might actually be incorrect behaviour if we want this act like run-all. Not sure if this is something you looked at?

hugorut: yeah we accomodate that for the autodetect but not for this ignore, I'll change

PR #2845, opened 2024-01-29
feat: add auto-detect cousin support

hugorut:
Adds support for var files found in cousin directories, or put another way any var files found in sub directories of aunts. I have also added additional env names to the global list, to improve env var file detection.


aliscott (approved):
LGTM

PR #2868, opened 2024-02-07
feat: add fallback to include all paths functionality

hugorut:
Changes autodetect functionality so that if no paths are detected from an initial provider/backend block project detection we fallback to include all paths. This is particularly usefull for module repositories which often do not contain provider/backend blocks.


aliscott (approved):
Nice!

PR #2871, opened 2024-02-08
refactor: group diagnostic codes

hugorut:
This change groups the diagnostic codes projects into corresponding codes. This allows us to abstract a diagnostic into a higher order grouping of errors that is easier to display to the user.

With this change I've removed the `Warning` construct and instead the only warning "missing vars" into diagnostics. Adding the corresponding fields `IsError` and `FriendlyMessage` to accomodate this merge. This means we now only have one place to define diagnostics and codes/numbers do not collide.

Note: this change depends on various data migrations and changes we need to make in Infracost cloud before it is safe to release.


internal/schema/project.go
```diff
@@ -30,6 +38,125 @@ type ProjectDiag struct {
...
+			"source":       sourceType,
+			"moduleSource": source,
+		},
+		err,
+	)
+}
+
+func newDiag(code int, message string, isError bool, data interface{}, err error) *ProjectDiag {
+	// if the error is already a ProjectDiag, return it rather than creating a new
+	// one. This is to avoids collision of diagnostics.
```
vdmgolub, line 123:
```suggestion
	// one. This is to avoid collision of diagnostics.
```

cmd/infracost/testdata/diff_with_compare_to_with_current_and_past_project_error/diff_with_compare_to_with_current_and_past_project_error.golden
```diff
@@ -21,14 +21,16 @@
...
           {
-            "code": 2,
+            "code": 102,
             "message": "Error loading Terraform modules: failed to parse file testdata/diff_with_compare_to_with_current_and_past_project_error/dev/mod_with_error.tf diag: testdata/diff_with_compare_to_with_current_and_past_project_error/dev/mod_with_error.tf:1,9-10: Invalid block definition; Either a quoted string block label or an opening brace (\"{\") is expected here., and 1 other diagnostic(s)",
-            "data": null
+            "data": null,
+            "isError": true
           },
           {
             "code": 2,
```
vdmgolub, line 30: As we've replaced the codes, this should also be `102` right?
```suggestion
            "code": 102,
```

cmd/infracost/testdata/diff_with_compare_to_with_past_project_error/diff_with_compare_to_with_past_project_error.golden
```diff
@@ -23,7 +23,8 @@
           {
             "code": 2,
```
vdmgolub, line 24:
```suggestion
            "code": 102,
```

vdmgolub (approved):

PR #2875, opened 2024-02-09
fix: specify go toolchain for `Update Fig autocomplete spec` job

chenrui333:
- fix: specify go toolchain for `Update Fig autocomplete spec` job
- also improve the setup-go usage

image

- relates to https://github.com/Homebrew/homebrew-core/pull/162174


aliscott (approved):
Thanks @chenrui333!

PR #2881, opened 2024-02-15
fix: use `GetOutputs` option to fetch nested depdendencies

hugorut:
Requires: https://github.com/infracost/terragrunt/pull/13

Refactors the `TerragruntHCLProvider` to include a `terragruntPathToValue` which is passed to the `TerragruntOptions`. This is used by the Terragrunt library to fetch certain nested deps. See https://github.com/infracost/terragrunt/pull/13 PR description for more details.


aliscott (approved):
This is nice!

PR #2886, opened 2024-02-16
fix: wrong comment on usage example file for aws_rds_cluster

### LINKED ISSUE #2885: Incorrect Comments in infracost-usage.yml Template for RDS Cluster Read/Write Metrics
Hello,

I've encountered an issue with the generated `infracost-usage.yml` file when executing the command `infracost breakdown --path . --sync-usage-file --usage-file infracost-usage.yml`. Specifically, the comments for the `aws_rds_cluster.postgresql.read_requests_per_sec` and `aws_rds_cluster.postgresql.write_requests_per_sec` parameters appear to be mistakenly swapped. Here is the relevant portion of the generated file:

```yaml
...
    # write_requests_per_sec: 0 # Total number of reads per second for the cluster.
    # read_requests_per_sec: 0 # Total number of writes per second for the cluster.
...
```

As you can see, the comment for `write_requests_per_sec` incorrectly states it's for the "Total number of reads per second" when it should refer to writes, and similarly, the comment for `read_requests_per_sec` incorrectly states it's for the "Total number of writes per second" when it should refer to reads.

thisjulio:
wrong comment on usage example file for aws_rds_cluster.write_requests_per_sec and aws_rds_cluster.read_requests_per_sec

this resolve #2885

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -326,8 +326,8 @@ resource_usage:
   aws_rds_cluster.my_cluster:
     capacity_units_per_hr: 50          # Number of aurora capacity units per hour. Only used when engine_mode is "serverless"
     storage_gb: 200                    # Storage amount in GB allocated to the aurora cluster.
-    write_requests_per_sec: 100        # Total number of reads per second for the cluster.
-    read_requests_per_sec: 100         # Total number of writes per second for the cluster.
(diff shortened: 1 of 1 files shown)


aliscott (approved):
Thanks @thisjulio

PR #2901, opened 2024-02-24
Refactor: migration Azure resources

aliscott:
Re-opening @sinabakh's previous work in https://github.com/infracost/infracost/pull/1851, since we want to add new resources that depend on the Azure Virtual Machines cost components. This migrates several Azure resources to the newer provider-agnostic format.

I've fixed these up to add any new cost component lookup fixes, and also fixed panics that were happening with the previous PR.

Outside the scope of this PR:
* Fixing up any stylistic issues in the previous code. The purpose of this is just to migrate the existing resources, which was done semi-automatically.
* Migrate these resources to the new CoreResource type yet, but that should be pretty trivial and can be done separately.


internal/providers/terraform/azure/testdata/image_test/image_test.golden
```diff
@@ -46,7 +46,7 @@
  ├─ storage_data_disk                                                                                                  
  │  └─ Storage (P4, LRS)                                                     1  months                           $5.81 
  └─ storage_data_disk                                                                                                  
-    └─ Storage (P1, LRS)                                                     1  months                           $0.78 
+    └─ Storage (P4, LRS)                                                     1  months                           $5.81
```
aliscott, line 49: From what I can tell this is actually correct since the `disk_size_gb` is set to `null` so should default to 30GB and be a P4 disk.

internal/providers/terraform/azure/frontdoor.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 // getAzureRMFrontdoorRegistryItem returns a registry item for the resource
```
hugorut, line 12:
```suggestion
// getFrontdoorRegistryItem returns a registry item for the resource
```

internal/providers/terraform/azure/frontdoor_firewall_policy.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // getAzureRMFrontdoorFirewallPolicyRegistryItem returns a registry item for the
```
hugorut, line 10:
```suggestion
// getFrontdoorFirewallPolicyRegistryItem returns a registry item for the
```

hugorut (approved):

PR #2909, opened 2024-02-28
fix: ensure expressions that expect bools return bools

aliscott:
If a logical expression (|| or &&) had a sub-expression that was mocked to a non-boolean value, this would always mean the expression always returned false. We now handle this in the same way that we had condition expressions to ensure the mocked expressions return a boolean value if the operator expects boolean values.

### DIFF
--- a/internal/hcl/attribute.go
+++ b/internal/hcl/attribute.go
@@ -528,7 +528,7 @@ func mockExpressionCalls(expr hcl.Expression, diagnostics hcl.Diagnostics, mocke
 		}
 	case *hclsyntax.ConditionalExpr:
 		return &hclsyntax.ConditionalExpr{
-			Condition:   mockCondition(diagnostics, t.Condition, mockedVal),
+			Condition:   mockBoolExpressionCall(t.Condition, diagnostics, mockedVal),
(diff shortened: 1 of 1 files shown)


internal/hcl/attribute.go
```diff
@@ -594,16 +594,51 @@ func mockExpressionCalls(expr hcl.Expression, diagnostics hcl.Diagnostics, mocke
 			MarkerRange: t.MarkerRange,
 		}
 	case *hclsyntax.BinaryOpExpr:
+		// Logical operators (|| and &&) expect bools on both sides, so we need to mock the expression
+		// with a bool value.
+		impl := t.Op.Impl
+		params := impl.Params()
+		lhsParam := params[0]
+		rhsParam := params[1]
```
hugorut, line 602: is this always expected to be a slice of this size?

hugorut: we should make add a size check just in case?

hugorut (commented):
looks good - minor question about the params size

internal/hcl/attribute.go
aliscott: Yeah, all BinaryOp functions have 2 params, but updated this anyway to make sure.

PR #2927, opened 2024-03-05
feat: exit early if Infracost API Key is invalid

hugorut:
Changes the `breakdown`, `diff`, `upload` and `comment` commands to exit early if the INFRACOST_API_KEY is invalid. This is done so that users can better debug their pipelines where an API key might be incorrectly set.

To accomodate this change i've introduced the idea of command middleware which will wrap commands before their execution. This means we can easily chain on additional functionality/checks in the future if required. The specific checkAPIKeyIsValid middleware works by making a empty request to the pricing API service and check if the error returned has an Invalid API Key code. I think this the simplest and most effective way to achieve the API key validation. But you have thoughts of a better way, please let me know.

I have checked with @alikhajeh1 to make sure the error messaging is acceptable.

### DIFF
--- a/cmd/infracost/testdata/breakdown_invalid_apikey/breakdown_invalid_apikey.golden
+++ b/cmd/infracost/testdata/breakdown_invalid_apikey/breakdown_invalid_apikey.golden
@@ -1,0 +1,5 @@
+
+Err:
+Error: Invalid API Key, please check your INFRACOST_API_KEY environment variable or REPLACED_CREDENTIALS_PATH credentials file.
+If you recently regenerated your API key, you can retrieve it from https://dashboard.infracost.io.
+See https://infracost.io/support if you continue having issues.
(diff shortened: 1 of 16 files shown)


cmd/infracost/testdata/breakdown_invalid_apikey/breakdown_invalid_apikey.golden
```diff
@@ -0,0 +1,5 @@
+
+Err:
+Error: Invalid API Key, please check your INFRACOST_API_KEY environment variable or /Users/hugorut/.config/infracost/credentials.yml credentials file.
```
aliscott, line 3: Is this going to fail for people who aren't called hugorut :trollface:?

hugorut: ah yes 🙃

aliscott (approved):

PR #2933, opened 2024-03-07
feat: add support for `RDS` `ExtendedSupport` cost component

hugorut:
Adds cost component to `db_instance` and `rds_cluster_instance` for an extended support cost. This cost is only shown if the instnce engine and version fall into AWS's extended support category, and the current date is within the extended support window.

This is currently only applicable for mysql 5.7, for which extended support started March 2024. However, I have added tests for the rest of the versions listed in AWS docs so that we catch them in future.

The implementation for this feature means we hard code the support window in the resource. This is less than ideal but is the simplest way to accomodate the functionality right now. The pricing API response does contain an `effectiveStartDate` which we could use to filter the prices.

I avoided this, however, as it would involve a decent amount of refactoring and introduce some new concepts on the price fetching end. With this approach we would essentially "filter" the returned prices from the HCPAPI as well as during the lookup. e.g. `effectivePriceDate` is after current date, e.t.c.

I definitely think this is something we should accomodate in the future, however I am hesitant to do it now. Nevertheless, if you think this is a good time to try this, I can have a stab at it.

### DIFF
--- a/cmd/infracost/testdata/breakdown_terraform_use_state_v0_12/breakdown_terraform_use_state_v0_12.golden
+++ b/cmd/infracost/testdata/breakdown_terraform_use_state_v0_12/breakdown_terraform_use_state_v0_12.golden
@@ -1,43 +1,43 @@
 Project: infracost/infracost/cmd/infracost/testdata/terraform_v0.12_state.json
 
- Name                                                              Monthly Qty  Unit   Monthly Cost 
-                                                                                                    
- aws_instance.instance_1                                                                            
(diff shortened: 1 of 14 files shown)


hugorut:
ah I think I've missed the vCPU multiplier of this, @aliscott I'll add that now

aliscott (approved):
> we hard code the support window in the resource

I think this is fine for just now.

If we want to start using effective dates better we should do so as part of a bigger refactor.

PR #2936, opened 2024-03-09
fix: don't evaluate or marshal `depends_on` or `lifecycle` meta attributes

aliscott:
These meta attributes never need to be evaluated since they will never affect the costs/tags.

Additionally, since `depends_on` attributes point to entire resources, we run into problems marshaling them, since the type of the object includes the types of any nested resources referenced by the `depends_on` attribute.

### DIFF
--- a/internal/hcl/block.go
+++ b/internal/hcl/block.go
@@ -816,6 +816,16 @@ func (b *Block) Children() Blocks {
 		return nil
 	}
 
+	var children Blocks
+	for _, child := range b.childBlocks {
+		// Skip lifecycle meta argument blocks since it never needs to be evaluated
+		if supportsLifecycle(child) && child.Type() == "lifecycle" {
+			continue
+		}
+
+		children = append(children, child)
+	}
+
 	return b.childBlocks
 }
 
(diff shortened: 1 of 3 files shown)


internal/hcl/block.go
```diff
@@ -1046,7 +1061,7 @@ func (b *Block) values() cty.Value {
 	if b.Type() == "provider" {
 		for _, child := range b.Children() {
 			key := child.Type()
-			if key == "dynamic" || key == "depends_on" {
```
aliscott, line 1049: I removed this from here since it's not needed and `depends_on` is an attribute not a child block.

internal/hcl/block.go
```diff
@@ -843,6 +853,11 @@ func (b *Block) GetAttributes() []*Attribute {
 
 	hclAttributes := b.getHCLAttributes()
 
+	// Remove the depends_on meta attribute since it never needs to be evaluated
+	if b.Type() == "resource" || b.Type() == "data" || b.Type() == "module" {
+		delete(hclAttributes, "depends_on")
+	}
```
hugorut, line 859: can we move this into the actual `getHCLAttributes` method instead of here, this means we don't have to allocate redundant data to the `hcl.Attributes` map.

hugorut (commented):
minor tweak with the filtering logic

hugorut (approved):

PR #2949, opened 2024-03-15
feat: add support for `env_var_extensions` autodetect configuration

hugorut:
Some projects contain env vars that are have non-standard naming conventions and don't follow the common `.tfvars` and `.tfvars.json` patterns. In this case autodetect fails to pick up the variable files as part of the `ProjectLocator` configuration.

This change adds support for a autodetect configuration to allow us to specify non-standard naming conventions. This, alongside the existing `env_names` configuration field should satifsy most tfvar requirements. 

When specifying `env_var_extensions` we must validate that the file contents are valid HCL, as in many cases custom extensions might have collisions with other non var files. We do this by parsing the file using the hclsyntax library. We only do this validation when extensions are specified so as not to impact performance for other users.

However, we still have the issue of the `tfvars-env.json` usecase which is used by some customers. I've had to make exceptions for this in this PR because I think adding "configurable" support for this might be a bit more complex. In future we might look to add a `env_var_pattern` config, which accepts a regex. This means we could catch the prefix/suffix combo scenario.

### DIFF
--- a/cmd/infracost/testdata/generate/env_var_extensions/expected.golden
+++ b/cmd/infracost/testdata/generate/env_var_extensions/expected.golden
@@ -1,0 +1,40 @@
+version: 0.1
+
+projects:
+  - path: apps/bar
+    name: apps-bar-baz
+    terraform_var_files:
+      - ../default.tfvars
+      - ../baz-custom-ext
+    skip_autodetect: true
+  - path: apps/bar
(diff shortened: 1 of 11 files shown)


hugorut:
seeing some wierd stuff testing this with real data, hold off reviewing

aliscott (approved):
Nice!

PR #2953, opened 2024-03-18
feat: Add usage quantities and footnote to diff and breakdown output

tim775:
Adds several enhancements to better support usage defaults in CLI and comment output.  Individual changes are describe in the commits. There are a large number of golden files changed, they probably only need a cursory review.


hugorut (approved):
looks good, thanks for breaking this down into smaller commits

PR #2957, opened 2024-03-19
enhance: update matching files to env names to be case insensitive

aliscott:
If we have a file called Dev.tfvars in one directory, that should be counted as the same env as dev.tfvars when we do autodetect.

I went with the approach of updating `clean` to lowercase the file name when we do the matching. This means it works across `hasEnvPrefix`, `hasEnvSuffix`, `EnvName`, and `IsEnvName`.

I think the behaviour of `CollectRootPaths` should also be okay with this, since we want to treat the env files as found if they match case insensitively too.


hugorut (approved):
lgtm

PR #2959, opened 2024-03-20
perf: batch usage api queries

tim775:
Improves the performance of usage API queries by batching graphql requests (10 per http request) and issuing the queries concurrently.

### DIFF
--- a/internal/prices/usages.go
+++ b/internal/prices/usages.go
@@ -181,7 +181,7 @@ func FetchUsageData(ctx *config.RunContext, project *schema.Project) (schema.Usa
 		}(jobs, responses)
 	}
 
-	// Feed the workers the jobs of getting prices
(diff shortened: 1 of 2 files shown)


internal/prices/usages.go
```diff
@@ -126,55 +126,113 @@ func popResourceActualCosts(ctx *config.RunContext, c *apiclient.UsageAPIClient,
...
+				responses <- res
 			}
+		}(jobs, responses)
+	}
 
-			usageMap[address] = &schema.UsageData{
-				Address:    address,
-				Attributes: attributes,
-			}
+	// Feed the workers the jobs of getting prices
```
hugorut, line 184:
```suggestion
	// Feed the workers the jobs
```

hugorut (approved):
looks good

PR #2961, opened 2024-03-20
feat: add prerelease pipeline

hugorut:
Adds two prerelease pipelines. The first `create-prerelease` which calls the release go script with a new `RELEASE_ID` env variable. This variable is a static env variable that points to the [master github prerelease](https://github.com/infracost/infracost/releases/tag/untagged-0017de5d992d120d6984) that will have its assets overwritten on each master branch push. This enables users use a master branch build if they so wish.

The second `create-docker-prerelease` uploads docker image to the registry with the master tag. It does not set it as the latest release, as per the https://github.com/docker/metadata-action docs.

I've chosen to split these two actions to make the release step a bit quicker. We can run the two build steps concurrently on prerelease as we don't have as much worry that the build might be bad.

I've also had to change the `get_version.sh` script to ignore the "preview" tag so it doesn't fail tests.

### DIFF
--- a/.github/workflows/create-docker-prerelease.yml
+++ b/.github/workflows/create-docker-prerelease.yml
@@ -1,0 +1,60 @@
+name: Create docker prerelease
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: Create docker prerelease
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ secrets.DOCKER_ORG }}/${{ secrets.DOCKER_REPOSITORY }}
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/create-docker-prerelease.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ secrets.DOCKER_ORG }}/${{ secrets.DOCKER_REPOSITORY }}
```
aliscott, line 30: I think we want to set a tag here. I think the following is the convention:
```
tags: |
  type=edge
```

We will also want to check this tag type doesn't set the latest tag, or we can add the following to be safe:
```
flavor: |
  latest=false # auto is default
```

.github/workflows/create-docker-prerelease.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: docker rmi ${{ steps.build_push_docker.outputs.imageid }} -f
+
+      - name: Docker meta (CI)
+        id: meta-ci
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ secrets.DOCKER_ORG }}/${{ secrets.DOCKER_REPOSITORY }}
+          flavor: |
+            prefix=ci-,onlatest=true
```
aliscott, line 51: Same here, we might want something like:
I think we want to set a tag here. I think the following is the convention:
```
tags: |
  type=edge
flavor: |
  prefix=ci-
  latest=false
```

aliscott (changes requested):

aliscott (approved):
LGTM

PR #2964, opened 2024-03-21
enhance: sort diff and breakdown with most costly resources first

tim775:
The change in the first two commits results in a ton a golden file changes which are updated in the remaining commits.


aliscott (approved):
Looks good, but the tests don't seem to have run.

PR #2966, opened 2024-03-26
enhance: tweak usage based output

tim775:
Update the diff output to separate baseline and usage costs in the summary table, and reduce noise by filtering out zero cost components and sub resources.


aliscott (approved):
LGTM

PR #2967, opened 2024-03-26
feat(azure): add `azurerm_network_ddos_protection_plan` support

hugorut:
Adds support for the Azure DDoS Protection service. This is a managed service that helps protected resources from DDos attacks. This service is enabled using the `azurerm_network_ddos_protection_plan` resource in Terraform.

The resource has two cost components associated with it:

1. DDoS Protection Plan: The cost of the DDoS Protection Plan.
2. Overage charges: The cost of the overage charges for the DDoS Protection Plan. This is the number of resources that fall outside the base coverage offered by the protection plan (100). This amount is defined in the usage file as it is difficult to infer the number of resources that fall outside the base coverage from the IaC.

### DIFF
--- a/internal/resources/azure/network_ddos_protection_plan.go
+++ b/internal/resources/azure/network_ddos_protection_plan.go
@@ -1,0 +1,103 @@
+package azure
+
+import (
+	"github.com/shopspring/decimal"
+
+	"github.com/infracost/infracost/internal/resources"
+	"github.com/infracost/infracost/internal/schema"
+)
+
+// NetworkDdosProtectionPlan struct represents Azure DDoS Protection Plan.
+// DDoS Protection Plan is a resource that provides DDoS protection for virtual networks and IPs.
+//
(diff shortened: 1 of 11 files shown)


internal/resources/azure/network_ddos_protection_plan.go
```diff
@@ -0,0 +1,103 @@
...
+					Service:       strPtr("Azure DDOS Protection"),
+					ProductFamily: strPtr("Networking"),
+					AttributeFilters: []*schema.AttributeFilter{
+						{Key: "skuName", Value: strPtr("Network Protection")},
+						{Key: "meterName", Value: strPtr("Network Protection Plan")},
+					},
+				},
+			},
+			{
+				Name:           "Overage charges",
```
hugorut, line 86: I have chosen this language for the cost component as this is what Azure uses in their pricing docs: https://azure.microsoft.com/en-us/pricing/details/ddos-protection/#overview

aliscott (approved):
LGTM

PR #2969, opened 2024-03-26
enhance: tweak output for usage defaults

tim775:
* Update text to clarify where usage defaults come from when there is no link
* Remove usage-based costs summary sentence
* Rename "usage estimates" to "usage defaults"
* Don't fall back to logging for usage spinners
* Log "no-usage" resources with debug instead of info


tim775:
@aliscott There are a bunch of golden file changes which you can mostly ignore, I put them in separate commits to make it easier to review.  If you review in the morning, please merge and release runners if it looks good.

internal/output/table.go
```diff
@@ -363,7 +363,7 @@ func filterZeroValComponents(costComponents []CostComponent, resourceName string
 	var filteredComponents []CostComponent
 	for _, c := range costComponents {
 		if c.MonthlyQuantity != nil && c.MonthlyQuantity.IsZero() {
-			log.Info().Msgf("Hiding cost with no usage: %s '%s'", resourceName, c.Name)
+			log.Debug().Msgf("Hiding cost with no usage: %s '%s'", resourceName, c.Name)
```
aliscott, line 366: There's two other places we log this still with info level:
* https://github.com/infracost/infracost/blob/master/internal/output/table.go#L223
* https://github.com/infracost/infracost/blob/master/internal/output/html.go#L41

cmd/infracost/run.go
```diff
@@ -564,11 +564,11 @@ func (r *parallelRunner) fetchProjectUsage(projects []*schema.Project) map[*sche
 	}
 
 	spinnerOpts := ui.SpinnerOptions{
-		EnableLogging: r.runCtx.Config.IsLogging(),
+		EnableLogging: false,
```
aliscott, line 567: I don't think this does what you want it to do. It is used to switch between the spinner and logging, so if it's set to false then the spinner still runs and we end up with output like this:
```
2024-03-26T22:33:24Z INF Starting: Evaluating Terraform directory
  ✔ Retrieving usage defaults for 9 resources from Infracost Cloud
2024-03-26T22:33:25Z INF Starting: Retrieving cloud prices to calculate costs
```

I think we should keep these ones as they were anyway since they are useful log lines and they don't add much noise. The main noise has been removed by removing the 'Hiding resource with no usage' lines.

aliscott (approved):

PR #2970, opened 2024-03-27
feat(azure): add support for `azurerm_app_configuration`

hugorut:
Adds support for Azure App Configuration, a centralized service for managing application settings and feature flags, enabling secure and dynamic configuration updates without redeployment.

This service is supported by Infracost through the `azurerm_app_configuration` Terraform resource. Which contains the following cost components:

1. Instance & replica costs - there is a cost per day for each store and replica that is enabled in Azure App Configuration
2. Requests - Users are billed for every additional 10k requests (outside of the 200k daily allowance) to the service

Additionally, Azure App Configuration has both the Free and Standard skus. The former is free to run, so we simply return a free resource if a user has specified `sku = "free"` in the `azurerm_app_configuration`.


hugorut:
I'm not sure why the `TestNetworkDdosProtectionPlan` are failing now with the usage string, seems odd as these are in the usage example file 🤷

infracost-usage-defaults.large.yml
```diff
@@ -415,6 +415,8 @@ resource_type_default_usage:
   #
   # Terraform AzureRM resources
   #
+  azurerm_app_configuration.my_configuration:
+    requests: 50000 # Monthly number of requests which are above teh 200,000 a day free tier. This is the total number of requests made to the service, including replica requests.
```
tim775, line 419: We need to update the contributing docs, but the idea with this file is to target $20/month per cost component so at $0.06/10k  that works out to about 3333333 overage requests.  The .medium should be half of that, and .small half again.

tim775 (changes requested):

hugorut:
@tim775 digging around here https://azure.microsoft.com/en-gb/pricing/calculator/ I realised that the 200k number is just the number of requests included in the base cost of the instance/replica. It is not a daily number but a monthly one.

tim775:
> @tim775 digging around here https://azure.microsoft.com/en-gb/pricing/calculator/ I realised that the 200k number is just the number of requests included in the base cost of the instance/replica. It is not a daily number but a monthly one.

Are you sure?  I know it's not definitive, but [this page](https://learn.microsoft.com/en-us/azure/azure-app-configuration/faq) says:

> Cost: Standard tier stores have a daily usage charge. The first 200,000 requests each day are included in the daily charge. There's also an overage charge for requests past the daily allocation. There's no cost to use a free tier store.

And the the $1.20/day per replica matches up with the 200,000 @ $0.06 per 10k.  So it seems like the 200,000 is kind of like a "minimum" requests per day.

tim775 (approved):

hugorut:
> Are you sure? I know it's not definitive, but [this page](https://learn.microsoft.com/en-us/azure/azure-app-configuration/faq)

Hmm i am not sure.. but that page looks semi-definitive, i'll revert the last commit and merge with the replica/day messaging

PR #2973, opened 2024-03-28
enhance: trim repo name prefix when outputting project name

tim775:
This reduces noise by omitting the repo name if it is part of the project name.


hugorut (approved):
Seems like some tests still need to be updated `TestBreakdownWithPrivateTerraformRegistryModule` is failing the golden tests

tim775:
We're scrapping this change in favor of a future change which will make `infracost breakdown` and `infracost diff` use the same logic for generate project names as is used in `infracost generate config`.  This will go in a future v0.11.* since it is a breaking change.

PR #2974, opened 2024-03-28
feat(azure): add support for `azurerm_federated_identity_credential`

hugorut:
Adds support for Microsoft Entra (Azure Active Directory) External Identities through the `azurerm_federated_identity_credential`. This resource is only billed for monthly active users above the 50000 free limit. Monthly active users are charged differently based on the Microsoft Entra license type (P1 vs P2), however we cannot infer the license type from the IaC so I've added two usage parameters to support both.

### DIFF
--- a/infracost-usage-example.yml
+++ b/infracost-usage-example.yml
@@ -834,6 +834,10 @@ resource_usage:
     monthly_jobs_compute_dbu_hrs: 1000 # Monthly number of Jobs Compute Databricks Units in DBU-hours.
     monthly_jobs_light_compute_dbu_hrs: 2000 # Monthly number of Jobs Light Compute Databricks Units in DBU-hours.
 
+  azurerm_federated_identity_credential.my_credential:
+    monthly_active_p1_users: 50100 # Monthly number of active users if you are using the Premium P1 license.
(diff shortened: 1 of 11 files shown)


infracost-usage-example.yml
```diff
@@ -834,6 +834,10 @@ resource_usage:
     monthly_jobs_compute_dbu_hrs: 1000 # Monthly number of Jobs Compute Databricks Units in DBU-hours.
     monthly_jobs_light_compute_dbu_hrs: 2000 # Monthly number of Jobs Light Compute Databricks Units in DBU-hours.
 
+  azurerm_federated_identity_credential.my_credential:
+    monthly_active_p1_users: 100 # Monthly number of active users above the 50000 free limit if you are using the Premium P1 license.
+    monthly_active_p2_users: 100 # Monthly number of active users above the 50000 free limit if you are using the Premium P2 license.
+
```
aliscott, line 840: I think we should either:
1. Call this `monthly_additional_active_pX_users` or similar if the free amount is across all resources of this type.
2. Subtract 50,000 from this number if the free amount is per resource.

This is what we do for other resources (e.g. `azurerm_network_watcher.monthly_diagnostic_checks`).,

hugorut: will go with 2

aliscott (approved):

PR #2980, opened 2024-04-01
chore: update usage file headers

alikhajeh1:


alikhajeh1:
@tim775 Why are some of the usage defaults commented out? seems confusing

tim775:
> @tim775 Why are some of the usage defaults commented out? seems confusing

The CLI needs to be updated to support cloud based defaults for those resources.

tim775 (approved):

PR #2994, opened 2024-04-02
chore: Add descriptions to usage default keys

tim775:
The usage default keys should have the same comments so they can be picked up Infracost Cloud for use in the overrides UI.


aliscott (approved):
Thanks!

PR #3003, opened 2024-04-05
refactor: log output 

hugorut:
Refactors the progress/log output for the evaluation/project evaluation.
Chnages the output of status lines so that they are simpler and more consistent
across CLI and CI.

* Cleans out spinner functionality as we move to a more consistent log based approach.
Lot's of old spinners have been demoted to `Debug` level log lines as we want to keep
the `Info` level relatively clean of noise.
* All fmt.Prints are now piped to the logs, this includes all the previous ui
  package interactions, appart from a few fatal errors.
* Some Warn statements are now changed to debug
* By default the log level is INFO and is not piped to io.Discard if not explicitly
  set.
* The log formatter now checks if the env is CI and formats the log accordingly
* Local logs have their timestamps omitted and additional colours
* Now we're using logging as the default console writer, I've also updated all the base testutil tests to capture the logs by default. I'm not sure why we didn't do this before and required non parrallel tests to capture the logs. It seems fine to capture in the buffer as implemented. Perhaps there is something i'm missing?


cmd/infracost/run.go
```diff
@@ -352,20 +416,20 @@ func (r *parallelRunner) runProvider(job projectJob) (*projectOutput, error) {
...
-	}
+	name := job.provider.ProjectName()
+	displayName := ui.ProjectDisplayName(r.runCtx, name)
 
-	if r.runCtx.Config.IsLogging() {
-		log.Info().Msg(m)
-	} else {
-		fmt.Fprintln(os.Stderr, m)
-	}
+	logging.Logger.Info().Msgf("Starting evaluation for project %s", displayName)
```
aliscott, line 422: After running this on a few large repos I think we should move this and the 'Finished evaluation...' logs to debug since they create too many log lines breaking up any warnings. In the future we probably want to have some sort of progress indicator instead.

aliscott (commented):
For some reason I see the following logs with this branch when I run against our internal infra:

```
WARN could not request Terraform workspace: my-workespace for organization: my-org error="the provided token is invalid" parser_path=../my-infra/dev project_name= project_path=../my-infra provider=terraform_dir
```

I don't see it in master so not sure what changed. This warning doesn't seem like it should be a warning since I don't have any remote vars in TFC to pull down, so I shouldn't need to provide a token.

aliscott (approved):

PR #3009, opened 2024-04-09
enhance: Remove links from usage cost headers to reduce noise in comments

vdmgolub:
GitHub (has title as tooltip):
image

GitLab (has title as tooltip):
image

Azure Repos (has title as tooltip):
image

Bitbucket (no tooltip):
image


aliscott (approved):
LGTM

PR #3015, opened 2024-04-11
feat: Add hasUsageFile flag to run metadata

vdmgolub:


tim775:
Do we need a new field here? It seems like we can check for the presence of `usageFilePath`

PR #3019, opened 2024-04-17
fix: don't remove child of errored prior projects that exist in current

tim775:
If a nested terragrunt project has syntax errors, the error is reported at the parent project path. To prevent the child project from appearing as "removed" when diff'd against a baseline without the syntax error, we search for and remove child projects of the parent path. BUT, since some child projects may have been processed successfully, we should only remove past children that don't exist in the current.


hugorut (approved):
Lgtm

PR #3021, opened 2024-04-17
fix: dynamic block names in logs

aliscott:
Previously these appeared in logs as `block_name=content.` This now fixes it so it shows `block_name=aws_security_group.my_sg.dynamic.ingress.content`.

We do this by looping through any parent blocks and prefixing the block name with their labels.


hugorut (approved):
looks good, let's make sure this is tested on some important repos

PR #3026, opened 2024-04-22
fix: don't autodetect paths if `--terraform-var-file` is used explicitly

hugorut:
Changes `Detect` functionality to skip autodetection if a Terraform var file is explicitly provided by the user. Terraform var file paths are relative to the project root, so it doesn't make sense to autodetect directories if an explicit var path has been used.

The original ticket for this change also mentioned that we should omit autodetect if `--terraform-var` was also used. I've decided not to do this, as I think this might be a valid case. For example providing a default variable across a set of autodetected projects, e.g. `git_sha`.


aliscott (approved):
> The original ticket for this change also mentioned that we should omit autodetect if --terraform-var was also used. I've decided not to do this, as I think this might be a valid case. For example providing a default variable across a set of autodetected projects, e.g. git_sha.

Makes sense. What is the precendence? It seems `terraform-var` should override anything autodetected?

PR #3041, opened 2024-04-26
feat: readd templatefile support

hugorut:
Changes `isPathInRepo` to use wd directory instead of `RepoPath` which is likely incorrect for Terragrunt projects.

The `RepoPath` property that is infered from detected projects was returning full project path for detected projects when the autdetected project was inferred from a config file. A fix for this will be addressed in a further PR.

To fix the templatefile logic I've opted to use the os working directory instead, as this is more robust.

### DIFF
--- a/internal/hcl/evaluator.go
+++ b/internal/hcl/evaluator.go
@@ -1150,7 +1150,7 @@ func (e *Evaluator) loadModules(lastContext hcl.EvalContext) {
 // ExpFunctions returns the set of functions that should be used to when evaluating
 // expressions in the receiving scope.
 func ExpFunctions(baseDir string, logger zerolog.Logger) map[string]function.Function {
-	return map[string]function.Function{
+	fns := map[string]function.Function{
 		"abs":              stdlib.AbsoluteFunc,
 		"abspath":          funcs.AbsPathFunc,
 		"basename":         funcs.BasenameFunc,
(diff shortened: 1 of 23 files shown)


internal/hcl/evaluator.go
```diff
@@ -115,7 +115,9 @@ func NewEvaluator(
 	isGraph bool,
 ) *Evaluator {
 	ctx := NewContext(&hcl.EvalContext{
-		Functions: ExpFunctions(module.RootPath, logger),
+		// Functions require the module path to be passed in rather than the RootPath
+		// as the latter will not resolve correct relative paths for file functions.
+		Functions: ExpFunctions(module.ModulePath, logger),
```
hugorut, line 120: This resolves a hidden bug that meant relative paths used in file functions in modules would always fail as the `basePath` was always the top level `RootPath`. E.g. a project with the following structure:

```
.
└── terraform
    ├── main.tf
    └── modules
        └── foo
            ├── file.json
            └── main.tf
```

The child `Evaluator` for module `foo` would use the `basePath` `terraform` in `ExpFunctions`. Meaning that if a the hcl tried to open a file.json e.g.`file("file.json")` we would attempt to open `terraform/file.json` which did not exist. 

Passing the `ModulePath` means that the base path resolves to `terraform/module/foo`, solving the relative filepath issue.

aliscott (approved):

PR #3059, opened 2024-05-02
feat: Add vcsBaseCommitSha to estimate metadata

vdmgolub:
It's empty on breakdown, and base branch's HEAD SHA on diff/output.
NB: When combining several JSONs in output command latest item's metadata will be used in the output.


tim775 (approved):
Can you update TestDiffWithCompareToFormatJSON and the other tests to show vcsBaseCommitSha being set properly.

PR #3062, opened 2024-05-07
fix: aws/azure failing tests

hugorut:
Commits the extended support costs for eks cluster and lightsail instance test. Changes the Azure AI services tests meter naming to fix prices not found. This also includes minor changes to the cognitive service pricing.

### DIFF
--- a/internal/output/table.go
+++ b/internal/output/table.go
@@ -283,13 +283,16 @@ func buildCostComponentRows(t table.Writer, currency string, costComponents []Co
 				c.Unit,
 			)
 
+			if c.PriceNotFound {
+				price = "not found"
+			}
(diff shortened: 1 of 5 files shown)


internal/output/table.go
```diff
@@ -283,13 +283,16 @@ func buildCostComponentRows(t table.Writer, currency string, costComponents []Co
 				c.Unit,
 			)
 
+			if c.PriceNotFound {
+				price = "not found"
+			}
```
hugorut, line 288: Changes the output table to show "not found" when component is a usage component - this is line with the non usage components.

aliscott (approved):
Thanks!

PR #3069, opened 2024-05-15
fix: support launch templates references which use `prefix`

hugorut:
This PR adds support for launch templates which are referenced by `name` in a resource but use a `prefix` to launch. This means that the `name` attribute can be `null` and prior to this change, references would not work. This change adds a Infracost generated `name` if it is detected that the `launch_template` has a `null` name.


aliscott (approved):
Neat fix

PR #3070, opened 2024-05-17
fix: propagate default provider tags to volume tags

hugorut:
Fixes issue where provider tags were not being used as defaults in volume tags. Provider defaults propagating to `volume_tags` was introduced in version `v5.39.0` of the aws provider https://github.com/hashicorp/terraform-provider-aws/issues/19890. 

To accomodate this functionality we add any of the default tags with the `volume_tags` prefix if the current resource supports `volume_tags` (currently just `aws_instance`). This is only done if the version constraint specified at the root level `terraform` `require_providers` block is compatible with the `5.39.0` version. e.g. `>= 5.0`. Versions which specify a maximum version, e.g. `~> 3.0` below `5.39` will not have the default tags passed to volume tags. If no version constraint is found/can be parsed we add default tags to volume_tags by default. This is to ensure we have no false positive tagging policy failures.

### DIFF
--- a/internal/providers/terraform/aws/tags.go
+++ b/internal/providers/terraform/aws/tags.go
@@ -49,6 +49,14 @@ func ParseTags(defaultTags *map[string]string, r *schema.ResourceData) *map[stri
 			keysAndValues = append(keysAndValues, k, v)
 		}
 
+		if r.Type == "aws_instance" {
+			for k, v := range *defaultTags {
+				k = fmt.Sprintf("volume_tags.%s", k)
+				tags[k] = v
+				keysAndValues = append(keysAndValues, k, v)
+			}
+		}
+
 		sort.Strings(keysAndValues)
 
 		h := sha256.New()
--- a/internal/providers/terraform/aws/tags.go
+++ b/internal/providers/terraform/aws/tags.go
@@ -29,7 +29,15 @@ func parseLaunchTemplateTags(tags map[string]string, r *schema.ResourceData) {
 	}
 }
 
-func ParseTags(defaultTags *map[string]string, r *schema.ResourceData) *map[string]string {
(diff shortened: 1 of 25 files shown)


internal/providers/terraform/aws/tags.go
```diff
@@ -49,6 +49,14 @@ func ParseTags(defaultTags *map[string]string, r *schema.ResourceData) *map[stri
 			keysAndValues = append(keysAndValues, k, v)
 		}
 
+		if r.Type == "aws_instance" {
+			for k, v := range *defaultTags {
```
tim775, line 53: Are we sure this works correctly? I didn't think terraform applied default tags to volume tags like this.

hugorut: yup after some testing with AWS this doesn't seem to actually be the case

hugorut: No sorry, default tags are actually propagated to the volume_tags but only in `v5.39.0` and above https://github.com/hashicorp/terraform-provider-aws/issues/19890. We'll have to write some logic to detect the provider version and propagate accordingly.

hugorut:
@tim775 can you have another look at this - see the updated PR desc for more info. Note: the team decided the propagation of default tags to volume_tags should happen by default unless a user has explicitly specified a version that is below `5.39`

tim775 (approved):
🙇 

This makes me wonder if the tag propogation should be handled in the hcl parsing or in the infracost json like we're doing now. I had been thinking it should probably be done in hcl, but maybe not if that would mean we have to add all the resource and provider specific rules in there.

hugorut:
@tim775 yeah I'm thinking it would be a lot easier/cleaner if we do it in HCL, but for now it's probably just a bit too much.

PR #3072, opened 2024-05-22
fix: index expression mocking

aliscott:
Before when we mocked index expressions, we would get an 'Invalid index' diagnostic, since we'd mock the key to value that was never in the collection.

This updates it to first check if the mocking causes an 'Invalid index' diagnostic, and if it does we return the first element from the collection. This may be the incorrect value, but seems the safest since at least the value returned will be the expected type.

If we see other diagnostics we return an empty string value. This may cause issues if the collection isn't meant to be a list of strings, but since we can't evaluate it this seems the safest.

### DIFF
--- a/internal/hcl/attribute.go
+++ b/internal/hcl/attribute.go
@@ -456,6 +456,8 @@ func (e *LiteralBoolValueExpression) Value(ctx *hcl.EvalContext) (cty.Value, hcl
 }
 
 type LiteralValueCollectionExpression struct {
+	// we embed the hclsyntax.LiteralValueExpr as the hcl.Expression interface
+	// has an unexported method that we need to implement.
 	*hclsyntax.LiteralValueExpr
 	Expression  hcl.Expression
 	MockedValue cty.Value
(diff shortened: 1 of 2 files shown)


internal/hcl/attribute.go
```diff
@@ -482,6 +484,47 @@ func (e *LiteralValueCollectionExpression) Value(ctx *hcl.EvalContext) (cty.Valu
...
+					return v, nil
+				}
+			}
+		}
+	}
+
+	// For other diagnostics we just return an empty string
+	// as we can't determine the correct value.
+	if diag.HasErrors() {
+		return cty.StringVal(""), nil
```
hugorut, line 522: can we instead return the `mockedVal` that is passed to `mockExpressionCalls` - this keeps it consistent with the rest of the logic. It also allows us to override what mock we return (e.g. objectvalue, list, e.t.c).

The `newLiteralValueIndexExpression` can take `mockedVal` and set it as a property which we return here

hugorut (approved):
slight tweak but looks good

PR #3075, opened 2024-05-23
fix: don't use existing context when evaluating resources, data or provider blocks

aliscott:
The `SetByDot` function handles the merging anyway, so we don't need to pull out the existing values. I've left the old functionality in the `evaluateResource` and `evaluatorProvider` functions because the legacy evaluator needs it.

This also fixes an issue where the data blocks were using `e.ctx.Get(b.TypeLabel())` which was getting the wrong context. The correct functionality would have been to use `e.ctx.Get("data")`. Otherwise, this caused issues where the type of a data block matched the name of a resource, e.g. `data.aws_lb.anything` and `aws_lb.aws_lb`. In this case it would pull the context of the resource, which could cause a panic if the resource had been expanded, because it would be an array instead of the expected map.

### DIFF
--- a/internal/hcl/graph_vertex_data.go
+++ b/internal/hcl/graph_vertex_data.go
@@ -81,7 +81,7 @@ func (v *VertexData) evaluate(e *Evaluator, b *Block) error {
 		existingVals = make(map[string]cty.Value)
 	}
 
-	val := e.evaluateResource(b, existingVals)
+	val := e.evaluateResourceOrData(b, existingVals)
 
 	v.logger.Debug().Msgf("adding data %s to the evaluation context", v.ID())
 	e.ctx.SetByDot(val, fmt.Sprintf("data.%s", b.TypeLabel()))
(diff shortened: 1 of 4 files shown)


internal/hcl/graph_vertex_data.go
```diff
@@ -73,15 +73,8 @@ func (v *VertexData) evaluate(e *Evaluator, b *Block) error {
...
-	existingCtx := e.ctx.Get(b.TypeLabel())
-	if !existingCtx.IsNull() {
-		existingVals = existingCtx.AsValueMap()
-	} else {
-		existingVals = make(map[string]cty.Value)
-	}
-
-	val := e.evaluateResource(b, existingVals)
+	// We don't care about the existing values, this is only needed by the legacy evaluator
+	val := e.evaluateResource(b, map[string]cty.Value{})
```
hugorut, line 77: Is it possible if we can change this method `evaluateResource` to be something like `evaluateResourceOrDataBlock` (there's probably a better name than that). Then if we could add some comments to that method about the fact that the `values` argument isn't really needed for the graph evaluator. Just so we are clear in the future when we inevitably hit some bug e.t.c.

hugorut (approved):
👍

PR #3088, opened 2024-05-30
feat(aws): Add support for Aurora Serverless V2

aliscott:
I added the cost for this at the `aws_rds_cluster_instance` level because it seems you need to create those for Aurora RDS Serverless, and the cost is the ACUs used multiplied by the instances.

There is a `serverlessv2_scaling_configuration` attribute for the [aws_rds_cluster](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster), which we could use to have a min and max for the ACUs, but for just now I've kept it consistent with Serverless V1.

### DIFF
--- a/internal/providers/terraform/aws/testdata/lightsail_instance_test/lightsail_instance_test.golden
+++ b/internal/providers/terraform/aws/testdata/lightsail_instance_test/lightsail_instance_test.golden
@@ -2,12 +2,12 @@
  Name                            Monthly Qty  Unit   Monthly Cost   
                                                                     
  aws_lightsail_instance.linux1                                      
- └─ Virtual server (Linux/UNIX)          730  hours        $82.42   
+ └─ Virtual server (Linux/UNIX)          730  hours        $78.49   
                                                                     
  aws_lightsail_instance.win1                                        
- └─ Virtual server (Windows)             730  hours        $19.62   
+ └─ Virtual server (Windows)             730  hours        $21.58   
                                                                     
- OVERALL TOTAL                                           $102.04 
+ OVERALL TOTAL                                           $100.07 
 
 *Usage costs can be estimated by updating Infracost Cloud settings, see docs for other options.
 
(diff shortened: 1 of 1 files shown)


internal/resources/aws/rds_cluster_instance.go
```diff
@@ -148,6 +108,86 @@ func (r *RDSClusterInstance) databaseEngineValue() string {
...
+		ProductFilter: &schema.ProductFilter{
+			VendorName:    strPtr("aws"),
+			Region:        strPtr(r.Region),
+			Service:       strPtr("AmazonRDS"),
+			ProductFamily: strPtr("ServerlessV2"),
+			AttributeFilters: []*schema.AttributeFilter{
+				{Key: "databaseEngine", Value: strPtr(databaseEngine)},
+				{Key: "usagetype", ValueRegex: regexPtr(usageType)},
+			},
+		},
```
tim775, line 187: Since this is a usage based cost component
```suggestion
		},
		UsageBased: true,
```

infracost-usage-example.yml
```diff
@@ -335,11 +335,12 @@ resource_usage:
...
+    monthly_cpu_credit_hrs: 24                              # Number of hours in a month, where you expect to burst the baseline credit balance of a "t3" instance type.
+    vcpu_count: 2                                           # (DEPRECATED this is now calculated automatically) Number of virtual CPUs allocated to your "t3" instance type. Currently instances with 2 vCPUs are available.
     monthly_additional_performance_insights_requests: 10000 # Monthly Performance Insights API requests above the 1000000 requests included in the free tier.
     reserved_instance_term: 1_year                          # Term for Reserved Instances, can be: 1_year, 3_year.
     reserved_instance_payment_option: partial_upfront       # Payment option for Reserved Instances, can be: no_upfront (only for 1_year term), partial_upfront, all_upfront.
+    capacity_units_per_hr: 50                               # Number of Aurora Capacity Units per hour. Only used for Aurora Serverless V2 (when instance class is "db.serverless")
```
tim775, line 343: we should also update instance-usage-defaults.*.yml and run `scripts/generateGlobalDefaultUsage.ts` in the dashboard.

tim775 (approved):

infracost-usage-example.yml
aliscott: I've updated these. The problem is the minimum is 0.5 which costs $44 a month, so I've had to use that for all the sizes.

PR #3096, opened 2024-06-04
fix: Add tfvar files to the project dependency paths

owenrumney:
- ensure that the tfvars files are included with the modules in the
  dependency paths block of the config file
- add tests to explicitly test modules and tfvars to ensure both are
  included
- update the the existing tests expected.golden files

```
  - path: db/prod
    name: db-prod-prod
    terraform_var_files:
      - ../../default.tfvars
      - ../../envs/prod.tfvars
    skip_autodetect: true
    dependency_paths:
      - db/modules/db_config
      - db/prod
      - ../../default.tfvars
      - ../../envs/prod.tfvars
```


hugorut (approved):
LGTM 🚀

PR #3099, opened 2024-06-04
enhance: Update PR comment's cost change summary

vdmgolub:


internal/output/diff.go
```diff
@@ -188,11 +188,18 @@ func tableForDiff(out Root, opts Options) string {
 			continue
 		}
 
+		var usageCost string
+		if out.Metadata.UsageApiEnabled {
```
tim775, line 192: I'm worried that this condition is not sufficient because usage costs can also come from the usage file.  Essentially, we only want to show the "-" when `usageCostsMessage` is the "Usage costs CAN be estimated..." one.  It looks like that works out to showing - when `!out.Metadata.UsageApiEnabled && out.Metadata.UsageFilePath == "" && !out.Metadata.ConfigFileHasUsageFile`

as a sanity check in the API comment I also checked to make sure that both usage costs were zero or not present (I was worried about putting in a "-" and then having the total change not equal baseline change.

tim775 (changes requested):

cmd/infracost/testdata/diff_terraform_plan_json/diff_terraform_plan_json.golden
```diff
@@ -67,12 +67,12 @@ Key: * usage cost, ~ changed, + added, - removed
...
-┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
-┃ Changed project                                              ┃ Baseline cost ┃ Usage cost ┃ Total change ┃
-┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━╋━━━━━━━━━━━━╋━━━━━━━━━━━━━━┫
-┃ infracost/infracost/cmd/infracost/testdata/example_plan.json ┃         +$925 ┃      +$437 ┃      +$1,361 ┃
-┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━┻━━━━━━━━━━━━┻━━━━━━━━━━━━━━┛
+Infracost estimate: Monthly estimate increased by $1,361 ↑
+┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┓
+┃ Changed project                                              ┃ Baseline cost ┃ Usage cost* ┃ Total change ┃
+┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━╋━━━━━━━━━━━━━━┫
+┃ infracost/infracost/cmd/infracost/testdata/example_plan.json ┃         +$925 ┃           - ┃      +$1,361 ┃
```
vdmgolub, line 74: @vdmgolub Missed this one.

internal/output/format.go
```diff
@@ -30,6 +30,16 @@ func formatCost(currency string, d *decimal.Decimal) string {
 	return formatRoundedDecimalCurrency(currency, *d)
 }
 
+func formatUsageCost(out Root, cost *decimal.Decimal) string {
+	noUsageCosts := cost == nil || cost.Equals(decimal.Zero)
+
+	if usageCostsEnabled(out) || noUsageCosts {
+		return formatCost(out.Currency, cost)
```
tim775, line 37: I think no usage costs check is backwards right? We show the cost if usageCostsEnabled or cost is non-zero


```suggestion
	hasUsageCosts := cost != nil && !cost.Equals(decimal.Zero)

	if usageCostsEnabled(out) || hasUsageCosts {
		return formatCost(out.Currency, cost)
```

tim775 (approved):

internal/output/format.go
vdmgolub: Oops, of course 👍

internal/output/diff.go
```diff
@@ -191,9 +191,9 @@ func tableForDiff(out Root, opts Options) string {
 		t.AppendRow(
 			table.Row{
 				truncateMiddle(project.Name, 64, "..."),
-				formatMarkdownCostChange(out.Currency, project.PastBreakdown.TotalMonthlyBaselineCost(), project.Breakdown.TotalMonthlyBaselineCost(), false, true),
-				formatMarkdownCostChange(out.Currency, project.PastBreakdown.TotalMonthlyUsageCost, project.Breakdown.TotalMonthlyUsageCost, false, true),
-				formatMarkdownCostChange(out.Currency, project.PastBreakdown.TotalMonthlyCost, project.Breakdown.TotalMonthlyCost, false, false),
+				formatMarkdownCostChange(out.Currency, project.PastBreakdown.TotalMonthlyBaselineCost(), project.Breakdown.TotalMonthlyBaselineCost(), false, true, false),
+				formatMarkdownCostChange(out.Currency, project.PastBreakdown.TotalMonthlyUsageCost, project.Breakdown.TotalMonthlyUsageCost, false, true, !usageCostsEnabled(out)),
```
vdmgolub, line 195: @tim775 This should also be inverted: when usage costs are disabled, `skipIfZero` is true, so we'll show dashes for zeros.

PR #3103, opened 2024-06-07
feat: Add experimental --comment-path flag to comment command

vdmgolub:
This flag expects a file with markdown comment content to post it to the PR/commit. When this flag is provided the --path flag is not needed.


cmd/infracost/comment_azure_repos.go
```diff
@@ -132,8 +138,10 @@ func commentAzureReposCmd(ctx *config.RunContext) *cobra.Command {
 	cmd.Flags().String("azure-access-token", "", "Azure DevOps access token")
 	_ = cmd.MarkFlagRequired("azure-access-token")
 	cmd.Flags().StringArrayP("path", "p", []string{}, "Path to Infracost JSON files, glob patterns need quotes")
-	_ = cmd.MarkFlagRequired("path")
 	_ = cmd.MarkFlagFilename("path", "json")
+	cmd.Flags().String("comment-path", "", "Path to comment markdown file (experimental)")
```
tim775, line 142: fwiw, this doesn't have to be markdown

tim775 (approved):

cmd/infracost/comment_azure_repos.go
vdmgolub: I agree, but I thought it's better to keep it consistent as markdown also supports HTML.

PR #3105, opened 2024-06-10
fix Get the relative path based on the starting path for var files

owenrumney:
Pattern matching doesn't work for the tfvar dependency paths when they're relative but works okay when they're relative to the starting path like module calls


hugorut (approved):
remember in future todo conventional style commits e.g. `fix: Get the relative path based on the starting path for var files`

PR #3107, opened 2024-06-11
refactor: add `region` field to resource data

hugorut:
Adds a `Region` property to the `schema.ResourceData` which is set when parsing the plan and set before evaluating each individual resource. This property is meant to replace the old `d.Get` approach which sets the region directly on the `RawValues` of the json.

This new approach sets the `Region` after the references have been parsed and changes the `azure` region functionality to use the `getLocation` logic that was previously done on demand per resource.

---

Adds a resource specific `GetRegion` property to the `RegistryItem` which
can be used to override the provider default functionality for fetching a
region for a resource.

Changes usage of `lookupRegion` functions in azure resources to use the new
`ResourceData.Region` property which is computed before evaluation.

Changes the `ResourceRegistryMap` to no longer use a `sync.Once` and instead be
built as a singleton `var`. This removes the sync requirement as callers only read
from this map.

### DIFF
--- a/internal/providers/terraform/azure/util.go
+++ b/internal/providers/terraform/azure/util.go
@@ -29,6 +29,113 @@ func toAzureCLIName(location string) string {
 	return strings.ToLower(sReg.ReplaceAllString(location, ""))
 }
 
+var lookupRegionFunctions = map[string]func(d *schema.ResourceData) string{
+	"azurerm_mssql_elasticpool": func(d *schema.ResourceData) string {
+		return lookupRegion(d, []string{"server_name", "resource_group_name"})
+	},
+	"azurerm_lb_rule": func(d *schema.ResourceData) string {
+		return lookupRegion(d, []string{"loadbalancer_id"})
+	},
+	"azurerm_data_factory_integration_runtime_azure": func(d *schema.ResourceData) string {
+		return lookupRegion(d, []string{"resource_group_name", "data_factory_id", "data_factory_name"})
+	},
+	"azurerm_traffic_manager_external_endpoint": func(d *schema.ResourceData) string {
+		if len(d.References("profile_id")) > 0 {
+			profile := d.References("profile_id")[0]
+			return lookupRegion(profile, []string{"resource_group_name"})
+		}
+
+		return lookupRegion(d, []string{"resource_group_name"})
+	},
+	"azurerm_storage_queue": func(d *schema.ResourceData) string {
+		return lookupRegion(d, []string{"storage_account_name"})
+	},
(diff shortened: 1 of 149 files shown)


internal/providers/terraform/azure/util.go
```diff
@@ -63,6 +170,24 @@ func convertRegion(region string) string {
 	}
 }
 
+// GetResourceRegion returns the region for the given azure resource data. By default,
+// this uses the "location" property, but if that is not set it will traverse the
+// references to find a location. The default references used are the resource
+// group name. However, some resources specify different resources to get the
+// location from, in which case a custom function is used that specifies which
+// references to use.
+func GetResourceRegion(d *schema.ResourceData) string {
```
aliscott, line 179: 2 other options here:

1. Should we change the `terraform/azure/*.go` resources to call this function instead of `lookupRegion` directly?

2. Should we put the `lookupRegion` function in the `RegistryItem` struct like we do for `ReferenceAttributes`? That way it can live alongside the other Terraform resource code in the `terraform/azure/*.go`

internal/providers/terraform/hcl_provider.go
```diff
@@ -148,7 +148,7 @@ func NewHCLProvider(ctx *config.ProjectContext, rootPath hcl.RootPath, config *H
 
 	var policyClient *apiclient.PolicyAPIClient
 	if runCtx.Config.PoliciesEnabled {
-		policyClient, err = apiclient.NewPolicyAPIClient(runCtx)
+		policyClient, _ = apiclient.NewPolicyAPIClient(runCtx)
```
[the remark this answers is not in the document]
hugorut: nope

internal/providers/terraform/azure/util.go
hugorut: 1. technically we don't have to now, we can just use the `Region` property on the `SchemaData` which is now set. This will be correct for each resource. I initially left this out because I thought it might be too big a change, but happy to change all the areas we use `lookupRegion` to use the property instead
2. Yeah I guess that would maybe work better

hugorut: I have now changed this function so that it only gets the default fallback for the provider (azure). In `terraform/parser.go` we now call the `RegistryItem.GetRegion` func if it exists otherwise we fallback to this function.

```go
regionFunc := ResourceRegistryMap.GetRegion(d.Type)
if regionFunc != nil {
    return regionFunc(d)
}

 ...
 
 return azure.GetResourceRegion(d)
```

hugorut: Additionally all azure resources now use the `d.Region` property at runtime rather than the `lookupRegion`.

aliscott (approved):
Looks nice!

PR #3108, opened 2024-06-11
enhance: Make --comment-path flag optional keeping --path required

vdmgolub:
Without infracost.json validAt timestamp is missing that can lead to posting stale comments if `infracost comment` is run simultaneously.


cmd/infracost/comment.go
```diff
@@ -119,6 +111,20 @@ func buildCommentOutput(cmd *cobra.Command, ctx *config.RunContext, paths []stri
...
+		commentData, err := output.LoadCommentData(commentPath)
+		if err != nil {
+			return nil, fmt.Errorf("Error loading %s used by --comment-path flag. %s", commentPath, err)
+		}
+
+		return &CommentOutput{
+			Body:    commentData,
+			HasDiff: combined.HasDiff(),
+			ValidAt: &combined.TimeGenerated,
+		}, nil
```
tim775, line 125: seems like we should keep the error return logic (policy & governance).  These will be important so github actions runner can correctly fail if governance comment fails

cmd/infracost/comment.go
vdmgolub, line 1: @tim775 Here's the fix.

tim775 (approved):

PR #3113, opened 2024-06-27
fix: Take PITR enabled into account

owenrumney:
Update the DynamoDB resource to read the `point_in_time_recovery` block
to ensure that it is enabled so PITR costs aren't inadvertently added to
the cost estimate.

Add test to validate

### DIFF
--- a/internal/providers/terraform/aws/dynamodb_table.go
+++ b/internal/providers/terraform/aws/dynamodb_table.go
@@ -40,6 +40,13 @@ func NewDynamoDBTableResource(d *schema.ResourceData) schema.CoreResource {
 		targets = append(targets, newAppAutoscalingTarget(ref, ref.UsageData))
 	}
 
+	pitr := &aws.PointInTimeRecovery{Enabled: false}
+	if d.Get("point_in_time_recovery").Exists() {
+		pitr = &aws.PointInTimeRecovery{
+			Enabled: d.Get("point_in_time_recovery").Array()[0].Get("enabled").Bool(),
+		}
+	}
+
 	a := &aws.DynamoDBTable{
 		Address:              d.Address,
 		Region:               d.Get("region").String(),
@@ -49,6 +56,7 @@ func NewDynamoDBTableResource(d *schema.ResourceData) schema.CoreResource {
 		ReadCapacity:         intPtr(d.Get("read_capacity").Int()),
 		ReplicaRegions:       replicaRegions,
 		AppAutoscalingTarget: targets,
+		PointInTypeRecovery:  pitr,
 	}
 	return a
 }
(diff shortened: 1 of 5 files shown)


internal/providers/terraform/aws/dynamodb_table.go
```diff
@@ -40,6 +40,13 @@ func NewDynamoDBTableResource(d *schema.ResourceData) schema.CoreResource {
 		targets = append(targets, newAppAutoscalingTarget(ref, ref.UsageData))
 	}
 
+	pitr := &aws.PointInTimeRecovery{Enabled: false}
+	if d.Get("point_in_time_recovery").Exists() {
+		pitr = &aws.PointInTimeRecovery{
+			Enabled: d.Get("point_in_time_recovery").Array()[0].Get("enabled").Bool(),
```
hugorut, line 46: probably want to check the length here - this could cause exceptions with malformed code (often see with vscode)

internal/resources/aws/dynamodb_table.go
```diff
@@ -11,6 +11,10 @@ import (
 	"github.com/infracost/infracost/internal/usage/aws"
 )
 
+type PointInTimeRecovery struct {
+	Enabled bool `infracost_usage:"enabled"`
+}
+
```
hugorut, line 17: I'm not sure why you have usage keys here, I don't think this works. The `infracost_usage` relates to the specific key in the usage file under `aws_dynamodb_table`. 

1. `enabled` is not defined
2. the nesting is not handled.

but I don't believe usage is required here as this attribute is inferred from the code

internal/providers/terraform/aws/dynamodb_table.go
```diff
@@ -40,6 +40,13 @@ func NewDynamoDBTableResource(d *schema.ResourceData) schema.CoreResource {
 		targets = append(targets, newAppAutoscalingTarget(ref, ref.UsageData))
 	}
 
+	pitr := &aws.PointInTimeRecovery{Enabled: false}
+	if d.Get("point_in_time_recovery").Exists() {
```
hugorut, line 44: Would prefer we just use a boolean flag here rather than a separate type, no real need to. Just add a property on the dynamo db type called `PointInTimeRecoveryEnabled`

internal/resources/aws/dynamodb_table.go
owenrumney: yup, removed - misunderstanding on my part about what that annotation was for when I wrote it

internal/providers/terraform/aws/dynamodb_table.go
owenrumney: Yup, will do. Was relying on the length check inside `Exists`, but now checking explicitly

internal/providers/terraform/aws/dynamodb_table.go
owenrumney: Changed to a straight bool flag

hugorut (approved):

PR #3128, opened 2024-07-10
feat: Add support for tag propagation checks

liamg:
Collects a summary of tag propagation info for each resource that supports it, which is later used when checking tag policies to ensure the required tags will be propagated to dynamically created resources.

### DIFF
--- a/cmd/infracost/testdata/breakdown_with_policy_data_upload_hcl/breakdown_with_policy_data_upload_hcl-upload.golden
+++ b/cmd/infracost/testdata/breakdown_with_policy_data_upload_hcl/breakdown_with_policy_data_upload_hcl-upload.golden
@@ -42,12 +42,6 @@ storePolicyResources:
               "value": "bat"
             }
           ],
-          "tagPropagation": {
-            "to": "",
-            "from": null,
-            "tags": null,
-            "attribute": ""
-          },
           "values": {
             "ebs_block_device": [
               {
@@ -118,12 +112,6 @@ storePolicyResources:
               "value": "bat"
             }
           ],
-          "tagPropagation": {
-            "to": "",
-            "from": null,
-            "tags": null,
-            "attribute": ""
-          },
           "values": {
             "ebs_block_device": [
               {
--- a/cmd/infracost/testdata/breakdown_with_policy_data_upload_hcl/breakdown_with_policy_data_upload_hcl-upload.golden
+++ b/cmd/infracost/testdata/breakdown_with_policy_data_upload_hcl/breakdown_with_policy_data_upload_hcl-upload.golden
@@ -42,6 +42,12 @@ storePolicyResources:
               "value": "bat"
             }
           ],
+          "tagPropagation": {
+            "to": "",
+            "from": null,
+            "tags": null,
+            "attribute": ""
+          },
           "values": {
             "ebs_block_device": [
               {
@@ -112,6 +118,12 @@ storePolicyResources:
               "value": "bat"
             }
           ],
+          "tagPropagation": {
+            "to": "",
+            "from": null,
+            "tags": null,
+            "attribute": ""
+          },
           "values": {
             "ebs_block_device": [
               {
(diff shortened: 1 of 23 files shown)


cmd/infracost/testdata/breakdown_config_file/breakdown_config_file.golden
```diff
@@ -17,7 +17,7 @@
       "name": "infracost/infracost/cmd/infracost/testdata/breakdown_config_file/dev",
       "displayName": "testdata-breakdown_config_file-dev",
       "metadata": {
-        "path": "REPLACED_PROJECT_PATH/infracost/infracost/cmd/infracost/testdata/breakdown_config_file/dev",
+        "path": "/Users/liamg/src/infracost/cmd/infracost/testdata/breakdown_config_file/dev",
```
hugorut, line 20: hmm seems something has gone wrong here - this should have the `REPLACED_PROJECT_PATH` variable

cmd/infracost/testdata/breakdown_with_policy_data_upload_hcl/breakdown_with_policy_data_upload_hcl-upload.golden
```diff
@@ -42,6 +42,12 @@ storePolicyResources:
               "value": "bat"
             }
           ],
+          "tagPropagation": {
+            "to": "",
+            "from": null,
+            "tags": null,
+            "attribute": ""
+          },
```
hugorut, line 50: should we have empty values like this?

cmd/infracost/testdata/catches_runtime_error/catches_runtime_error.golden
```diff
@@ -4,7 +4,7 @@ ERROR An unexpected error occurred
 
 runtime error: REPLACED ERROR
 Environment:
-Infracost vREPLACED_VERSION
+Infracost
```
hugorut, line 7: Again the replaced version needs to be there

cmd/infracost/testdata/breakdown_terragrunt_with_dashboard_enabled/breakdown_terragrunt_with_dashboard_enabled.golden
```diff
@@ -1,47 +1,45 @@
 Project: dev
```
hugorut, line 1: not sure why this testfile is changed?

internal/apiclient/policy.go
```diff
@@ -135,14 +135,22 @@ type policy2Reference struct {
...
-	Tags         *[]policy2Tag            `json:"tags,omitempty"`
-	Values       json.RawMessage          `json:"values"`
-	References   []policy2Reference       `json:"references"`
-	Metadata     policy2InfracostMetadata `json:"infracostMetadata"`
-	Region       string                   `json:"region"`
+	ResourceType   string                   `json:"resourceType"`
+	ProviderName   string                   `json:"providerName"`
+	Address        string                   `json:"address"`
+	Tags           *[]policy2Tag            `json:"tags,omitempty"`
+	TagPropagation TagPropagation           `json:"tagPropagation,omitempty"`
```
hugorut, line 142: can we have this as a pointer so that it works with `omitempty`

internal/providers/terraform/parser.go
```diff
@@ -799,9 +799,33 @@ func (p *Parser) parseTags(data map[string]*schema.ResourceData, confLoader *Con
...
 		resourceData.Tags = tags
+		resourceData.TagPropagation = p.getTagPropagationInfo(resourceData)
 	}
 }
 
+func (p *Parser) getTagPropagationInfo(resource *schema.ResourceData) *schema.TagPropagation {
+	for _, expected := range aws.ExpectedPropagations {
+		if expected.ResourceType != resource.Type {
+			continue
+		}
```
hugorut, line 810: can we make the `ExpectedPropagations` a map so we can just lookup the value rather than having to loop

hugorut (commented):
mostly looks good, but theres some test failures and a few tweaks to fix

hugorut (approved):

PR #3138, opened 2024-07-16
feat: Better handling of default tags

liamg:
Collects default tags (now based on whether default tags are supported by the constrained provider version), as well as a flag indiciating default tag support.

Fixes a bug which was causing detection of support for propagation of tags to instance volumes to be unreliable.


owenrumney (approved):
LGTM

PR #3141, opened 2024-07-17
Add more config template functions

aliscott:
Adds some more config template functions.

Also updates the `parseJson` and `parseYaml` functions to not panic on errors. We don't want to stop the whole config template execution on this. It seems safer to log and not panic on these errors, since we shouldn't fail runs if we're unable to parse them. In the future if we want an option that does return error we can do what sprig does and create `mustParseYaml` and `mustParseJson` functions.


aliscott:
Test failures seem unrelated. @liamg is looking into it separately.

aliscott:
> Test failures seem unrelated. @liamg is looking into it separately.

Seemed to be a temporary TFC issue that is now resolved.

tim775 (approved):

PR #3142, opened 2024-07-17
feat: Add Google provider default labels

liamg:
Adds support for the Google providers `default_labels`. We only attempt to parse these when the provider version is constrained at `>= v5.0.0`.

Once default values are parsed, works in entirely the same way as AWS `default_tags`.

I added default labels to the `breakdown_format_json_with_tags_google` tests/golden file. I also refactored the contraints code to gather version constraints for Google, and to make it trivial to add further providers.

See https://www.hashicorp.com/blog/terraform-google-provider-adds-updates-to-default-labels for more info.

### DIFF
--- a/internal/hcl/constraints.go
+++ b/internal/hcl/constraints.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"github.com/hashicorp/go-version"
 	"github.com/zclconf/go-cty/cty/gocty"
-	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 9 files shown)


internal/hcl/constraints.go
```diff
@@ -124,47 +125,53 @@ func (p *ProviderConstraints) MarshalJSON() ([]byte, error) {
...
+			allConstraints.Google = constraints
+		}
+	}
+	return &allConstraints
+}
 
-			if body.IsNull() || !body.IsKnown() {
-				continue
-			}
+func readProviderConstraints(requiredProviderBlock *Block, provider string) (version.Constraints, error) {
```
owenrumney, line 146: Would it be worth adding a test for this?

liamg: Yes

owenrumney (approved):
LGTM

PR #3144, opened 2024-07-18
test: Fix test update path issue

liamg:
The replacement was expecting the project to be cloned in a parent directory named `infracost` e.g. `~/code/infracost/infracost`, but this is sometimes not the case.


liamg:
fixed elsewhere

PR #3145, opened 2024-07-18
fix: don't panic if the launchTemplateData map is nil

tim775:
we got a prod error with this stacktrace
   /panic.go:770                                    panic()
   internal/hcl/block.go:1385                       launchTemplateValues()
   internal/hcl/block.go:1062                       (*Block).Values()
   internal/hcl/evaluator.go:962                    (*Evaluator).evaluateResourceOrData()
...


aliscott (approved):
Do we want to lock down the version of the AFT module?

PR #3157, opened 2024-07-24
fix: detection of HCL attributes for *.tf.json files

aliscott:
There's a typo which means when the body of the parsed HCL file is a JSON body, we're not returning the attributes correctly.


aliscott:
Failing build test is fixed in https://github.com/infracost/infracost/pull/3155

hugorut (approved):

PR #3160, opened 2024-07-25
fix: adds module path to errored project

hugorut:
Adds `TerraformModulePath` to any errored project. This resolves an issue in the `diff` command. Errored projects that were discovered at a child path of the initial path provided in the config would have the same `Path` and or `Name` metadata. This would cause a `compare-to` conflict because we saw them as duplicate projects, even though these projects were actually located in different directories.


aliscott (approved):
Looks like https://github.com/infracost/infracost/pull/3140 didn't cover the case of child path projects.

PR #3172, opened 2024-08-06
feat: add spacelift remote variable support

hugorut:
Adds a remote variable loader which will fetch defined environment contexts for a stack matching the project repository and env name. In order to accomodate this I've changed the `hcl.Parser` var loader properties to a slice, this way we can define multiple remote loaders in response to a users configuration.


go.mod
```diff
@@ -277,3 +277,5 @@ replace github.com/spf13/cobra => github.com/spf13/cobra v1.4.0
 replace github.com/gruntwork-io/terragrunt => github.com/infracost/terragrunt v0.47.1-0.20240501143558-4c01e72103df
 
 replace github.com/heimdalr/dag => github.com/aliscott/dag v1.3.2-0.20231115114512-4ce18c825f94
+
+replace github.com/shurcooL/graphql => github.com/spacelift-io/graphql v1.2.0
```
hugorut, line 281: I've updated this as the spacelift client requires their internal fork, the fork adds some additional request options. There seems to be no significant divergance from the base library and considering https://github.com/shurcooL/graphql is stable I think this change is fine

schema/config.schema.json
```diff
@@ -24,7 +24,10 @@
     },
     "Project": {
       "required": [
-        "path"
+        "path",
+        "spacelift_api_key_endpoint",
```
liamg, line 28: Should these always be required?

hugorut: 👁️

hugorut: good spot, changed now

liamg (approved):
🚀

PR #3177, opened 2024-08-13
feat: Add basic Yor support for default tags

liamg:
Adds basic Yor support to pull default tags values from a yaml config/`$YOR_SIMPLE_TAGS` env var.

e.g. the following Yor config file will result in tags `A` and `B` being applied to all resources which support tags.

```yaml
tag_groups:
  - name: required_tags
    tags:
      - name: A
        value:
          default: 123456
      - name: B
        value:
          default: 654321
```

These are treated slightly differently from default tags in that they are still applied even if traditional default tags are not supported by the provider version/constraints.

### DIFF
--- a/cmd/infracost/run_test.go
+++ b/cmd/infracost/run_test.go
@@ -69,8 +69,7 @@ func TestCatchesRuntimeError(t *testing.T) {
 func TestConfigFileWithYorConfig(t *testing.T) {
 	GoldenFileCommandTest(t, testutil.CalcGoldenFileTestdataDirName(), []string{"breakdown", "--config-file", "./testdata/infracost-config-yor.yml", "--format", "json"}, &GoldenFileOptions{
 		IsJSON:      true,
-		JSONInclude: regexp.MustCompile(".*"),
-		JSONExclude: regexp.MustCompile("^path$"),
+		RegexFilter: regexp.MustCompile(`"path": "[^"]*",`),
 	})
 }
 
@@ -78,7 +77,6 @@ func TestConfigFileWithYorEnv(t *testing.T) {
 	t.Setenv("YOR_SIMPLE_TAGS", `{"Team": "Engineering"}`)
 	GoldenFileCommandTest(t, testutil.CalcGoldenFileTestdataDirName(), []string{"breakdown", "--config-file", "./testdata/infracost-config-yor.yml", "--format", "json"}, &GoldenFileOptions{
 		IsJSON:      true,
-		JSONInclude: regexp.MustCompile(".*"),
-		JSONExclude: regexp.MustCompile("^path$"),
+		RegexFilter: regexp.MustCompile(`"path": "[^"]*",`),
 	})
 }
--- a/cmd/infracost/run_test.go
+++ b/cmd/infracost/run_test.go
@@ -1,6 +1,7 @@
 package main_test
 
 import (
+	"regexp"
 	"testing"
 
 	"github.com/infracost/infracost/internal/config"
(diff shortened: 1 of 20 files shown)


internal/prices/prices.go
```diff
@@ -122,7 +122,7 @@ func (p *PriceFetcher) MissingPricesComponents() []string {
 			result = append(result, key)
 		}
 	}
-
+	sort.Strings(result)
```
liamg, line 125: This makes the output deterministic so we don't get random tests failing

internal/providers/terraform/aws/tags.go
```diff
@@ -105,16 +105,27 @@ func ParseTags(defaultTags *map[string]string, r *schema.ResourceData, config Ta
 
 	tags := make(map[string]string)
 
+	// external tags (e.g. yor)
```
tim775, line 108: is this the right precedence for yor tags? they will be "overwritten" by default tags and any explicitly defined tags?

cmd/infracost/run_test.go
```diff
@@ -64,3 +65,18 @@ func TestCatchesRuntimeError(t *testing.T) {
...
+		IsJSON:      true,
+		RegexFilter: regexp.MustCompile(`"path": "[^"]*",`),
+	})
+}
+
+func TestConfigFileWithYorEnv(t *testing.T) {
+	t.Setenv("YOR_SIMPLE_TAGS", `{"Team": "Engineering"}`)
+	GoldenFileCommandTest(t, testutil.CalcGoldenFileTestdataDirName(), []string{"breakdown", "--config-file", "./testdata/infracost-config-yor.yml", "--format", "json"}, &GoldenFileOptions{
+		IsJSON:      true,
+		RegexFilter: regexp.MustCompile(`"path": "[^"]*",`),
```
tim775, line 80: I'm curious why the 
```
JSONInclude: regexp.MustCompile("^(defaultTags|tags|name)$"),
JSONExclude: regexp.MustCompile("^(costComponents|pastBreakdown)$"),
``` 
didn't work for you? It would think it made the golden files a lost less noisy?

tim775 (approved):

cmd/infracost/run_test.go
liamg: My main problem was misunderstanding how they worked tbh, I think they would work if I listed needed properties in the include - I had tried `.*` in include and `path` in exclude which failed to exclude path because an ancestor had already been included.

internal/providers/terraform/aws/tags.go
liamg: Yeah, I think this makes sense for the first customer use-case - we effectively only fall back to Yor tags if the code doesn't already specify them.

tim775: but is that how terraform applies them? I would have guessed yor puts tags directly on resources so they would take precedence over default tags.

cmd/infracost/run_test.go
tim775: I guess it's "Show me only keys that match JSONInclude anywhere they appear in the tree, unless they appear under a key that matches JSONExclude"

internal/providers/terraform/aws/tags.go
aliscott: Yeah I think that's fair. After testing yor locally, it seems like the precedence is actually: yor tags, resource tags, default tags. Since the yor "default" tags actually overwrite any resource tags I have.

liamg: Yor overrides even regular tags - when it runs it rewrites the .tf files with the specified tag values. This is something we could do in future, but Yor gets a lot more complicated with conditionals etc., so for now we wanted to simply support taking values from the config to fall back to, regardless of conditionals etc.

But yes, it would be nice to fully support this in future, where we override defaults and tags with the Yor ones, and ideally only when conditionals match too. Yor is written in Go so we could probably just call it when a config is specified and run it "for real" once we have a Go runner.

liamg: (sorry, wrote that before your comment appeared Alistair)

liamg: Writing a PR to take Yor over the other tags now 👍

tim775: I see your point -- if we're not actually supporting yor fully it seems safer to let other tags take precedence. I'm ok if you want to keep it as is, maybe just add the above as a comment explaining why they are the lowest precedence

PR #3182, opened 2024-08-16
enhance: Make yor configs relative to the repo path

liamg:


aliscott (commented):
What's the behaviour of this if the `yor_config_path` is an absolute path?

liamg:
> What's the behaviour of this if the `yor_config_path` is an absolute path?

Hold on one sec while I add support for absolute paths ⌛

aliscott (approved):

PR #3185, opened 2024-08-28
chore: update the golden tests [IC-1572]

owenrumney:
Updating the goldens prior to a CLI release

The latest AzureRM provider (4.0.1) removes a number of resources and
changes a number.

- update tests terraform where appropriate/applicable
- ignore CLI tests where the provider change doesn't make sense to
  rework the test

### DIFF
--- a/internal/providers/terraform/azure/testdata/cognitive_account_test/cognitive_account_test.golden
+++ b/internal/providers/terraform/azure/testdata/cognitive_account_test/cognitive_account_test.golden
@@ -178,9 +178,9 @@
  └─ Text analytics for health (5K-500K)                               Monthly cost depends on usage: $25.00 per 1K records         
                                                                                                                                    
  azurerm_cognitive_account.with_usage["SpeechServices-S0"]                                                                         
- ├─ Speech to text                                                                    20  hours                           $20.00   
+ ├─ Speech to text                                                                    20  hours                        not found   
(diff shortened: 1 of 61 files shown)


internal/providers/terraform/azure/testdata/cognitive_account_test/cognitive_account_test.golden
```diff
@@ -178,9 +178,9 @@
  └─ Text analytics for health (5K-500K)                               Monthly cost depends on usage: $25.00 per 1K records         
                                                                                                                                    
  azurerm_cognitive_account.with_usage["SpeechServices-S0"]                                                                         
- ├─ Speech to text                                                                    20  hours                           $20.00   
+ ├─ Speech to text                                                                    20  hours                        not found
```
tim775, line 181: We should create a follow ticket to investigate why these are "not found" if you haven't already done so, just so we don't lose track of it.

tim775 (approved):

internal/providers/terraform/azure/testdata/cognitive_account_test/cognitive_account_test.golden
owenrumney: added to this ticket - https://www.notion.so/infracost/Different-Plans-5fd3a5e4fddc414690cdb7944d07001e?pvs=4

Thanks for review

PR #3186, opened 2024-08-29
fix: speed up parent Terragrunt config detection

hugorut:
This PR changes the parent Terragrunt config detection to parse the list of detected config files into a Tree before trying to determine which Terragrunt files are parent configuration. This significantly speeds up the time to detect parent configs as we need many less io operations to determine which files are parents. On a large repo with ~6k Terragrunt config files the speed of infracost generate went from ~50s to ~1.5s after this change.


owenrumney (approved):
LGTM - I've read through the tree logic separately and this makes sense as far as I can see

PR #3196, opened 2024-09-04
Terragrunt speed

aliscott:
This implements a two changes for improving the speed of Terragrunt repos which reference sources within large monorepos.

1. Enables the partial parse config cache (https://terragrunt.gruntwork.io/docs/reference/cli-options/#terragrunt-use-partial-parse-config-cache)
2. Adds an option to use `git sparse-checkout` for downloading only specific subdirectories of remote modules. This also handles special cases for local references to other subdirectories of that module, or symlinks within that module.

### DIFF
--- a/internal/hcl/modules/loader.go
+++ b/internal/hcl/modules/loader.go
@@ -425,6 +425,16 @@ func (m *ModuleLoader) loadRemoteModule(key string, source string) (*ManifestMod
 		return nil, err
 	}
 
+	// Parse the URL, add the subdir as a param
+	u, err := url.Parse(moduleAddr)
+	if err != nil {
+		return nil, err
+	}
+	q := u.Query()
+	q.Set("subdir", submodulePath)
+	u.RawQuery = q.Encode()
+	moduleAddr = u.String()
+
 	dest := m.downloadDest(moduleAddr, "")
 	moduleDownloadDir, err := m.cachePathRel(dest)
 	if err != nil {
--- a/internal/hcl/modules/loader.go
+++ b/internal/hcl/modules/loader.go
@@ -1,11 +1,13 @@
 package modules
 
 import (
+	"bytes"
 	"crypto/md5" //nolint
 	"errors"
 	"fmt"
 	"net/url"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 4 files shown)


internal/hcl/modules/loader.go
```diff
@@ -315,6 +346,199 @@ func (m *ModuleLoader) loadModule(moduleCall *tfconfig.ModuleCall, parentPath st
 	return manifestModule, nil
 }
 
+// checkoutPathIfRequired checks if the given directories are in the sparse-checkout file list and adds them if not.
+func (m *ModuleLoader) checkoutPathIfRequired(repoRoot string, dirs string) error {
+	// Lock the git repo root so we don't have multiple calls trying to read and update the sparse-checkout file list.
+	unlock := m.sync.Lock(repoRoot)
```
aliscott, line 352: I think it's okay to use the `m.sync` lock instead of creating a different one, since the `repoRoot` shouldn't overlap with any of the remote module keys.

internal/hcl/modules/loader.go
```diff
@@ -279,6 +287,29 @@ func (m *ModuleLoader) loadModule(moduleCall *tfconfig.ModuleCall, parentPath st
 		}
 
 		m.logger.Debug().Msgf("loading local module %s from %s", key, dir)
+
+		// If the module is a in a git repo we need to check that it has been checked out.
+		// If it hasn't been checked out then we need to add it to the sparse-checkout file list.
+		m.logger.Debug().Msgf("finding git repo root for path %s", parentPath)
```
hugorut, line 293: can we change this to Trace

internal/hcl/modules/loader.go
```diff
@@ -315,6 +346,199 @@ func (m *ModuleLoader) loadModule(moduleCall *tfconfig.ModuleCall, parentPath st
...
 }
 
+// checkoutPathIfRequired checks if the given directories are in the sparse-checkout file list and adds them if not.
+func (m *ModuleLoader) checkoutPathIfRequired(repoRoot string, dirs string) error {
+	// Lock the git repo root so we don't have multiple calls trying to read and update the sparse-checkout file list.
+	unlock := m.sync.Lock(repoRoot)
+	defer unlock()
+
+	// Get the list of sparse checkout directories (and assume sparse checkout is enabled if this succeeds)
+	m.logger.Debug().Msgf("getting sparse checkout directories for path %s", repoRoot)
```
hugorut, line 356: again - maybe Trace is better

internal/hcl/modules/loader.go
```diff
@@ -315,6 +346,199 @@ func (m *ModuleLoader) loadModule(moduleCall *tfconfig.ModuleCall, parentPath st
...
+
+		for _, symlinkedDir := range symlinkedDirs {
+			if !isCoveredByExistingDirs(existingDirs, symlinkedDir) {
+				additionalDirs = append(additionalDirs, symlinkedDir)
+			}
+		}
+	}
+
+	if len(additionalDirs) > 0 {
+		logger.Debug().Msgf("recursively adding symlinked dirs to sparse-checkout for repo %s: %v", repoRoot, additionalDirs)
```
hugorut, line 426: Trace

hugorut (approved):

PR #3204, opened 2024-09-12
fix: save posted comment when cloud is enabled

tim775:
* make sure the AddRunResponse is returned after uploading the run
* upload the pr comment when cloud upload is enabled (either explicitly or implicitly)

### DIFF
--- a/cmd/infracost/comment_github.go
+++ b/cmd/infracost/comment_github.go
@@ -134,7 +134,7 @@ func commentGitHubCmd(ctx *config.RunContext) *cobra.Command {
 					return err
 				}
 
-				if res.Posted && ctx.IsCloudUploadExplicitlyEnabled() {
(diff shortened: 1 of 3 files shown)


cmd/infracost/comment_github.go
```diff
@@ -134,7 +134,7 @@ func commentGitHubCmd(ctx *config.RunContext) *cobra.Command {
 					return err
 				}
 
-				if res.Posted && ctx.IsCloudUploadExplicitlyEnabled() {
+				if res.Posted && ctx.IsCloudUploadEnabled() {
```
owenrumney, line 137: This was me that did this so apoligies - I can see the code difference between thes, but what is the realterms difference in usage?

owenrumney (approved):

PR #3206, opened 2024-09-18
fix: correct reference lookups of count/each resources in count/each modules

tim775:
This fixes an issue where references for count or for each resources within a count or foreach module where not correctly matched because the reference lookup function incorrectly used the module index as the resource index. 

For example, a resource referencing `module.mymod[0].aws_resource.my_resource[1]` would mistakenly be mapped to `module.mymod[0].aws_resource.my_resource[0]`.


aliscott (approved):
Nice find!

PR #3219, opened 2024-10-10
Fix Terragrunt dep output mocks

aliscott:
The updates the pre-processing of Terragrunt dependencies to take into account instances where a function is called on the dependency. For example, previous we were looking for patterns like `dependency.my-mod.my-output[0]` to be able to tell if the output should be mocked as a list. Now we also look for patterns like `tolist(dependency.my-mod.my-output)[0]`.

Eventually we will want to replace the internal Terragrunt HCL decoding to use our own so that we handle mocks better, but this unblocks us for just now.

The tests were also failing because a remote module has been updated, so I had to fix a bug with the sparse checkouts to be able to update them.

### DIFF
--- a/internal/hcl/modules/loader.go
+++ b/internal/hcl/modules/loader.go
@@ -288,7 +288,7 @@ func (m *ModuleLoader) loadModule(moduleCall *tfconfig.ModuleCall, parentPath st
 
 		m.logger.Debug().Msgf("loading local module %s from %s", key, dir)
 
-		// If the module is a in a git repo we need to check that it has been checked out.
+		// If the module is in a git repo we need to check that it has been checked out.
 		// If it hasn't been checked out then we need to add it to the sparse-checkout file list.
 		m.logger.Trace().Msgf("finding git repo root for path %s", parentPath)
 		repoRoot, err := findGitRepoRoot(parentPath)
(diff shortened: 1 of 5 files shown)


internal/hcl/modules/loader.go
```diff
@@ -452,6 +463,22 @@ func findGitRepoRoot(startPath string) (string, error) {
...
+// isSparseCheckoutEnabled checks if sparse-checkout is enabled in the repository
+func isSparseCheckoutEnabled(repoRoot string) (bool, error) {
+	cmd := exec.Command("git", "config", "--get", "core.sparseCheckout")
+	cmd.Dir = repoRoot
+	output, err := cmd.Output()
+	if err != nil {
+		// if exit status is 1, then the config is not set
+		if exitErr, ok := err.(*exec.ExitError); ok && exitErr.ExitCode() == 1 {
+			return false, nil
+		}
```
liamg, line 475: Nitpick, feel free to ignore :D
```suggestion
        var exitErr *exec.ExitError
        if errors.As(err, &exitErr) && exitErr.ExitCode() == 1 {
            return false, nil
        }
```

liamg (approved):
LGTM

Issue #3228, opened 2024-10-17
AWS VPC Endpoint pricing tiers are wrong


jamescnowell:
The current configuration has tiers set at 1000GB and 4000GB:
https://github.com/infracost/infracost/blob/38e888cfc56a42fcc2603ea4eee65350482b8202/internal/resources/aws/vpc_endpoint.go#L70

The correct configuration should be 1000000GB (1PB) and 5000000GB (5PB) as documented here: https://aws.amazon.com/privatelink/pricing/

jamescnowell:
I am working on a fix for this now.

aliscott:
@jamescnowell thanks 🙏  this has been released in [v0.10.40](https://github.com/infracost/infracost/releases/tag/v0.10.40)

PR #3229, opened 2024-10-17
fix: AWS VPC Endpoint pricing tiers are wrong #3228

jamescnowell:
The previous configuration had 1TB and 5TB tiers instead of 1PB and 5PB.

This fix correctly computes costs at all traffic levels rather than improperly deflating costs.

See https://github.com/infracost/infracost/issues/3228


aliscott (approved):
Thanks @jamescnowell 🙏

jamescnowell:
Should I attempt to fix the failing tests? They appear to be unrelated to this change, and I'm not sure what exactly to do with them.

PR #3237, opened 2024-10-31
fix: fix Spacelift var loading

aliscott:
1. Update the Spacelift remote variable loader to load in variables from all 3 of the locations:
    - Environment
    - Spacelift environment
    - Attached contexts.

    We should look for any key starting with `TF_VAR_`

2. Update the Spacelift remote variable loader to better match on stack name. I think the following logic would work:
    1. Match the repository name exactly
    2. Match the projectRoot exactly based on the path of the module
    3. Query for multiple results
    4. If the project has a `moduleSuffix` then match on that using what is after the `:` in the project name. If there’s no `moduleSuffix`, then skip this step.
    5. Log a warning if there’s more than one match or no matches

### DIFF
--- a/internal/hcl/parser.go
+++ b/internal/hcl/parser.go
@@ -275,6 +275,14 @@ func OptionGraphEvaluator() Option {
 	}
 }
 
+// OptionWithProjectName sets the project name for the parser.
+// This is used if the project name has been explicitly set by the user or the autodetection
+func OptionWithProjectName(name string) Option {
+	return func(p *Parser) {
+		p.projectName = name
+	}
+}
+
 type DetectedProject interface {
 	ProjectName() string
 	EnvName() string
@@ -288,6 +296,7 @@ type DetectedProject interface {
 type Parser struct {
 	startingPath          string
 	detectedProjectPath   string
+	projectName           string
 	tfEnvVars             map[string]cty.Value
 	tfvarsPaths           []string
 	inputVars             map[string]cty.Value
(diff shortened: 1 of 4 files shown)


internal/providers/terraform/hcl_provider.go
```diff
@@ -186,14 +192,6 @@ func NewHCLProvider(ctx *config.ProjectContext, rootPath hcl.RootPath, config *H
...
 
 func (p *HCLProvider) ProjectName() string {
-	if p.ctx.ProjectConfig.Name != "" {
-		return p.ctx.ProjectConfig.Name
-	}
-
-	if p.ctx.ProjectConfig.TerraformWorkspace != "" {
-		return p.Parser.ProjectName() + "-" + p.ctx.ProjectConfig.TerraformWorkspace
-	}
-
```
aliscott, line 196: This doesn't seem to be needed anymore since I now pass in the config's project name to the parser. So this can now just be consistent with the parser's logic.

internal/hcl/parser.go
```diff
@@ -568,7 +583,7 @@ func (p *Parser) EnvName() string {
 		return p.moduleSuffix
 	}
 
-	return p.ProjectName()
+	return p.envMatcher.EnvName(p.ProjectName())
```
aliscott, line 586: Not sure if there's a better way of doing this, but this seems to give me the results I want since it matches against the env names in the config file.

internal/hcl/parser.go
```diff
@@ -553,10 +562,16 @@ func (p *Parser) RelativePath() string {
...
 func (p *Parser) ProjectName() string {
+	if p.projectName != "" {
+		return p.projectName
+	}
+
 	name := config.CleanProjectName(p.RelativePath())
 
 	if p.moduleSuffix != "" {
 		name = fmt.Sprintf("%s-%s", name, p.moduleSuffix)
+	} else if p.workspaceName != "" && p.workspaceName != defaultTerraformWorkspaceName {
```
aliscott, line 573: Added this so it could be more consistent with how `hcl_provider.ProjectName` worked.

hugorut (approved):

PR #3243, opened 2024-11-12
fix: Fix config generator when env name is in the project path IC-1911

vdmgolub:
When an env name is in the path already, we should ignore the other envs.

### DIFF
--- a/internal/providers/detect.go
+++ b/internal/providers/detect.go
@@ -160,6 +160,7 @@ func autodetectedRootToProviders(projectContext *config.ProjectContext, rootPath
 					options,
 					hcl.OptionWithTFVarsPaths(append(autoVarFiles.ToPaths(), env.TerraformVarFiles.ToPaths()...), true),
 					hcl.OptionWithModuleSuffix(env.Name),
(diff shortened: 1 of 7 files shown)


aliscott:
@hugorut is probably better reviewing this.

internal/providers/detect.go
```diff
@@ -160,6 +160,7 @@ func autodetectedRootToProviders(projectContext *config.ProjectContext, rootPath
 					options,
 					hcl.OptionWithTFVarsPaths(append(autoVarFiles.ToPaths(), env.TerraformVarFiles.ToPaths()...), true),
 					hcl.OptionWithModuleSuffix(env.Name),
+					hcl.OptionWithProjectEnv(rootPath.DetectedPath, env.Name),
```
vdmgolub, line 163: I should probably just include this option into `OptionWithModuleSuffix` as the order matters now.

hugorut: agreed, this logic should go into the module suffix option

hugorut (approved):
Looks good to me

PR #3248, opened 2024-11-14
S3 module cache updates

aliscott:
1. Add a lower TTL for branch references. If the reference looks like a tag or commit SHA then we can use a higher TTL since those are unlikely to change.
  - In the future, we could hit the GitHub API for the latest commit SHA of the branch and store that in the cache as well to make this better.

2. Update the sparse-checkout functionality to work with the remote cache. I originally tried moving a lot of the sparse checkout logic into go-getter but that got complicated with the cache, sparse-checkout list and symlinks, so I've kept it in the loader just now.

3. Update the Terragrunt source load to work with the remote cache. This requires https://github.com/infracost/terragrunt/pull/18.


internal/providers/terraform/terragrunt_hcl_provider.go
```diff
@@ -643,24 +664,26 @@ func splitModuleSubDir(moduleSource string) (string, string, error) {
 	return moduleAddr, submodulePath, nil
 }
 
-// downloadSourceOnce thread-safely makes sure the sourceURL is only downloaded once
-func downloadSourceOnce(sourceURL string, opts *tgoptions.TerragruntOptions, terragruntConfig *tgconfig.TerragruntConfig, logger zerolog.Logger) (string, error) {
+// loadSourceOnce thread-safely makes sure the sourceURL is only downloaded once.
```
aliscott, line 667: I renamed this to avoid confusion with the `downloadSource` function.

internal/hcl/modules/fetch.go
```diff
@@ -136,20 +183,13 @@ func (p *PackageFetcher) fetchFromRemote(moduleAddr, dest string) (bool, error)
...
-	}
-
-	getters["git"] = &CustomGitGetter{new(getter.GitGetter)}
-
 	client := getter.Client{
 		Src:           moduleAddr,
 		Dst:           dest,
 		Pwd:           dest,
-		Mode:          getter.ClientModeDir,
+		Mode:          getter.ClientModeAny,
```
aliscott, line 190: Terragrunt needs this and it seems to work okay. Go-getter still detects archives as directories.

internal/hcl/modules/s3_cache.go
```diff
@@ -55,19 +56,32 @@ func (cache *S3Cache) applyPrefix(key string) string {
...
-	_, err := cache.s3Client.HeadObject(&s3.HeadObjectInput{
+	headObj, err := cache.s3Client.HeadObject(&s3.HeadObjectInput{
 		Bucket: aws.String(cache.bucketName),
 		Key:    aws.String(prefixedKey),
 	})
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok {
 			switch awsErr.Code() {
-			case s3.ErrCodeNoSuchKey, "NotFound":
+			case s3.ErrCodeNoSuchKey, "NotFound", "Forbidden":
```
owenrumney, line 66: Should we add a case for forbidden and log it out. Seems more of a failure case than just it not being in the cache?

owenrumney (approved):
LRGTM!

internal/hcl/modules/s3_cache.go
aliscott: I thought about that, but I think this will just spam the logs. If you don't have `ListBucket` permission which is what we do then every case of NotFound shows as Forbidden so there's no way of differentiating.

PR #3249, opened 2024-11-15
enhance: Add curent and past errors to project metadata

tim775:
The project metadata previously contained a single errors field containing any errors that occurred in either side of a diff. This change makes it possible to determine which side of the diff the errors came from.


liamg (approved):
LGTM

I added these fields to the runner and rolled it out so it doesn't freak out about new fields: https://github.com/infracost/runner/pull/109

tim775:
> LGTM
> 
> I added these fields to the runner and rolled it out so it doesn't freak out about new fields: [infracost/runner#109](https://github.com/infracost/runner/pull/109)

Yeah I had sent https://github.com/infracost/runner/pull/108

PR #3253, opened 2024-11-19
feat: Add option to create azure comments as active instead of defaulting to closed

liamg:
image


cmd/infracost/comment_azure_repos.go
```diff
@@ -41,9 +41,11 @@ func commentAzureReposCmd(ctx *config.RunContext) *cobra.Command {
 
 			token, _ := cmd.Flags().GetString("azure-access-token")
 			tag, _ := cmd.Flags().GetString("tag")
+			initActive, _ := cmd.Flags().GetBool("init-active")
```
vdmgolub, line 44: What do you think if we provide `--comment-status=closed|active` flag (with a default of 'closed')? Using such a flag we can expand the flag with more statuses in the future compared to introducing a new flag or changing the existing one.

vdmgolub (approved):

cmd/infracost/comment_azure_repos.go
liamg: I'd like to involve YAGNI in this case - I like the idea but if we need it we can switch to this approach later - saves having to validate the passed in value etc. 👍

vdmgolub: Changing a public flag may result in broken behaviors for users. I wouldn't use YAGNI for user facing features and think in advance.

liamg: IMHO it should apply to user-facing features - it stops us increasing complexity by adding features we're not going to use.

vdmgolub: I agree with that, but "init-active" looks more like a one-off fix than a feature flag. Maybe renaming it to `keep-active` or something like will do the trick. With behavior `update` "init" doesn't make sense anymore as the comment can already exist and the flag will override it.

liamg: Yeah, I see what you mean. I was meaning initialise the comment as active, i.e. don't change it on `update`. Even `keep-active` is confusing because if the user changed it to `wontfix`, it's not clear what the CLI will do. I'll have a think 🤔

vdmgolub: Ah, so we don't want to change the status of existing comments. Then the name makes total sense 👍

liamg: Yeah, that's right, the theory being the user may mark the comment as `wontfix` or whatever and we don't want to mess with the status once they've done so - so we only set it to `active` when the comment is first made 👍

PR #3267, opened 2024-12-06
perf: Performance optimisations + metrics

liamg:
- preallocates backing arrays for slices (where we know the max size)
- switches to adding graph edges individually, saving on memory and some redundant checks
- removes some unnecessary allocations
- swaps out a map+mutex combination to use sync.Map instead
- replaces a .AttributesAsMap()[attrName] with .GetAttribute(attrName)
- defines and uses a lightweight TopologicalWalk function which walks the graph in a more efficent way
- adds performance metrics (must be set in config to do this) which are written to a file for - analysis in the runner/api
- fixes a small bug with content of events we're recording
- fixed a bug with conflicting module downloads causing git process to hang/fail
- fixed module cache usage

### DIFF
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -296,7 +296,7 @@ func (r *parallelRunner) run() ([]projectResult, error) {
 		projectCounts[provider.DisplayType()] = 1
 	}
 
-	var order []string
+	order := make([]string, 0, len(projectCounts))
 	for displayType := range projectCounts {
 		order = append(order, displayType)
 	}
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -114,9 +114,7 @@ func runMain(cmd *cobra.Command, runCtx *config.RunContext) error {
 
 	for _, projectResult := range projectResults {
 		projectContexts = append(projectContexts, projectResult.ctx)
-		for _, project := range projectResult.projectOut.projects {
-			projects = append(projects, project)
-		}
+		projects = append(projects, projectResult.projectOut.projects...)
 	}
 
 	r, err := output.ToOutputFormat(runCtx.Config, projects)
--- a/cmd/infracost/run.go
+++ b/cmd/infracost/run.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/Rhymond/go-money"
+	"github.com/infracost/infracost/internal/metrics"
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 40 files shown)


internal/hcl/graph.go
```diff
@@ -385,16 +373,49 @@ func (g *Graph) AsJSON() ([]byte, error) {
...
 func (g *Graph) Walk() {
 	v := NewGraphVisitor(g.logger, g.vertexMutex)
+	TopologicalWalk(g.dag, v.Visit)
+}
 
-	flowCallback := func(d *dag.DAG, id string, parentResults []dag.FlowResult) (interface{}, error) {
-		vertex, _ := d.GetVertex(id)
-
-		v.Visit(id, vertex)
+func TopologicalWalk(graph *dag.DAG, visitor func(id string, vertex Vertex)) {
```
hugorut, line 379: I think the only thing that might be missing from this is a cycles check. This is built into the DescendantsFlow function. However, the last time I looked at the benchmarks the cycles check took a ton of mem to implement, so it might be best to rip it out, as you've done here. Just noting here for the sake of audit at a later date.

internal/hcl/graph_vertex_module_call.go
```diff
@@ -78,8 +78,6 @@ func (v *VertexModuleCall) Visit(mutex *sync.Mutex) error {
 }
 
 func (v *VertexModuleCall) evaluate(e *Evaluator, b *Block, mutex *sync.Mutex) error {
-	mutex.Lock()
```
hugorut, line 81: Is the removal of these locks now because the visits are not concurrent? Or is there another lock I'm missing somewhere.

hugorut (approved):
looks good, my main concern is around the change of the graph walk func, we need to verify on a few other customers that this doesn't cause disrepancies in price match/resources before releasing.

cmd/infracost/run.go
```diff
@@ -375,37 +384,52 @@ func (r *parallelRunner) run() ([]projectResult, error) {
...
-				var configProjects *projectOutput
-				ctx := job.ctx
-				if job.err != nil {
-					configProjects = newErroredProject(job.provider, job.ctx, job.err)
-				} else {
-					configProjects, err = r.runProvider(job)
-					ctx = job.provider.Context()
-					if err != nil {
-						configProjects = newErroredProject(job.provider, ctx, err)
+				func() {
```
vdmgolub, line 387: Could you please educate me why the func is needed here? Is it because you need to defer the timer stopping?

internal/hcl/graph_vertex_module_call.go
liamg: Yeah exactly, they're sequential at the moment. We could parallelise the walk of the graph in future, but I didn't go that far for now.

vdmgolub (approved):

liamg:
> looks good, my main concern is around the change of the graph walk func, we need to verify on a few other customers that this doesn't cause disrepancies in price match/resources before releasing.

Yep, doing some comparisons with some of the bigger customers to ensure matches before release 👍

PR #3276, opened 2024-12-17
fix: handle S3 bucket module ID issues

owenrumney:
Resolves an issue where a known module (https://github.com/terraform-aws-modules/terraform-aws-s3-bucket) returns a coalesced ID starting with the id of the bucket policy.

This does not work with our generated IDs and can result in cases where
a lifecycle configuration that is referencing the `s3_bucket_id` output
on the module generated bucket.

This fix is specifically very targetted to
`aws_s3_bucket_lifecycle_configuration`, we should really only add the
rewrites when we know there is an issue


hugorut (approved):
Looks good

PR #3281, opened 2024-12-18
fix: Workaround go-getter ctx issue and set timeout on git-getter directly

liamg:


internal/hcl/modules/fetch.go
```diff
@@ -66,6 +65,10 @@ func NewPackageFetcher(remoteCache RemoteCache, logger zerolog.Logger, opts ...P
 		opt(p)
 	}
 
+	if p.timeout > 0 {
+		getters["git"].(*CustomGitGetter).GitGetter.Timeout = p.timeout
```
hugorut, line 69: Any chance we can change it at the point of init: https://github.com/infracost/infracost/blob/master/internal/hcl/modules/fetch.go#L56

e.g. 

```go
 &CustomGitGetter{getter.GitGetter{Timeout: p.timeout} }
 ```

liamg: I didn't do that as there's an option that can modify it (`WithTimeout` or something), but lets simplify and remove the option as nothing uses it 👍

hugorut (approved):

PR #3283, opened 2024-12-18
fix: don't panic when truncating to negative lengths

tim775:
Not sure how it's happening, but somehow the truncate calculation in markdown.go is passing in passing in negative maxLen.

### DIFF
--- a/internal/output/format.go
+++ b/internal/output/format.go
@@ -124,6 +124,10 @@ func formatTitleWithCurrency(title, currency string) string {
 }
 
 func truncateMiddle(s string, maxLen int, fill string) string {
+	if maxLen <= 0 {
+		return ""
+	}
+
 	r := []rune(s)
 
 	if len(r) <= maxLen {


liamg (commented):
I'm concerned we're going to see empty messages - it looks like it's probably here:

https://github.com/infracost/infracost/blob/master/internal/output/markdown.go#L337-L341

I can't figure out exactly what that's doing, but it seems the only place a negative number could be passed in.

liamg:
I suppose it's when the size of the entire message is much bigger than just the diff message?

tim775:
> I suppose it's when the size of the entire message is much bigger than just the diff message?

yeah that's got to be the spot which is why i was thinking truncating the whole diff would be ok, but you're right if it's the rest of the message that's the problem then we'd just end up in a recursion loop. I think the truncateMiddle fix is ok but i'm going to look at fixing that call too.

liamg:
Okay, will approve for now then, up to you whether you fix elsewhere, I'm lacking context here 😅

liamg (approved):

aliscott (approved):

PR #3284, opened 2024-12-19
feat(aws): add aws_cloudfront_function support

### LINKED ISSUE #3268: Add support for Cloudfront Function
Hi, I'm planning to use this infracost but I see that infracost does not support Cloudfront Function price calculation yet. Can it be added in the future?

lewisloofis:
## Objective:

Add support for AWS CloudFront Function. Fixes #3268.

## Pricing details:

I am using https://aws.amazon.com/cloudfront/pricing/ as the source of my calculation. To calculate the price, users have to sum their cloudfront distribution requests, and multiply it with existing AWS prices which is $0.1 per 1M invocations / request. The difficult thing is that this resource is not on AWS Pricing calculator yet.

## Status:

- [ ] Added test cases without usage-file
- [x] Created a PR to update "Supported Resources" in the [docs](https://github.com/infracost/docs/pull/683)

## Issues:

Fix #3268

### DIFF
--- a/internal/providers/terraform/aws/cloudfront_function.go
+++ b/internal/providers/terraform/aws/cloudfront_function.go
@@ -1,0 +1,21 @@
+package aws
+
+import (
+	"github.com/infracost/infracost/internal/resources/aws"
+	"github.com/infracost/infracost/internal/schema"
+)
+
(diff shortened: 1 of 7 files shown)


aliscott (approved):
Thanks @lewisloofis 🙏

PR #3292, opened 2024-12-30
Go 1.23

aliscott:
Updates Go to 1.23 and fixes any failing tests

### DIFF
--- a/internal/hcl/modules/fetch.go
+++ b/internal/hcl/modules/fetch.go
@@ -95,7 +95,8 @@ func (p *PackageFetcher) Fetch(moduleAddr string, dest string) error {
 	}
 
 	if err != nil {
-		p.logger.Warn().Msgf("error fetching module %s from local cache: %s", util.RedactUrl(moduleAddr), err)
+		// Log the error at debug level because the local cache might be invalid and in that case we just refetch it so it's not a big deal
+		p.logger.Debug().Msgf("error fetching module %s from local cache: %s", util.RedactUrl(moduleAddr), err)
 	}
 
 	fetched, err = p.fetchFromRemoteCache(moduleAddr, dest)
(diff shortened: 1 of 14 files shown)


internal/hcl/modules/fetch.go
```diff
@@ -95,7 +95,8 @@ func (p *PackageFetcher) Fetch(moduleAddr string, dest string) error {
 	}
 
 	if err != nil {
-		p.logger.Warn().Msgf("error fetching module %s from local cache: %s", util.RedactUrl(moduleAddr), err)
```
aliscott, line 98: Looks like the`TestAutoscalingGroup` test had started failing with the below:
```
+WARN error fetching module git::https://github.com/terraform-aws-modules/terraform-aws-autoscaling?ref=72561e6e0062a3a57b13fc759d358012ecc9bef8 from local cache: failed to create directory '/var/folders/q9/fpmzkgv55tgcngkn1m0w44740000gn/T/TestAutoscalingGroupHCL_Graph2324842685/001/.infracost/terraform_modules/8d5b2db2ae5ad88e49b37cee1c01e875': mkdir /var/folders/q9/fpmzkgv55tgcngkn1m0w44740000gn/T/TestAutoscalingGroupHCL_Graph2324842685/001/.infracost/terraform_modules/8d5b2db2ae5ad88e49b37cee1c01e875: no such file or directory project_name= project_path=/var/folders/q9/fpmzkgv55tgcngkn1m0w44740000gn/T/TestAutoscalingGroupHCL_Graph2324842685/001 provider=terraform_dir
```

From what I can tell this started happening after https://github.com/infracost/infracost/pull/3267. I couldn't work out the exact cause, but I also don't think it's an issue since if we fail to load from the local cache we just retry. @liamg might know more why this was happening.

tim775 (approved):

PR #3295, opened 2024-12-31
fix: vars causing unknown keys

tim775:
This adds test cases and an idea I had for fixing for tag policy errors not triggering when they should (false negative). Unfortunately the fix doesn't work and the test cases don't do what I expect.

@liamg Could you take a look at the tess and give some help on sorting out in what cases we should see the `missingVarsCausingUnknownDefaultTagKeys` flag set on resources?

### DIFF
--- a/cmd/infracost/breakdown_test.go
+++ b/cmd/infracost/breakdown_test.go
@@ -214,6 +214,29 @@ func TestBreakdownFormatJsonPropagateDefaultsToVolumeTags(t *testing.T) {
 	)
 }
 
+func TestBreakdownFormatJsonMissingDefaultTags(t *testing.T) {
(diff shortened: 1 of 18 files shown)


liamg:
@tim775 I've taken a look at this and added a couple of extra test projects.

`missingVarsCausingUnknownDefaultTagKeys` is set when an object is defined but may have unknown keys. For example, it should trigger for `tags = merge({ Name = "something"}, var.whatever)` when `var.whatever` is undefined, but not for `tags = var.whatever`. It originally came from IC-1541, I think as we were seeing false positives when `tags` was partially defined but had some missing keys, probably due to a `merge` parameter being undefined.

PR #3308, opened 2025-01-07
fix: Handle slices of Tags

owenrumney:
Some AWS resources have the tags as a slice of objects rather than
string map, so we need to be able to convert that list into a string
map.

This change adds support for that kind of Tags

eg;

```
tags = [
    {
      key = "Name"
      value = "Name"
      propagate_at_launch = true

    }
  ]
```


internal/providers/terraform/aws/tags.go
```diff
@@ -103,7 +103,28 @@ func ParseTags(externalTags, defaultTags map[string]string, r *schema.ResourceDa
...
+	switch t := rTagBlock.Value().(type) {
+	case []interface{}:
+		rTags = make(map[string]gjson.Result, len(t))
+		for _, el := range t {
+			tag, ok := el.(map[string]interface{})
+			if !ok {
+				// This should never happen, but if it does, we should skip it
+				continue
+			}
+			k := tag["key"].(string)
```
liamg, line 120: might be worth doing `k, ok := tag["key"]` and then `ks, ok := k.(string)` on these just in case somebody misses the key or value attribute fro mthe terraform, or sets it to a number or something

PR #3320, opened 2025-01-28
fix: treat underlying Terraform errors as Terragrunt project errors

hugorut:
Prior to this change, when evaluating a Terragrunt project, we treated the underlying Terraform module errors as just a warning diagnostic. This created issues with the diff logic as baseline Terragrunt projects would have warning metadatas but a zero cost as the project could not be evaluated. If this was an intermittent error, which is often the case with Terragrunt and the caching logic then users would see a 100% cost increase for the project as it was seen as going from 0 to the current project price.

I've changed the logic now to treat Terraform errors as first class Terragrunt evaluation errors in the evaluator. This means that when diffing we'll show that we couldn't compute the price instead of a new project addition.

### DIFF
--- a/internal/providers/terraform/terragrunt_hcl_provider.go
+++ b/internal/providers/terraform/terragrunt_hcl_provider.go
@@ -32,7 +32,6 @@ import (
 	hcl2 "github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
 	"github.com/hashicorp/hcl/v2/hclparse"
-	"github.com/infracost/infracost/internal/metrics"
 	"github.com/rs/zerolog"
 	"github.com/sirupsen/logrus"
 	"github.com/zclconf/go-cty/cty"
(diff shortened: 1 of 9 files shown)


internal/providers/terraform/terragrunt_hcl_provider.go
```diff
@@ -678,15 +677,7 @@ func (p *TerragruntHCLProvider) runTerragrunt(opts *tgoptions.TerragruntOptions)
...
 
 	mod := h.Module()
-	if mod.Error != nil {
-		path := ""
-		if mod.Module != nil {
-			path = mod.Module.RootPath
-		}
-		info.addWarning(schema.NewDiagTerragruntModuleEvaluationFailure(mod.Error))
-		p.logger.Warn().Msgf("Terragrunt config path %s returned module %s with error: %s", opts.TerragruntConfigPath, path, mod.Error)
-	}
```
hugorut, line 688: I'm not sure exactly why we treated these errors as warnings previously, I can't find anything. However my assumption is that this logic worked previously in Infracost cloud but we at some point changed the behaviour to be less stringent with "warnings". 

I think treating these errors as proper "fail" errors is a safe and correct change, I don't believe these should be warnings.

aliscott: Agree, not sure why this was the case. These should be errors.

aliscott (approved):

hugorut:
I'm going to merge as the failures aren't related to this change and are already in master

PR #3322, opened 2025-01-29
fix: Add local modules to manifest if they are source mapped

vdmgolub:
When a module's source has a relative path like `../../modules/` but needs to be mapped to another path, the module is still loaded using the original source value.
Important: this mapping only works when the path starts with the mapping value.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/infracost/infracost
 
-go 1.23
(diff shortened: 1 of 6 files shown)


hugorut (approved):

hugorut:
Would be good if we could add a test for this, but don't worry if too much

PR #3348, opened 2025-03-06
fix: fix the wrong error return value

tiaoxizhan:
Since we have already checked err before and returned when err != nil, err must be nil here. 

So we should  reassign err.


aliscott (approved):
Thanks @tiaoxizhan 🙏

PR #3355, opened 2025-03-17
feat: add isProduction template function

hugorut:
adds a template function to check if a project is marked as a production project from the configured settings.


hugorut:
test failures do not look related

aliscott (approved):

PR #3395, opened 2025-05-23
Fix Azure cost issue within elastic pool

### LINKED ISSUE #3375: Cost for Azure SQL DB within elastic pool calculated incorrectly
We have 4 databases within an elastic pool, hence the compute cost is billed once per elastic pool (not per single DB). If we now add a new DB to the elastic pool, infracost considers the DB to be billed as single DB and shows a cost increase for compute that is incorrect.

Image

mohsenkhosroanjam:
This PR tries to fix #3375


aliscott:
Closing since we need more information to reproduce this issue.

PR #3398, opened 2025-05-28
fix: add trailing registry URL slash

drewblas:
Typically, the https://registry.terraform.io/.well-known/terraform.json returns `{"modules.v1":"/v1/modules/","providers.v1":"/v1/providers/"}`

If instead, a custom registry returns no trailing slash, like `{"modules.v1":"/v1/modules","providers.v1":"/v1/providers"}
Then we'd fetch URLs mashing together the path with the module name.   This fixes that, akin to https://github.com/hashicorp/terraform/commit/34b4000be964095232df9a04e53596f61bb921c5

### DIFF
--- a/internal/hcl/modules/registry.go
+++ b/internal/hcl/modules/registry.go
@@ -111,6 +111,9 @@ func (d *Disco) ModuleLocation(source string) (RegistryURL, bool, error) {
 	if err != nil {
 		return RegistryURL{}, false, fmt.Errorf("unable to discover registry service using host %s %w", host, err)
 	}
+	if !strings.HasSuffix(serviceURL.Path, "/") {
+		serviceURL.Path += "/"
+	}
 
 	r := RegistryURL{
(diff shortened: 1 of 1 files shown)


internal/hcl/modules/registry.go
```diff
@@ -137,6 +140,9 @@ func (d *Disco) DownloadLocation(moduleURL RegistryURL, version string) (string,
 	if err != nil {
 		return "", fmt.Errorf("unable to discover registry service using host %s %w", moduleURL.Host, err)
 	}
+	if !strings.HasSuffix(serviceURL.Path, "/") {
+		serviceURL.Path += "/"
+	}
```
drewblas, line 145: I thought about this second one a lot, not totally sure it's necessary, I think the resolver below was actually already account for mashing two paths together correctly.  But included this anyway for consistency

aliscott (approved):

PR #3423, opened 2025-07-06
Feature: Subtracting cost

### LINKED ISSUE #2360: Feature request: Subtracting costs
For one of our projects we are setting up a zonal GKE cluster, of which one is free under the GCP tier (as is 1 autopilot cluster).
However, as Infracost is not aware of the amount of (zonal) GKE clusters it wouldn't know about that and still indicates that the Cluster Management Fee is $73 per month rather than $0. This significantly increases the price estimation.

Is there a way to subtract a specific amount from the estimation Infracost does? 

For instance:
* The cluster costs $73
* The instance cost $7 per month

In this case Infracost estimates the total costs to be $80.
I'd like to subtract $73 making the total costs only $7.

mohsenkhosroanjam:
This pull request tries to fix #2360 .


aliscott:
Closing this one since it needs more of a design discussion around it.

PR #3429, opened 2025-07-14
fix: nil ptr in azurerm_{windows,linux}_virtual_machine_scale_set

aliscott:


bazaah:
Azure test is failing because of https://github.com/hashicorp/terraform-provider-azurerm/pull/30101, we can fix that in the next one

bazaah (approved):

PR #3519, opened 2026-03-20
Populate EBS throughput for instance and autoscaling group resources

eugeneotto:
The Throughput field on EBSVolume was never set when the volume was created from an `aws_instance` (`root_block_device`, `ebs_block_device`, or launch template `block_device_mappings`) or from an autoscaling group launch configuration/template. This caused gp3 provisioned throughput costs to be silently omitted from cost estimates. Only standalone `aws_ebs_volume` resources had throughput costs calculated correctly.


aliscott (approved):
Thanks @eugeneotto, looks good 🙏

PR #3523, opened 2026-03-26
feat: add support for azurerm_managed_redis

### LINKED ISSUE #3495: Add support for Azure Managed Redis
Terraform - azurerm_managed_redis -  https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/managed_redis

Pricing -  https://azure.microsoft.com/en-us/pricing/details/managed-redis/

Please and thanks

Kunalbehbud:
Fixes #3495

## Summary
This adds support for `azurerm_managed_redis` in Infracost.

The implementation supports the 4 Azure Managed Redis SKU families:
- Balanced
- Memory Optimized
- Compute Optimized
- Flash Optimized

## Pricing model
This MVP calculates the fixed instance cost for Managed Redis by mapping Terraform `sku_name` values such as `Balanced_B3` to the Azure pricing catalog fields used by Infracost.

It currently matches the Redis Cache pricing data using:
- `service = Redis Cache`
- `productFamily = Databases`
- family-specific `productName` values
- matching `skuName` and `meterName`

## Scope
Included in this first version:
- `azurerm_managed_redis` resource support
- pricing for Balanced, Memory Optimized, Compute Optimized, and Flash Optimized SKUs
- golden tests for all 4 SKU families

Not included in this first version:
- High Availability-specific pricing
- usage-based pricing metrics
- persistence/backup/geo-replication cost modeling

## Testing
Ran:
- `INFRACOST_API_KEY=00000000000000000000000000000000 go test ./internal/providers/terraform/azure -run TestAzureManagedRedisGoldenFile -update`
- `INFRACOST_API_KEY=00000000000000000000000000000000 go test ./internal/providers/terraform/azure -run TestAzureManagedRedisGoldenFile`
- `golangci-lint run ./internal/resources/azure ./internal/providers/terraform/azure`

### DIFF
--- a/internal/resources/azure/managed_redis.go
+++ b/internal/resources/azure/managed_redis.go
@@ -61,8 +61,8 @@ func (r *ManagedRedis) BuildResource() *schema.Resource {
 		CostComponents: []*schema.CostComponent{
 			{
 				Name:           fmt.Sprintf("Cache usage (%s)", r.SKU),
-				Unit:           "instances",
-				UnitMultiplier: schema.HourToMonthUnitMultiplier,
+				Unit:           "hours",
+				UnitMultiplier: decimal.NewFromInt(1),
 				HourlyQuantity: decimalPtr(decimal.NewFromInt(instanceCount)),
 				ProductFilter: &schema.ProductFilter{
 					VendorName:    strPtr(vendorName),
--- a/internal/resources/azure/managed_redis.go
+++ b/internal/resources/azure/managed_redis.go
@@ -1,0 +1,96 @@
+package azure
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/shopspring/decimal"
+
+	"github.com/infracost/infracost/internal/resources"
+	"github.com/infracost/infracost/internal/schema"
+)
+
+var managedRedisFamilyNames = map[string]string{
+	"Balanced":         "Balanced",
+	"ComputeOptimized": "Compute Optimized",
+	"FlashOptimized":   "Flash Optimized",
+	"MemoryOptimized":  "Memory Optimized",
+}
+
+// ManagedRedis represents an Azure Managed Redis instance.
+//
+// Resource information: https://learn.microsoft.com/azure/redis/overview
+// Pricing information: https://azure.microsoft.com/en-us/pricing/details/managed-redis/
+type ManagedRedis struct {
+	Address          string
+	Region           string
+	SKU              string
+	HighAvailability bool
+}
+
+func (r *ManagedRedis) CoreType() string {
+	return "ManagedRedis"
+}
+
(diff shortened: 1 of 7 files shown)


Kunalbehbud:
Quick follow-up now that the CLA is sorted. If the Azure Managed Redis handling looks reasonable, I would appreciate a review when someone has time. Happy to adjust pricing logic or tests if needed.

internal/resources/azure/managed_redis.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return &schema.Resource{
+		Name:        r.Address,
+		UsageSchema: r.UsageSchema(),
+		CostComponents: []*schema.CostComponent{
+			{
+				Name:           fmt.Sprintf("Cache usage (%s)", r.SKU),
+				Unit:           "hours",
+				UnitMultiplier: decimal.NewFromInt(1),
+				HourlyQuantity: decimalPtr(decimal.NewFromInt(1)),
```
aliscott: Looks like we need to handle `high_availability_enabled`. This default to true, so I think this field should also default 2 and only be 1 when that is false.

There should be test cases for this as well.

aliscott (changes requested):
Thanks @Kunalbehbud 🙏, one comment.

internal/resources/azure/managed_redis.go
Kunalbehbud: Updated in the latest push. Managed Redis now defaults to 2 instances for high availability and uses 1 instance only when `high_availability_enabled = false`. I added a golden test case for the false case as well. I also registered `azurerm_managed_redis` in the Azure tag support map so external/YOR tags are preserved.

internal/resources/azure/managed_redis.go
```diff
@@ -0,0 +1,96 @@
...
+
+// ManagedRedis represents an Azure Managed Redis instance.
+//
+// Resource information: https://learn.microsoft.com/azure/redis/overview
+// Pricing information: https://azure.microsoft.com/en-us/pricing/details/managed-redis/
+type ManagedRedis struct {
+	Address       string
+	Region        string
+	SKU           string
+	InstanceCount int64
```
aliscott: This would be cleaner if the core resource specified a `HighAvailability` instead, and then within the core resource we work out the instances for high availability.
```suggestion
	HighAvailability bool
```

aliscott (changes requested):

internal/resources/azure/managed_redis.go
Kunalbehbud: Updated, thanks. The core resource now stores `HighAvailability bool`, and `BuildResource` derives the instance count from that. The Terraform provider keeps the Azure default as high availability enabled unless `high_availability_enabled = false` is set.

aliscott (approved):
Thanks @Kunalbehbud! 🙏

PR #3563, opened 2026-05-03
debug: investigate force-cli failure in CI [DO NOT MERGE]

aliscott:
Debug PR to investigate why TestBreakdownPlanError/HCL fails in CI but passes locally. Built on top of #3561's lint pin.

Adds two debug prints:
- in loadRunFlags after reading --terraform-force-cli flag
- in providers.Detect before calling DetectProjectType

Will be closed once we have the answer.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
         with:
-          version: latest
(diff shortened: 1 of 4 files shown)


aliscott:
Investigation complete. Root cause identified: terraform 1.15.1 (what setup-terraform@v3 installs by default in CI) silently succeeds on `-var-file=<missing>` instead of erroring. The test was relying on the older behavior where a missing var-file caused terraform plan to fail. Terraform 1.6.3 locally still errors as expected.

Closing this debug branch. Will follow up with a fix on a separate PR.

PR #3564, opened 2026-05-03
ci: pin terraform to 1.14.0

aliscott:
## Why

`hashicorp/setup-terraform@v3` installs the latest terraform by default. terraform 1.15.x (released 2026-05-01) introduced a behaviour change: `terraform plan -var-file=<missing>` now silently succeeds where it previously errored with `Failed to read variables file`.

`TestBreakdownPlanError` relies on this failure path to verify the breakdown error output. With terraform 1.15.x, the plan succeeds and produces cost data, breaking the test.

This pins terraform to 1.14.0 - confirmed locally to error correctly on missing var-files. Builds on top of #3561 (lint pin).

## Validation

- terraform 1.6.3: errors with "Failed to read variables file" (correct)
- terraform 1.14.0: errors with "Failed to read variables file" (correct)
- terraform 1.15.1: silently succeeds, returns plan output (broken)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -20,6 +20,8 @@ jobs:
 
       - name: Install Terraform
         uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd # v3
+        with:
+          terraform_version: 1.14.0
 
       - name: Check Terraform format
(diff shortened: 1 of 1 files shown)


aliscott:
Pushed to #3561 instead.

PR #3584, opened 2026-06-12
fix: support for_each'd providers with indexed references

### LINKED ISSUE #3369: Support for_each'd providers
When I use a provider generated using a for_each block, I get an error message saying `Invalid provider reference; Provider argument requires a provider name followed by an optional alias, like "aws.foo".`

The thing is, I'm using a valid provider reference _in OpenTofu_.  For example, I'm generating region-specific AWS providers like this:

```hcl
provider "aws" {
  alias = "by_region"
  for_each = toset([
    "us-east-1"
  ])
  region = each.value

  allowed_account_ids = [
    var.account_id,
  ]

  default_tags {
    tags = local.aws_default_tags
  }
}
```

Maybe that's not the best example, since it only has one instance.  But that valid structure means I reference the provider like `aws.by_region["us-east-1"]`.  This works just fine, and is really useful when I have multiple github orgs or aws regions or k8s clusters or whatever.  Terraform has been sitting on that request since at least 2019, but it's implemented and working well in current OpenTofu.

I suspect all that needs to happen here is updating a regex, but maybe it's more involved. 🤷  I'm not trying to downplay the complexity of this project; I really just don't know. 🤣

dannysauer:
## Summary

Adds support for OpenTofu provider configurations that use `for_each`.

Infracost can now resolve indexed provider instances selected dynamically by resources and data sources:

```hcl
provider = aws.by_region[each.value.region]
```

It also supports passing an indexed provider instance to each instance of a child module:

```hcl
module "regional" {
  for_each = var.regions
  source   = "./regional"

  providers = {
    aws = aws.by_region[each.key]
  }
}
```

## Implementation

- Allows valid OpenTofu indexed provider expressions through module discovery, which otherwise rejects dynamic provider indexes.
- Evaluates every provider instance with the corresponding `each.key` and `each.value` in an isolated context.
- Preserves the unindexed provider configuration key, such as `aws.by_region`, in generated plan data.
- Passes the selected provider instance through module `providers` mappings, including nested module calls.
- Preserves provider and index-expression dependencies in the graph evaluator.

Closes #3369
References #3594

## Testing

```sh
go test ./internal/hcl/modules ./internal/hcl -count=1
```

Coverage includes:

- legacy and graph evaluators
- literal and dynamic provider indexes
- `each.key` and nested `each.value` expressions
- indexes derived from locals
- resource, data source, and module provider selection
- nested module provider forwarding
- malformed provider-reference handling

### DIFF
--- a/internal/hcl/attribute.go
+++ b/internal/hcl/attribute.go
@@ -215,6 +215,22 @@ func (attr *Attribute) ProvidersValue() cty.Value {
 	return attr.Value()
 }
 
+func decodeProviderKey(keyExpr *hclsyntax.ObjectConsKeyExpr, ctx *hcl.EvalContext) (string, bool) {
+	if wrapped, ok := keyExpr.Wrapped.(*hclsyntax.ScopeTraversalExpr); ok {
+		return traversalAsString(wrapped.AsTraversal()), true
+	}
+
+	keyValue, diags := keyExpr.Value(ctx)
+	if diags.HasErrors() || !keyValue.IsKnown() || keyValue.IsNull() {
+		return "", false
+	}
+	var key string
+	if err := gocty.FromCtyValue(keyValue, &key); err != nil {
+		return "", false
+	}
+	return key, true
+}
+
 // DecodeProviders decodes the providers block into a map of provider names to provider aliases.
 // This is used by the graph evaluator to make sure the correct edges are created when providers are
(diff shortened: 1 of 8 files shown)


dannysauer:
(FWIW, the branch name says Claude, but Claude Code only started the work)

PR #3587, opened 2026-06-28
fix(aws): avoid panic on malformed ElastiCache node_type with reserved options

ardittirana:
### Problem
`elasticacheReservationResolver.PriceFilter()` reads the second segment of the cache node type without checking the length:

```go
nodeType := strings.Split(r.cacheNodeType, ".")[1] // cache.m3.large -> m3
```

`cacheNodeType` comes from the resource's `node_type`. When it's empty — which happens whenever Infracost can't resolve the value — or otherwise isn't in the expected `cache.<type>.<size>` form, `strings.Split` returns a single-element slice and `[1]` panics:

```
panic: runtime error: index out of range [1] with length 1
.../internal/resources/aws/elasticache_cluster.go:186
```

This crashes a cost run for any `aws_elasticache_cluster` that has `reserved_instance_term` set but an unresolved/odd `node_type`.

### Fix
Guard the split and return an error, consistent with the other validations in the same function. The caller already handles that path gracefully (logs a warning and falls back to the on-demand price filter), so the run no longer crashes.

### Tests
Adds `TestElasticacheReservationResolverPriceFilterMalformedNodeType` covering empty and dot-less node types. It panics on `master` and passes with this change; `go test ./internal/resources/aws/` is green, `gofmt`/`go vet` clean.

### DIFF
--- a/internal/resources/aws/elasticache_cluster.go
+++ b/internal/resources/aws/elasticache_cluster.go
@@ -183,7 +183,11 @@ func (r elasticacheReservationResolver) PriceFilter() (*schema.PriceFilter, erro
 	if !stringInSlice(validOptions, r.paymentOption) {
 		return def, fmt.Errorf("Invalid reserved_instance_payment_option, ignoring reserved options. Expected: %s. Got: %s", strings.Join(validOptions, ", "), r.paymentOption)
 	}
-	nodeType := strings.Split(r.cacheNodeType, ".")[1] // Get node type from cache node type. cache.m3.large -> m3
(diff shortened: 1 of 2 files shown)


ardittirana:
Thanks! CLA's signed now ✅ Happy to tweak anything if you'd like 🙏

glenngillen:
Thank you for your contribution. However, as outlined in our contribution guide we are no longer accepting PRs or new issues directly into this project. If you have a change you'd like to see please open a discussion first where the team will discuss it further if appropriate. Issues (and by extension PRs) are now reserved for things we've agreed to prioritise and are willing to accept/merge once reviewed.

PR #3590, opened 2026-07-01
fix(security): scope Terraform Cloud/registry tokens to trusted hosts

liamg:
Several code paths attached a configured secret token (Terraform Cloud/Enterprise token, or the terragrunt registry token) to an HTTP request whose destination host was derived from untrusted `.tf` / module-source input, with no check that the host was the trusted endpoint. An attacker supplying Terraform that Infracost scans (e.g. a PR) could set the host to their own domain and receive the victim's token as a `Bearer` credential.

The fix applies one rule across all three paths: **only send a configured token to a host that matches the trusted/configured endpoint** (`TERRAFORM_CLOUD_HOST` if set, otherwise `app.terraform.io`).

- **Remote-variables loader** (`remote_variables_loader.go`): the `.tf` `hostname` no longer blindly overrides the trusted host. If it matches, proceed; if a different host is explicitly configured, skip loading (don't send the token); if only a token is set and the `.tf` points elsewhere, error out and tell the user to set `TERRAFORM_CLOUD_HOST`.
- **Remote-plan** (`runRemotePlan`/`cloudAPI`): the primary token is only used when the remote-run host matches the trusted host; otherwise it refuses rather than leaking the token, falling back to the existing host-scoped lookup when no primary token is set.
- **Terragrunt registry** (`terragrunt_registry_service.go`): the registry token is only attached when the request host is in an allowlist (`registry.terraform.io`, `registry.opentofu.org`, `app.terraform.io`, plus `TERRAFORM_CLOUD_HOST` when configured).

## Testing

- New unit tests cover all three paths (loader error/skip cases, remote-plan token selection, registry host allowlist).
- `go build ./...`, `gofmt`, and `go vet` clean on the touched packages.

### DIFF
--- a/internal/providers/terraform/terragrunt_registry_service.go
+++ b/internal/providers/terraform/terragrunt_registry_service.go
@@ -31,6 +31,32 @@ const (
 	authTokenEnvVarName   = "TG_TF_REGISTRY_TOKEN" //nolint
 )
 
+// isTrustedRegistryHost reports whether the TG_TF_REGISTRY_TOKEN may be sent to
+// the given host. The token is a Terraform Cloud/Enterprise registry token, so
+// we only attach it to the well-known public registries and to the configured
+// Terraform Cloud/Enterprise host - never to a host derived from an untrusted
+// module source.
+func isTrustedRegistryHost(host string) bool {
+	trusted := []string{
+		"registry.terraform.io", // public Terraform registry
+		"registry.opentofu.org", // public OpenTofu registry
+		"app.terraform.io",      // Terraform Cloud (incl. private module registry)
+	}
+	// TERRAFORM_CLOUD_HOST lets Terraform Enterprise users point at their own
+	// host, so trust it too when it is explicitly configured.
+	if h := os.Getenv("TERRAFORM_CLOUD_HOST"); h != "" {
+		trusted = append(trusted, h)
+	}
+
+	for _, t := range trusted {
+		if hostsEqual(host, t) {
+			return true
+		}
+	}
+
+	return false
+}
+
 // TerraformRegistryServicePath is a struct for extracting the modules service path in the Registry.
 type TerraformRegistryServicePath struct {
 	ModulesPath string `json:"modules.v1"`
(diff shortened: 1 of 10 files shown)


internal/providers/terraform/terragrunt_registry_service.go
```diff
@@ -31,6 +31,32 @@ const (
...
+// module source.
+func isTrustedRegistryHost(host string) bool {
+	trusted := []string{
+		"registry.terraform.io", // public Terraform registry
+		"registry.opentofu.org", // public OpenTofu registry
+		"app.terraform.io",      // Terraform Cloud (incl. private module registry)
+	}
+	// TERRAFORM_CLOUD_HOST lets Terraform Enterprise users point at their own
+	// host, so trust it too when it is explicitly configured.
+	if h := os.Getenv("TERRAFORM_CLOUD_HOST"); h != "" {
```
aliscott: Do we need to support `TG_TF_DEFAULT_REGISTRY_HOST` as well?

aliscott: Also, do we need anything here to support the `terraform_cloud_host` config value as well?

aliscott (approved):

internal/providers/terraform/terragrunt_registry_service.go
liamg: Good calls — added both.

Since we needed the configured host inside the getter anyway, we've switched the `tfr` getter over to our local `TerraformRegistryGetter` (carrying the `INFRACOST_REGISTRY_PROXY` support across) rather than the fork's, so the host scoping applies at runtime.

- **`terraform_cloud_host`**: the getter now trusts the resolved config value (covers both the `terraform_cloud_host` config option and the `TERRAFORM_CLOUD_HOST` env var), seeded from `ctx.ProjectConfig.TerraformCloudHost`.
- **`TG_TF_DEFAULT_REGISTRY_HOST`**: now trusted as a token host, and also used as the default registry domain for unqualified sources.

Both covered by tests in `cloud_security_test.go`.

liamg:
Verified the fix end-to-end against a local mock registry (logging the `Authorization` header) plus native runs for the pre-request refusals:

- **Path 1 (remote-var loader)** — attacker `hostname` in a `cloud{}` block refuses when no trusted host is set, skips+warns when a different host is pinned, and proceeds normally to `app.terraform.io`/a matching configured host. Token only ever leaves for the trusted host.
- **Path 3 (terragrunt registry token)** — `TG_TF_REGISTRY_TOKEN` is attached only when the host is trusted (public registries, or a host pinned via `terraform_cloud_host`/`TERRAFORM_CLOUD_HOST`/`TG_TF_DEFAULT_REGISTRY_HOST`); an untrusted `tfr://` host gets no `Authorization` header. Confirmed against the getter that is actually wired up.
- **Regression** — public registry modules still download and price fine via both the plain Terraform loader and the terragrunt `tfr://` getter.
- **Path 2 (`runRemotePlan`/`cloudAPI`)** — covered by unit tests (`TestDirProvider_tokenForRemoteHost`); it only fires in Terraform CLI remote-execution mode so it is not practical to exercise end-to-end locally.

PR #3599, opened 2026-07-30
fix(azure): support FC1 Flex Consumption pricing for azurerm_service_plan

srpomeroy:
## Summary

Fixes [ENG-963](https://linear.app/infracost/issue/ENG-963/azurerm-service-plan-fc1-flex-consumption-misrouted-to-free) — `azurerm_service_plan` with `sku_name="FC1"` (Azure Functions Flex Consumption) reported price missing/not found instead of a cost estimate. Raised by a trial customer via [Pylon #1906](https://app.usepylon.com/issues?issueNumber=1906).

**Root cause:** any `sku_name` starting with `f` was unconditionally routed into the Free Plan (`F1`) pricing branch without reassigning the lookup SKU string, so the product filter searched for a nonexistent `"Free Plan"` + `"FC1"` combination and silently resolved to nothing.

**Fix:**
- `azurerm_service_plan`: FC1 now skips as no-price, consistent with the existing EP/WS/Y1 handling — Flex Consumption has no direct cost on the Service Plan itself.
- `azurerm_linux_function_app` / `azurerm_windows_function_app`: now price FC1-backed function apps using the correct Flex Consumption meters. Verified live against the Azure Retail Prices API:
  - `Flex Consumption` / `On Demand` skuName (not `Standard`, which is the legacy Y1 Consumption meter)
  - 100,000 free GB-seconds/month, 250,000 free executions/month
- Mirrors the existing Y1 Consumption cost component pattern (same usage fields: `monthly_executions`, `execution_duration_ms`, `memory_mb`).

**Known limitation:** Always Ready (pre-warmed) instance costs aren't modelled yet — that depends on `site_config.always_ready`, which isn't currently read from Terraform data. Left as a follow-up.

A companion one-line fix for the newer pricing engine is in [infracost/providers](https://github.com/infracost/providers) (same ticket).

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/resources/azure/... ./internal/providers/terraform/azure/...` (short mode) passes
- [x] Golden file tests (`TestAzureRMServicePlan`, `TestAzureRMLinuxAppFunctionGoldenFile`) regenerated and pass for `HCL`/`HCL Graph`. `Terraform CLI` subtest fails identically on unmodified `master` (pre-existing azurerm-v5-provider fixture incompatibility, unrelated to this change) — verified via `git stash`.
- [x] Manually verified with a locally built binary against a reproduction of the customer's scenario: `azurerm_service_plan` (FC1) now shows as free/no-cost instead of a broken price lookup, and the linked `azurerm_linux_function_app` shows a real, non-zero cost ($22.78 execution time + $1.42 executions) computed from live pricing data.

### DIFF
--- a/internal/providers/terraform/azure/function_app.go
+++ b/internal/providers/terraform/azure/function_app.go
@@ -67,6 +67,8 @@ func newFunctionAppFromAppServicePlanRef(d *schema.ResourceData, data *schema.Re
 	skuName := data.Get("sku_name").String()
 	if strings.HasPrefix(strings.ToLower(skuName), "ep") {
 		tier = "premium"
+	} else if strings.HasPrefix(strings.ToLower(skuName), "fc") {
+		tier = "flexconsumption"
 	}
 
 	return &azure.FunctionApp{
--- a/internal/providers/terraform/azure/testdata/function_linux_app_test/function_linux_app_test.golden
+++ b/internal/providers/terraform/azure/testdata/function_linux_app_test/function_linux_app_test.golden
@@ -97,6 +97,10 @@
  ├─ vCPU (EP1)                                                                                                                      1  vCPU                           $126.29    
  └─ Memory (EP1)                                                                                                                  3.5  GB                              $31.43    
                                                                                                                                                                                  
+ azurerm_linux_function_app.function_with_usage["plan-Linux-FC1"]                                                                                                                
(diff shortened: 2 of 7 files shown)


srpomeroy:
Wrong repo... Closing.

PR #3607, opened 2026-08-20
fix(docker): bump builder to golang 1.26.6 for CVE-2026-39821

aliscott:
Follow-up to #3606: the `golang` Docker images set `GOTOOLCHAIN=local`, so the `toolchain go1.26.6` directive in go.mod is ignored inside Docker builds and the published `edge`/`ci-edge` images still ship binaries built with the 1.26.5 stdlib (confirmed via scans of downstream images that embed them).

- Bumps the builder stage in `Dockerfile` and `Dockerfile.ci` to `golang:1.26.6`, which is what actually controls the stdlib in the published images.
- #3606 remains correct for the non-Docker release paths, which run `actions/setup-go` with the default `GOTOOLCHAIN=auto`.
- Adds `GOTOOLCHAIN=path` to both builder stages (suggested by tommyknows) so image/toolchain-directive drift fails the build loudly instead of silently building with the image's stdlib.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.26.5 AS builder
+FROM golang:1.26.6 AS builder
 
 ARG ARCH=linux
 ARG DEFAULT_TERRAFORM_VERSION=0.15.5
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,11 @@ ARG TERRAGRUNT_VERSION=0.31.8
 SHELL ["/bin/bash", "-c"]
 ENV HOME=/app
 ENV CGO_ENABLED=0
+# Force the golang image tag above to be in sync with the `toolchain` directive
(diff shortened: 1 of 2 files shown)


vdmgolub (approved):

tommyknows (approved):
In some other builds we do this in the Dockerfile:
```
# force the GO_VERSION above to be in sync with the `toolchain` directive in go.mod.
# This avoids downloading other Go Toolchains, which happens silently and would slow
# down our image builds.
ENV GOTOOLCHAIN=path
```

I think it'd make sense to add since it would have failed loudly instead of just building with the wrong toolchain.

aliscott:
Good shout, added in the latest commit. `GOTOOLCHAIN=path` in both builder stages, so an image/toolchain-directive drift now fails the build loudly instead of silently building with the image's stdlib (which is exactly how this CVE slipped through).

PR #3617, opened 2026-09-09
Mark azurerm_storage_account_queue_properties as a free resource

srpomeroy:
## Summary
- \`azurerm_storage_account_queue_properties\` only configures logging/retention settings on a storage account's queue service and has no billable cost component.
- Adds it to the free resources list so Infracost reports it as \$0 instead of leaving it unrecognized.

## Test plan
- [ ] Run against a Terraform plan containing this resource and confirm it's shown as free, not unsupported.


aliscott (approved):

srpomeroy:
Closing — this change belongs in infracost/parser instead, see the follow-up PR there.

Issue #3612, opened 2026-09-02
azurerm_managed_disk: PremiumV2_LRS (Premium SSD v2) not priced, silently shows $0 instead of being marked unsupported


justminime:
## Description
Infracost does not support Azure `PremiumV2_LRS` (Premium SSD v2) managed disks and silently fails to price them, rather than reporting them as unsupported.

## Steps to reproduce
```hcl
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~>5.0"
    }
  }
}

provider "azurerm" {
  features {}
}

resource "azurerm_resource_group" "example" {
  name     = "example-resources"
  location = "eastus"
}

resource "azurerm_managed_disk" "example" {
  name                 = "example-disk"
  location             = azurerm_resource_group.example.location
  resource_group_name  = azurerm_resource_group.example.name
  storage_account_type = "PremiumV2_LRS"
  create_option        = "Empty"
  disk_size_gb          = 10752
  disk_iops_read_write = 6000
  disk_mbps_read_write = 500
  zone                 = "1"
}
```
Run `infracost breakdown --path .`

## Actual behavior
The disk is priced at $0 with no warning in the PR comment. The CLI output (with debug logging) shows:
```
WARN Could not map disk type PremiumV2 and size 10752 to disk name
```
This resource is not listed in the "not supported yet" section of the PR comment either, since `azurerm_managed_disk` is generally supported for `Premium_LRS`/`Standard_LRS`/`StandardSSD_LRS` — only the `PremiumV2_LRS` SKU silently fails to resolve a price.

Because of this, a real, billable disk resize (e.g. 10240 GB -> 10752 GB on a `PremiumV2_LRS` disk) shows up as a $0 cost diff in a PR, even though Azure bills `PremiumV2_LRS` per-GiB-provisioned plus provisioned IOPS/throughput (a different, non-tiered pricing model from the legacy Premium SSD P-series).

## Expected behavior
- `azurerm_managed_disk` with `storage_account_type = "PremiumV2_LRS"` should be priced using Azure's Premium SSD v2 billing model (per-GiB-month storage + `disk_iops_read_write` + `disk_mbps_read_write` meters), OR
- If unsupported, it should appear in the "not supported yet" resource list in the PR comment instead of silently pricing at $0.

## Environment
- Infracost CLI version: 0.10.45
- Terraform provider: `hashicorp/azurerm` ~>5.0

[...]

PR #3618, opened 2026-09-11
fix(azure): add PremiumV2_LRS disk pricing support (#3612)

Tyagiquamar:
## Problem
Infracost did not map Azure \PremiumV2_LRS\ (Premium SSD v2) managed disks and silently returned .00 cost components instead of pricing provisioned capacity, IOPS, and throughput.

## Root Cause
\managedDiskCostComponents\ in \internal/resources/azure/managed_disk.go\ passed \PremiumV2\ to standard disk mapping which failed to find a matching fixed SKU size and returned nil.

## Fix
Added \premiumV2DiskCostComponents\ to map \PremiumV2_LRS\ provisioned capacity, IOPS (>3000), and throughput (>125 MBps) cost components.

## Verification
Ran \go test ./internal/resources/azure...\ green.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

